PR #149, opened 2022-01-11
K8s labels filters

arikalon1:


docs/catalog/triggers/kubernetes.rst
```diff
@@ -46,3 +46,25 @@ The following wildcard triggers will fire for any supported Kubernetes resource:
...
 * on_kubernetes_any_resource_all_changes
+
+Supported filters
+--------------------
+
+All Kubernetes triggers support the following filters:
+
+* ``name_prefix`` - Name prefix to match resources.
+* ``namespace_prefix`` - Namespace prefix to match resources.
+* ``labels_selector`` - Match resources with these labels. The format is: ``label1=value1,label2=value2``. If more than one labels is provided, **all** needs to match.
```
aantn, line 57: Should be `need` not `needs`

docs/catalog/triggers/kubernetes.rst
```diff
@@ -46,3 +46,25 @@ The following wildcard triggers will fire for any supported Kubernetes resource:
 * on_kubernetes_any_resource_update
 * on_kubernetes_any_resource_delete
 * on_kubernetes_any_resource_all_changes
+
```
aantn, line 49: Good job documenting it.

docs/catalog/triggers/kubernetes.rst
```diff
@@ -46,3 +46,25 @@ The following wildcard triggers will fire for any supported Kubernetes resource:
...
+Supported filters
+--------------------
+
+All Kubernetes triggers support the following filters:
+
+* ``name_prefix`` - Name prefix to match resources.
+* ``namespace_prefix`` - Namespace prefix to match resources.
+* ``labels_selector`` - Match resources with these labels. The format is: ``label1=value1,label2=value2``. If more than one labels is provided, **all** needs to match.
+
+For example:
```
aantn, line 59: Can we put the example at the top of the page along with one sentence on the motivation?

I think we're jumping in too fast into the fine details without explaining the motivation. We can't assume that people read through the docs and reached this page in the order we expect.

See for example, what I did at https://docs.robusta.dev/master/catalog/triggers/prometheus.html

src/robusta/integrations/kubernetes/base_triggers.py
```diff
@@ -40,6 +40,22 @@ class K8sBaseTrigger(BaseTrigger):
     operation: K8sOperationType = None
     name_prefix: str = None
     namespace_prefix: str = None
+    labels_selector: str = None
+    _labels_map: dict = PrivateAttr()
```
aantn, line 44: I think this has to be `Dict` (from `typing`) and not `dict` for compatibility with lower python versions

aantn (approved):
Looks good. Left some comments on the docs.

docs/catalog/triggers/kubernetes.rst
arikalon1: fixing

docs/catalog/triggers/kubernetes.rst
arikalon1: good idea. Fixing

src/robusta/integrations/kubernetes/base_triggers.py
arikalon1: Fixing

PR #153, opened 2022-01-12
Run skaffold without remote push

RoiGlinik:


Dockerfile
```diff
@@ -1,7 +1,7 @@
 # see https://pythonspeed.com/articles/alpine-docker-python/ for the reason we don't use alpine
 FROM python:3.9-slim
 RUN apt-get update \
-    && apt-get install -y --no-install-recommends git ssh socat wget curl libcairo2 python3-dev libffi-dev socat \
+    && apt-get install -y --no-install-recommends git ssh socat wget curl gcc libcairo2 python3-dev libffi-dev socat \
```
aantn, line 4: Can we use a multi-stage Dockerfile so that we can exclude gcc from the final image? (gcc has a bunch of dependencies so the final image will be smaller this way)

https://docs.docker.com/develop/develop-images/multistage-build/#use-multi-stage-builds

aantn (approved):

Dockerfile
```diff
@@ -16,18 +16,20 @@ RUN mkdir /app
...
-RUN pip3 install git+https://github.com/astanin/python-tabulate.git@b2c26bcb70e497f674b38aa7e29de12c0123708a#egg=tabulate
 
 COPY playbooks/ /etc/robusta/playbooks/defaults
 
 # -u disables stdout buffering https://stackoverflow.com/questions/107705/disable-output-buffering
-CMD [ "python3", "-u", "-m", "robusta.runner.main"]
\ No newline at end of file
+CMD [ "python3", "-u", "-m", "robusta.runner.main"]
+
+FROM dev as prd
```
aantn, line 33: Due to the way you're inheriting from the previous stage here, we still carry the layer that contains `gcc` with us. GCC isn't present in the final image (because you uninstall it) but it is present in one of the previous layers which inflates the image size.

I think you have to either switch to a new base image (FROM newimage) which is clean or do the trick where you install and uninstall gcc on the same line

Dockerfile
```diff
@@ -16,18 +16,20 @@ RUN mkdir /app
...
+COPY pyproject.toml poetry.lock /app/
 WORKDIR /app
-RUN /root/.local/bin/poetry install --no-root --extras "all"
+RUN  /root/.local/bin/poetry install --no-root --extras "all" \
+ &&  apt-get purge -y --auto-remove gcc 
 
 COPY src/ /app/src
 
 RUN pip3 install --use-feature=in-tree-build .
-# Install tabulate version that fixes column width wrapping. Cannot be added to pypi as a git dependency, so adding it here
```
aantn, line 27: I'm not totally sure, but I think we need to keep installing tabulate here. We can't upload the `robusta-cli` package to pypi if it has a git dependency in pyproject.toml

PR #154, opened 2022-01-13
Verify slack channel at config gen

Avi-Robusta:
when generating a config if you choose to add slack it will attempt to send a message to the channel, if it fails it will not generate the config so you wont run a bad config in your cluster


src/robusta/integrations/slack/sender.py
```diff
@@ -257,6 +272,27 @@ def send_finding_to_slack(
             else:
                 blocks.extend(enrichment.blocks)
 
-        self.__send_blocks_to_slack(
+        return self.__send_blocks_to_slack(
             blocks, attachment_blocks, finding.title, slack_channel,unfurl, sink_name, finding.severity
         )
+
+    def send_welcome_message_to_slack_channel(self, channel_name) -> bool:
```
arikalon1, line 279: I think it would be easier to use the slack sdk client for this purpose, and not the slack sender.
The slack sender has a lot more dependencies, so using it in the CLI will make it harder to separate later on.
In addition, the hello message isn't really a finding. wdyt?

Avi-Robusta: sounds good

src/robusta/cli/main.py
```diff
@@ -55,26 +56,6 @@ class HelmValues(BaseModel):
     enablePlatformPlaybooks: bool = False
 
 
-def slack_integration(
```
arikalon1, line 58: thank you..

pyproject.toml
```diff
@@ -61,7 +61,7 @@ rsa = "^4.8"
 
 
 [tool.poetry.extras]
-all = ["Flask", "grafana-api", "manhole", "watchdog", "dulwich", "better-exceptions", "CairoSVG", "tabulate", "kafka-python", "prometheus-api-client", "slack-sdk", "supabase-py", "datadog-api-client", "pygal", "tinycss", "cssselect", "rsa"]
```
arikalon1, line 64: Can we make tabulate optional again?

src/robusta/cli/main.py
```diff
@@ -141,6 +122,9 @@ def gen_config(
                 )
             )
         )
+        if not verify_slack_channel(slack_api_key, cluster_name, slack_channel):
+            typer.echo(f"Invalid channel \"{slack_channel}\", Installation aborted.")
```
arikalon1, line 126: I would try to elaborate a little more here.
i.e.: "We couldn't send our welcome message to channel {slack_channel}. Does this channel exist? Did you connect Robusta to the correct slack workspace?"

src/robusta/cli/slack_verification.py
```diff
@@ -0,0 +1,70 @@
+import typer
+from slack_sdk import WebClient
+from slack_sdk.errors import SlackApiError
+
+SLACK_WELCOME_MESSAGE_TITLE = ":large_green_circle: INFO - Welcome to Robustas"
+SLACK_WELCOME_MESSAGE_HEADER_WITH_CLUSTER_NAME = "You have signed up for alerts for cluster: '{}'."
```
arikalon1, line 6: I think 'alerts' is a little undermining.
I would try something more impressive.
signed for insights, monitoring, or something

arikalon1:
Good job
Left few minor comments

arikalon1:
very nice!

PR #156, opened 2022-01-15
Secret values

arikalon1:


docs/user-guide/configuration.rst
```diff
@@ -171,6 +171,47 @@ Here is a full example showing how to configure all possible sinks:
...
+Configuration secrets
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Some of the configuration values are considered secrets, and cannot be saved in plain text format.
+We recommend using `SealedSecrets <https://github.com/bitnami-labs/sealed-secrets>`_
+or one of the other secret management system for Kubernetes, to encrypt the secret values.
+
+As an alternative, we can pull secret values from Kubernetes secrets.
+
+Here's how:
```
aantn, line 183: I think this would be easier to understand if you reversed the order: first show the secret, then create the env variable, then pull the value into the string. (It doesn't make sense to me to explain first how you pull it in from an environment variable before you've explained why it is an environment variable.)

I would also remove the text "Here's how" as it is obvious, but that's just a personal preference

docs/user-guide/configuration.rst
```diff
@@ -171,6 +171,47 @@ Here is a full example showing how to configure all possible sinks:
...
+
+First, define that the value should be pulled from an environment variable:
+
```
aantn, line 186: I would be more explicit here in the text (and not just the example) that you can use the syntax {{ env.PARAM_NAME }} 

E.g:

  First, define that the value should be pulled from an environment variable by using the special {{ env.VARIABLE }} syntax

src/robusta/core/model/runner_config.py
```diff
@@ -32,3 +33,7 @@ class RunnerConfig(BaseModel):
     ]
     global_config: Optional[dict] = {}
     active_playbooks: Optional[List[PlaybookDefinition]] = []
+
+    @validator('global_config')
```
aantn, line 37: It seems weird to do this with a validator, but I don't have any real objection to that.

aantn (approved):

docs/user-guide/configuration.rst
arikalon1: cool. Fixing

docs/user-guide/configuration.rst
arikalon1: sure

src/robusta/core/model/runner_config.py
arikalon1: I feel the same, but think it's the best possible implementation

PR #158, opened 2022-01-16
the OOMKilled status is not always in the lastState field of the cont…

aantn:
…ainer status, but is sometimes in the state field (as in the case of job-related pods). We therefore try to look for OOMKilled in both fields

w

### DIFF
--- a/playbooks/robusta_playbooks/oom_killer.py
+++ b/playbooks/robusta_playbooks/oom_killer.py
@@ -4,12 +4,20 @@ from datetime import datetime, timezone
 from collections import namedtuple
 
 
-def is_oom_status(status: ContainerStatus):
+def is_last_state_in_oom_status(status: ContainerStatus):
     if not status.lastState:
         return False
     if not status.lastState.terminated:
         return False
     return status.lastState.terminated.reason == "OOMKilled"
+
+
+def is_state_in_oom_status(status: ContainerStatus):
+    if not status.state:
+        return False
+    if not status.state.terminated:
+        return False
+    return status.state.terminated.reason == "OOMKilled"
 
 
 OOMKill = namedtuple("OOMKill", ["datetime", "message"])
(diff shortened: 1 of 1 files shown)


playbooks/robusta_playbooks/oom_killer.py
```diff
@@ -36,9 +44,21 @@ def oom_killer_enricher(event: NodeEvent):
...
+        oom_statuses_from_state = filter(is_state_in_oom_status, pod.status.containerStatuses)
+        for status in oom_statuses_from_state:
+            container_id_to_oom_killed_state[status.state.terminated.containerID] = (status, status.state)
+
+        oom_statuses_from_last_state = filter(is_last_state_in_oom_status, pod.status.containerStatuses)
+        for status in oom_statuses_from_last_state:
+            container_id_to_oom_killed_state[status.lastState.terminated.containerID] = (status, status.lastState)
+
+        for v in container_id_to_oom_killed_state.values():
+            status = v[0]
```
aantn, line 58: should be `containerID` not `status` right?

can you just make this `for (containerID, state) in container_id_to_oom_killed_state.values()`?

playbooks/robusta_playbooks/oom_killer.py
```diff
@@ -4,14 +4,22 @@
...
-def is_oom_status(status: ContainerStatus):
+def is_last_state_in_oom_status(status: ContainerStatus):
     if not status.lastState:
         return False
     if not status.lastState.terminated:
         return False
     return status.lastState.terminated.reason == "OOMKilled"
 
 
+def is_state_in_oom_status(status: ContainerStatus):
```
aantn, line 15: can we avoid the code duplication?

aantn (commented):
Looks good and I'm merging. I've left some aesthetic comments, so please clean those up whne you have a chance.

playbooks/robusta_playbooks/oom_killer.py
djarv1337: We can, but I think it makes the code less readable. Are you okay with avoiding this change?

PR #159, opened 2022-01-16
Skaffold push ruamel compile

RoiGlinik:
fix ruamel.yaml.clib compile error from missing wheel.
use skaffold default push strategy.


aantn (approved):
Awesome, looks good and works for me with a remote cluster. Merging!

PR #160, opened 2022-01-17
break installation steps for ease of use. add pip3 step to solve path…

RoiGlinik:
other solution is to use the python prefix. lets discuss this.

### DIFF
--- a/docs/getting-started/installation.rst
+++ b/docs/getting-started/installation.rst
@@ -9,45 +9,29 @@ The standard installation uses Helm and the robusta-cli, but :ref:`other alterna
 Standard Installation
 ------------------------------
 
-1. Download the Helm chart:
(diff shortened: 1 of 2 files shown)


RoiGlinik:
Thats a second option
image

idea behind each option.
1. installation with pip3 (without the python prefix) will make sure pip3 is in the path and if so, our CLI will be aswell.  handling pip3 not in path is an easier issue (most packages use this approach).

2. adding the prefix python -m to our installation guide for seamless installation, ( users will get issues with later robusta-cli use). the added option.

aantn:
1. The change re. `pip3` looks good to me.

2. I'm not as crazy about turning it from 4 installation steps to 6 steps. It is definitely more correct, but it makes the instructions seem longer and more complicated IMO.

@arikalon1 thoughts?

aantn:
Awesome, merging. (And deleting the merged branch as usual)

PR #164, opened 2022-01-18
Improve the way that the python_memory action chooses which process to profile

aantn:
1. Use the same logic as the python_debugger
2. Move that logic to a shared class
3. Add a function for parsing json with better error reporting (this is in response to several errors we've seen recently where we were missing context)

### DIFF
--- a/src/robusta/core/model/base_params.py
+++ b/src/robusta/core/model/base_params.py
@@ -1,12 +1,21 @@
-from typing import List
+from typing import List, Callable
 from pydantic import SecretStr
 from ...utils.documented_pydantic import DocumentedModel
+from ...integrations.kubernetes.custom_models import Process, RobustaPod
+from ...core.reporting.blocks import (
+    BaseBlock,
+    CallbackBlock,
+    CallbackChoice,
+    MarkdownBlock,
+    TableBlock,
+)
 
 
 class ActionParams(DocumentedModel):
     """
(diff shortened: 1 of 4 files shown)


src/robusta/core/model/base_params.py
```diff
@@ -48,8 +60,106 @@ class GrafanaAnnotationParams(GrafanaParams):
...
+    :var process_substring: process name (or substring).
+    :var pid: pid
+    :var interactive: if more than one process matches, interactively ask which process to choose.
+    """
+
+    process_substring: str = ""
+    pid: int = None
+    interactive: bool = True
+
+    def has_exactly_one_match(self, all_processes: List[Process]) -> bool:
```
arikalon1, line 81: I think the logic shouldn't be part of the base params.
We mix between the model and logic.
It doesn't seem like the right place logically.
In addition, it imports a lot of irrelevant stuff into base params which is used everywhere.
It will be harder to separate the params to a different module if we want it.
I would put that logic in a common utility class

arikalon1:
added a single comment. 
Looks good

aantn:
@arikalon1 please re-review. Don't merge yet though - still testing

arikalon1:
good job

arikalon1:
Feel free to merge

aantn:
@arikalon1 finished testing this and merged

PR #166, opened 2022-01-19
Optional sentry integration to runner

RoiGlinik:


helm/robusta/templates/runner.yaml
```diff
@@ -31,6 +31,12 @@ spec:
             value: "true"
           - name: LOG_LEVEL
             value: {{ .Values.runner.log_level }}
+          - name: SECRET_SENTRY_DSN
+            valueFrom:
+              secretKeyRef:
+                name: secret-sentry
```
arikalon1, line 37: I would name the secret a little more generic
like, robusta-runner-secret
We will probably add other secrets to it sometime in the future.

and because of the above, I would rename dsn to sentry-dsn or similar

helm/robusta/templates/runner.yaml
```diff
@@ -131,4 +137,13 @@ spec:
   selector:
     matchLabels:
       app: {{ .Release.Name }}-runner
+{{ end }}
+{{ if .Values.runner.sentry_dsn }}
```
arikalon1, line 141: I think the manifest separator is missing here '---'
It works without it, but it's easier to read

src/robusta/runner/main.py
```diff
@@ -18,6 +19,11 @@ def main():
     loader = ConfigLoader(registry, event_handler)
     if os.environ.get("ENABLE_MANHOLE", "false").lower() == "true":
         manhole.install(locals=dict(getmembers(robusta_api)))
+
+    sentry_dsn = os.environ.get("SECRET_SENTRY_DSN", "")
+    if sentry_dsn:
+        sentry_sdk.init(sentry_dsn, traces_sample_rate=0.5)
```
arikalon1, line 25: I would take that value from an env variable, and put the default there.
In case we'll need to change it, we can override using an env variable rather than creating a new code version.

see env_vars.py

src/robusta/runner/main.py
```diff
@@ -18,6 +19,11 @@ def main():
     loader = ConfigLoader(registry, event_handler)
     if os.environ.get("ENABLE_MANHOLE", "false").lower() == "true":
         manhole.install(locals=dict(getmembers(robusta_api)))
+
+    sentry_dsn = os.environ.get("SECRET_SENTRY_DSN", "")
+    if sentry_dsn:
```
arikalon1, line 24: Sentry initialization might fail here in case of corrupted url.
I think we don't want to stop the runner initialization if that happens.

I think we need to catch that error, log it, and continue to work without sentry

arikalon1:
very nice !

Few minor comments

PR #171, opened 2022-01-23
Feat: option for out of the box minikube and kind via gen-config

Avi-Robusta:
we wanted to change the size of memory consumption of all our pods to work on smaller clusters for testing

### DIFF
--- a/src/robusta/cli/main.py
+++ b/src/robusta/cli/main.py
@@ -7,7 +7,7 @@ import urllib.request
 import uuid
 import click_spinner
 from distutils.version import StrictVersion
-from typing import Optional, List, Union
+from typing import Optional, List, Union, Dict
 from zipfile import ZipFile
 
 import requests
--- a/src/robusta/cli/main.py
+++ b/src/robusta/cli/main.py
@@ -72,8 +72,8 @@ class HelmValues(BaseModel):
 
     def set_pod_configs_for_small_clusters(self):
         self.kubewatch = PodConfigs.gen_config(FORWARDER_CONFIG_FOR_SMALL_CLUSTERS)
-        self.grafanaRenderer = PodConfigs.gen_config(RUNNER_CONFIG_FOR_SMALL_CLUSTERS)
-        self.runner = PodConfigs.gen_config(GRAFANA_CONFIG_FOR_SMALL_CLUSTERS)
+        self.runner = PodConfigs.gen_config(RUNNER_CONFIG_FOR_SMALL_CLUSTERS)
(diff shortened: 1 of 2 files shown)


Avi-Robusta:
- [ ] need to test final time on minikube
- [ ] need to test final time  on kind

Avi-Robusta:
[] for the test robusta gen-config you need to add the flag --no-is-small-cluster

src/robusta/cli/main.py
```diff
@@ -29,6 +29,10 @@
     replace_in_file,
 )
 
+FORWARDER_CONFIG_FOR_SMALL_CLUSTERS = "256Mi"
```
arikalon1, line 32: I think 64MB should be enough here

src/robusta/cli/main.py
```diff
@@ -29,6 +29,10 @@
...
+RUNNER_CONFIG_FOR_SMALL_CLUSTERS = "512Mi"
+GRAFANA_CONFIG_FOR_SMALL_CLUSTERS = "256Mi"
```
arikalon1, line 34: I think 64MB should be enough here as well. (I just looked at the memory chart for both on small clusters, and it was less than 64MB)
In addition, I would name it GRAFANA_RENDERER , cause grafana is a different pod. might be confusing

src/robusta/cli/main.py
```diff
@@ -47,14 +51,29 @@ class GlobalConfig(BaseModel):
     account_id: str = ""
 
 
+class PodConfigs(Dict[str, Dict[str, Dict[str, str]]]):
+    __root__: Dict[str, Dict[str, Dict[str, str]]]
+
+    @classmethod
+    def gen_config(cls, memory_size: str = "256Mi"):
```
arikalon1, line 58: Why do we need a default here?

src/robusta/cli/main.py
```diff
@@ -99,6 +119,10 @@ def gen_config(
             "Please specify a unique name for your cluster or press ENTER to use the default",
             default=guess_cluster_name(),
         )
+    if is_small_cluster is None:
+        is_small_cluster = typer.confirm(
+            "Are you running a small local cluster? (Like minikube or kind)"
```
arikalon1, line 124: Will "Are you running a mini cluster? (Like ...)" sound better?

arikalon1:
Good job.
Few minor comments

src/robusta/cli/main.py
Avi-Robusta: when you mean for both do you mean robusta runner too?

arikalon1: No, The runner does need 512MB
The renderer and forwarder will be able to work with 64MB

PR #176, opened 2022-01-25
Added python debug stack trace

Avi-Robusta:


playbooks/robusta_playbooks/pod_troubleshooting.py
```diff
@@ -256,6 +256,110 @@ def get_debugger_warnings(data):
...
+    Prints a stack track of a python process and child threads
+
+    Create a finding with the stack trace results.
+    """
+    pod = event.get_pod()
+    if not pod:
+        logging.info(f"stack_trace - pod not found for event: {event}")
+        return
+
+    if not params.pid:
```
arikalon1, line 271: Is pid really mandatory?
Can you ask just as pod stack trace, without a pid filter? (some pods may have only 1 process )

playbooks/robusta_playbooks/pod_troubleshooting.py
```diff
@@ -256,6 +256,110 @@ def get_debugger_warnings(data):
...
+    if not process_finder.has_exactly_one_match():
+        logging.info(f"debugger_stack_trace - pid does not exist")
+        return
+
+    # we have exactly one process to debug
+    pid = params.pid
+    finding = Finding(
+        title=f"Stacktrace on pid {pid}:",
+        source=FindingSource.MANUAL,
+        aggregation_key="python_debugger",
```
arikalon1, line 285: I think the aggregation key here should be 'debugger_stack_trace'

playbooks/robusta_playbooks/pod_troubleshooting.py
```diff
@@ -256,6 +256,110 @@ def get_debugger_warnings(data):
...
+@action
+def advanced_debugging_options(event: PodEvent, params: DebuggerParams):
+    """
+
+    Create a finding with alternative debugging options for received processes ; i.e. Stack-trace or Memory-trace.
+
+    """
+    pod = event.get_pod()
+    if not pod:
+        logging.info(f"python_debugger - pod not found for event: {event}")
```
arikalon1, line 318: should be: advanced_debugging_options

playbooks/robusta_playbooks/pod_troubleshooting.py
```diff
@@ -256,6 +256,110 @@ def get_debugger_warnings(data):
...
+        return
+    finding = Finding(
+        title=f"Advanced debugging for pod {pod.metadata.name} in namespace {pod.metadata.namespace}:",
+        source=FindingSource.MANUAL,
+        aggregation_key="python_debugger",
```
arikalon1, line 323: aggregation key should be advanced_debugging_options

playbooks/robusta_playbooks/pod_troubleshooting.py
```diff
@@ -256,6 +256,110 @@ def get_debugger_warnings(data):
...
+            updated_params = params.copy()
+            updated_params.process_substring = ""
+            updated_params.pids = None
+            updated_params.pid = proc_pid
+            choices[f"StackTrace {updated_params.pid}"] = CallbackChoice(
+                action=debugger_stack_trace,
+                action_params=updated_params,
+                kubernetes_object=pod,
+            )
+            choices[f"MemoryTrace {updated_params.pid}"] = CallbackChoice(
```
arikalon1, line 352: Why do we add the MemoryTrace here?

arikalon1:
Nice.
Few comments

playbooks/robusta_playbooks/pod_troubleshooting.py
```diff
@@ -256,6 +256,110 @@ def get_debugger_warnings(data):
...
+        aggregation_key="python_debugger",
+        subject=FindingSubject(
+            pod.metadata.name,
+            FindingSubjectType.TYPE_POD,
+            pod.metadata.namespace,
+        ),
+    )
+    event.add_finding(finding)
+    finding.add_enrichment(
+        [MarkdownBlock(f"Please Select an advanced debugging choice:")]
```
aantn, line 332: lowercase "S" in Select

playbooks/robusta_playbooks/pod_troubleshooting.py
```diff
@@ -256,6 +256,99 @@ def get_debugger_warnings(data):
...
+
+    Create a finding with the stack trace results.
+    """
+    pod = event.get_pod()
+    if not pod:
+        logging.info(f"debugger_stack_trace - pod not found for event: {event}")
+        return
+
+    process_finder = ProcessFinder(pod, params, ProcessType.PYTHON)
+    if not params.pid and not process_finder.get_lowest_relevant_pid():
```
arikalon1, line 272: Wouldn't it be better to call it only once? (process_finder.get_lowest_relevant_pid())
We call it again at line 275

Isn't it enough to just check if the result of process_finder.get_lowest_relevant_pid() is empty?

in addition, don't you want to stop the playbook if there's no pid? (like return)

playbooks/robusta_playbooks/pod_troubleshooting.py
```diff
@@ -256,6 +256,99 @@ def get_debugger_warnings(data):
...
+        ),
+    )
+    event.add_finding(finding)
+    finding.add_enrichment(
+        [MarkdownBlock(f"Please select an advanced debugging choice:")]
+    )
+    process_finder = ProcessFinder(pod, params, ProcessType.PYTHON)
+    relevant_processes_pids = process_finder.get_pids()
+    if not relevant_processes_pids :
+        logging.info(f"No processes selected for advanced_debugging_options")
```
arikalon1, line 332: What would that look like in slack?
I think it will be: "Please select an advanced debugging choice:", with nothing to select, cause we already created the finding and enrichment.

Can we make it more user friendly ?

PR #177, opened 2022-01-25
Update README.md

Shubh28698:
Hi team,

I am trying to make a small correction here. Instead of using "builtin", we normally use "built-in" nowadays everywhere. I have researched on google about it & seems to be correct.

Hope it helps and is useful.

Thank you

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,14 +22,14 @@ sinks:
   - slack
 ```
 
-There are also builtin automations to analyze common alerts.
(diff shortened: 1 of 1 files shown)


aantn:
Looks good to me! 

Thank you. Please join our Slack (https://join.slack.com/t/robustacommunity/shared_invite/zt-10rkepc5s-FnXKvGjrBmiTkKdrgDr~wg) and DM me your address so we can send you swag.

PR #180, opened 2022-01-26
Instructions for installing robusta in different namespace

avinashupadhya99:
Adds instructions for installing robusta in different namespace(robusta namespace)


aantn (approved):
Looks good to me! :)

PR #183, opened 2022-01-27
Oom killer root cause

djarv1337:


playbooks/robusta_playbooks/oom_killer.py
```diff
@@ -1,30 +1,48 @@
...
-    if not status.lastState.terminated:
-        return False
-    return status.lastState.terminated.reason == "OOMKilled"
+class OomKillerEnricherParams(ActionParams):
+    """
+    :var new_oom_kills_duration_in_sec: In order to avoid duplicated, This playbook will only report OOMKills
+    that occurred in the last new_oom_kills_duration_in_sec seconds.
+    For example, if new_oom_kills_duration_in_sec is 1200, only OOMKills from the last 20 minutes will be considered.
+    """
+    new_oom_kills_duration_in_sec: int = 1200
```
arikalon1, line 13: These parameters are too internal for the common user.
We don't expect them to change that.
I think we should just take those from an env variable. 
maybe add this documentation just in the playbook's code, for a coding user

playbooks/robusta_playbooks/oom_killer.py
```diff
@@ -1,30 +1,48 @@
...
+    :var node_memory_threshold: The maximal amount of memory percentage a node can use.
+    If this amount of memory was exceeded (in the last metrics_duration_in_secs seconds), the playbook will
+    report the corresponding node as the reason for the OOMKill.
+    """
+    node_memory_threshold: float = 0.95
 
 
 @action
-def oom_killer_enricher(event: NodeEvent):
+def oom_killer_enricher(event: NodeEvent, config: OomKillerEnricherParams):
```
arikalon1, line 45: The event here should be a PrometheusEvent.
I don't think this enricher should be able to run on a NodeChange

playbooks/robusta_playbooks/oom_killer.py
```diff
@@ -37,43 +55,211 @@ def oom_killer_enricher(event: NodeEvent):
...
-        for status in oom_statuses_from_last_state:
-            container_id_to_oom_killed_state[status.lastState.terminated.containerID] = (status, status.lastState)
-
-        for v in container_id_to_oom_killed_state.values():
-            status = v[0]
-            state = v[1]
-
-            dt = parse_kubernetes_datetime_to_ms(state.terminated.finishedAt)
-            oom_kills.append([dt, pod.metadata.name, status.name, status.image])
+    if not isinstance(event, PrometheusKubernetesAlert):
```
arikalon1, line 58: If we change the event to PrometheusKubernetesAlert, we remove this

playbooks/robusta_playbooks/oom_killer.py
```diff
@@ -37,43 +55,211 @@ def oom_killer_enricher(event: NodeEvent):
...
                 )
             ]
         )
     else:
         logging.info(f"found no oom killers on {node.metadata.name}")
         event.add_enrichment([])
+
+
+@dataclass
+class MemorySpecs:
```
arikalon1, line 90: We mostly use pydantic for data classes

aantn:
@djarv1337 how did you reproduce each case? Are there any scripts you used for that?

playbooks/robusta_playbooks/oom_killer.py
```diff
@@ -37,43 +55,211 @@ def oom_killer_enricher(event: NodeEvent):
...
+class OomKill:
+    time: float
+    pod_name: str
+    container_name: str
+    image: str
+    memory_specs: MemorySpecs
+    reason: Optional[str]
+
+
+# This class is used only because the OomKillsExtractor should appear before KubernetesOomKillReasonInvestigator, but still know its interface.
```
arikalon1, line 105: I think you can remove the need for that interface, by placing the  KubernetesOomKillReasonInvestigator first

playbooks/robusta_playbooks/oom_killer.py
```diff
@@ -37,43 +55,211 @@ def oom_killer_enricher(event: NodeEvent):
...
+        self.memory_transformer = K8sMemoryTransformer()
+
+    def extract_oom_kills(self) -> List[OomKill]:
+        results: PodList = Pod.listPodForAllNamespaces(
+            field_selector=f"spec.nodeName={self.node.metadata.name}"
+        ).obj
+
+        oom_kills: List[OomKill] = []
+        for pod in results.items:
+            pod_oom_kills = self.get_oom_kills_from_pod(pod)
```
arikalon1, line 127: might be cleaner to just use: 
for pod in results.items:
    oom_kills.extend(self.get_oom_kills_from_pod(pod))

playbooks/robusta_playbooks/oom_killer.py
```diff
@@ -37,43 +55,211 @@ def oom_killer_enricher(event: NodeEvent):
...
+            oom_kill = OomKill(time=dt, pod_name=pod.metadata.name, container_name=c_status.name,
+                               image=c_status.image, memory_specs=memory_specs, reason=None)
+            oom_kill.reason = self.oom_kill_reason_investigator.get_reason(oom_kill)
+            oom_kills.append(oom_kill)
+
+        return oom_kills
+
+    def get_oom_killed_state(self, c_status: ContainerStatus) -> Optional[ContainerState]:
+        # Check if the container OOMKilled by inspecting the state field
+        if self.is_state_in_oom_status(c_status):
```
arikalon1, line 166: Is this really possible? 
That the current status (not previous) is OOMKilled

playbooks/robusta_playbooks/oom_killer.py
```diff
@@ -37,43 +55,211 @@ def oom_killer_enricher(event: NodeEvent):
...
+    @staticmethod
+    def is_last_state_in_oom_status(status: ContainerStatus):
+        if not status.lastState:
+            return False
+        if not status.lastState.terminated:
+            return False
+        return status.lastState.terminated.reason == "OOMKilled"
+
+    @staticmethod
+    def get_memory_specs(resources: Optional[ResourceRequirements]) -> MemorySpecs:
```
arikalon1, line 193: dpath is very nice for such logics.

from dpath.util import get

then, this whole function can be:
        return MemorySpecs(
            requests=get(resources, "memory/requests", None),
            limits=get(resources, "memory/limits", None)
        )

src/robusta/integrations/resource_analysis/memory_analyzer.py
```diff
@@ -0,0 +1,125 @@
...
+        """
+        Gets the maximal memory usage (in percentage) for the node with the given node name, in the time range between now and now - duration.
+        :return: a float between 0 and 1, representing the maximal percentage of memory in use by the given node
+        """
+
+        # We estimate the node memory usage in percentage by the expression (1 - MemAvailable)/MemTotal.
+        # However, since the corresponding metrics contain only the name of the node exporter's instance, and not the node name,
+        # we perform a join of the instance name with the node_uname_info metric, in order to filter on the node name.
+
+        max_memory_usage_in_percentage = self._get_max_value_in_first_series_of_query(
```
arikalon1, line 30: I'm not sure I fully understood the explanation here.
Regardless, in the built-in grafana dashboard the node cpu utilization query is:

instance:node_memory_utilisation:ratio{job="node-exporter", job="node-exporter", instance="$instance"}

Can this help here?

playbooks/robusta_playbooks/oom_killer.py
```diff
@@ -37,43 +55,211 @@ def oom_killer_enricher(event: NodeEvent):
...
+        self.node_reason = None
+
+    def get_reason(self, oom_kill: OomKill) -> str:
+        container_reason = self.get_busy_container_reason(oom_kill)
+        if container_reason is not None:
+            return container_reason
+
+        # Calculate if the node is the reason for the the OOMKill only once, rather than for every OomKill
+        if not self.node_reason_calculated:
+            self.node_reason = self.get_busy_node_reason()
```
arikalon1, line 224: Can there be oom kill on the node, even with low node cpu utilization?
Like, lets say our node has 4gb of memory.
We have 3 pods with 1gb of memory request. Their actual usage is only 100MB each.

 Now, lets assume a process on the node, is trying to use 1.5GB
 Will it be oom killed (cause the node has 4gb, and 3gb is reserved for the pods, even though not used)
 If so, expect the cpu utilization here to be low. 
Can this happen?

playbooks/robusta_playbooks/oom_killer.py
```diff
@@ -37,43 +55,211 @@ def oom_killer_enricher(event: NodeEvent):
...
+        # Check if the container OOMKilled by inspecting the lastState field
+        if self.is_last_state_in_oom_status(c_status):
+            return c_status.state
+
+        # OOMKilled state not found
+        return None
+
+    @staticmethod
+    def is_state_in_oom_status(status: ContainerStatus):
+        if not status.state:
```
arikalon1, line 178: here I would also use: return get(status, "state/terminated/reason", default="") == "OOMKilled"

src/robusta/integrations/resource_analysis/memory_analyzer.py
```diff
@@ -0,0 +1,125 @@
...
+            step,
+            {
+                "timeout": self.default_prometheus_params["timeout"]
+            }
+        )
+
+        return results
+
+
+class K8sMemoryTransformer:
```
arikalon1, line 103: This can probably be a static class, it doesn't have any state

src/robusta/runner/config_loader.py
```diff
@@ -151,7 +152,8 @@ def __load_playbooks_repos(
 
                 playbook_packages.append(playbook_package)
             except Exception as e:
-                logging.error(f"Failed to add playbooks reop {playbook_package} {e}")
+                logging.error(f"Failed to add playbooks repo {playbook_package} due to the following:")
```
arikalon1, line 155: I think you can just use:

logging.error(f"Failed to ...", exc_info=True)

arikalon1:
nice.
Left some comments

playbooks/robusta_playbooks/oom_killer.py
djarv1337: I disagree here: I want to avoid the use of environment variables for configuration altogether (As it has no hierarchy), especially in the case of a playbooks configuration. The params are the configuration for the playbook, and it has a hierarchy and is automatically documented- so why use another mechanism?

Is the purpose here to avoid showing these "advanced" configurations in the docs? If so, I think a better way to address this would be to introduce something like an @advanced decorator for parameters and show the advanced parameter in the docs only if the user clicks "Advanced configuration"

playbooks/robusta_playbooks/oom_killer.py
djarv1337: I agree, changed.

Also making sure the alert is of type HostOomKillDetected.

playbooks/robusta_playbooks/oom_killer.py
djarv1337: Removed

playbooks/robusta_playbooks/oom_killer.py
djarv1337: Yep, sorry :)
Changed.

playbooks/robusta_playbooks/oom_killer.py
djarv1337: Yep, I am aware of that and even wrote it explicitly in the comment.

If that's fine though, I would like to leave it as is, because I like to write my code as a "story":  that is, you read it from top to bottom. That means that if function f() uses function g(), I prefer g() to appear below f(), and the same applies for classes.

playbooks/robusta_playbooks/oom_killer.py
djarv1337: I agree, changed.

playbooks/robusta_playbooks/oom_killer.py
djarv1337: Yep. In fact, that was exactly why we failed to detect OOM kills exactly for that reason (See https://github.com/robusta-dev/robusta/pull/158).

To be honest, the distinction between lastState and state is not exactly clear to me, and I did notice that sometimes the lastState is in OOMKilled state, so the code is simply trying to check both.

playbooks/robusta_playbooks/oom_killer.py
djarv1337: Cool library!

Unfortunately here it doesn't help much, as resources.requests and resources.limits are object and not dictionaries, and the get functions applies for nested dictionaries like json objects:
https://pypi.org/project/dpath/

playbooks/robusta_playbooks/oom_killer.py
djarv1337: Did you mean low node **memory** utilization?

In any case, I don't think the scenario you described is possible.

Basically, the Scheduler splits the pods to 3 QoS classes: Guaranteed, Burstable and BestEffort.
-Guaranteed means all containers of the pod have a "request" and a "limit" in their spec, and that these are equal.
-Burstable means at least one container has a "request" in its spec, and yet it's not Guaranteed.
-BestEffort means there are no "request" in any container specs.
Now, when the Scheduler evicts pods, it will favor the Guaranteed class and will evict Burstable and BestEffort pods first (which means that Guaranteed pods will never be evicted due to a resource usage of other non-Guaranteed pods).

So in the case above, let's assume for the sake of the argument the pod whose process trying to use 1.5GB is BestEffort- which are the least prioritized, and that the 3 others are Guaranteed (request=limit).  In that case the scheduler will allow the BestEffort pod to run, as the node has enough resources for it. However, if the guaranteed pods will actually use 1GB each in some point in the future, the node will lack resources, and the scheduler will evict the BestEffort (1.5GB pod) first.

playbooks/robusta_playbooks/oom_killer.py
djarv1337: Here too: the dpath.util.get applies for nested dictionary-like objects, not for objects ( (it looks for indexers, not fields).

src/robusta/integrations/resource_analysis/memory_analyzer.py
djarv1337: Changed (initialization of factors is now at the module level, as there are no static initializers in Python)

src/robusta/runner/config_loader.py
djarv1337: True, it works. Changed.

djarv1337: That's true, changed.

djarv1337:
Addressed the comments (except Natan's, which will require a short Notion I'm working on now)

Pushing the changes for CR approval / disapproval. If everything is fine please just approve and don't merge, I want to run the manual tests one more time and make sure everything works :)

djarv1337:
@aantn Github doesn't allow to reply to general comments for some reason.
Anyways, with regard to the tests and the scripts I used to test this playbook, I documented it all here:
https://www.notion.so/Manual-tests-c8b881f395b34311b8e8c0ea6a7f8eeb

In general, we lack a testing framework in the meantime, so let's see this manual tests document as part of the CR until we have one

playbooks/robusta_playbooks/oom_killer.py
arikalon1: Are you sure that's the case? 
Because only this week we had an issue with one of our customers, which had the exact same scenario.
A host process was oom killed, even though the host memory utilization was low.
It had few pods with memory requests, that weren't actually using most of it.

playbooks/robusta_playbooks/oom_killer.py
arikalon1: cool

playbooks/robusta_playbooks/oom_killer.py
arikalon1: yeah, you're right, my bad.
Is there a similar package that does that for objects?
I guess implementing one using getattr is also an option
It's not critical though, your call

playbooks/robusta_playbooks/oom_killer.py
arikalon1: I agree scope wise it doesn't belong to env vars.
I just don't see any user ever changing these, cause it's really internal, and if that's the case, why exposing it.
dunno

playbooks/robusta_playbooks/oom_killer.py
arikalon1: I think that might be too much type validation. I think PrometheusAlert is good enough

playbooks/robusta_playbooks/oom_killer.py
djarv1337: Mmm I see, is converting to CONSTS fine then?

arikalon1: yeap, thx

src/robusta/integrations/resource_analysis/memory_analyzer.py
arikalon1: I just thought that there's a calculated metric for node_memory_utilization, which is used in the grafana dashboard, which might be easier to use without calculations.
I'm not sure if both provide the exact same output.

djarv1337: fixed

playbooks/robusta_playbooks/oom_killer.py
djarv1337: fixed

PR #185, opened 2022-01-28
don't track replicas in the babysitter because it's extremely spammy with autoscaling

aantn:
@arikalon1 tiny change, but please review.

I tested this with a regular deployment (not autoscaled). I changed the replicacount and verified that no babysitter notification was sent.

I didn't test this with the autoscaler, but I did review some existing autoscaler diffs in the UI and verified that each of the fields changed there are excluded by the babysitter.


arikalon1:
nice

PR #192, opened 2022-02-02
webhook sink

arikalon1:


src/robusta/core/sinks/webhook/webhook_sink.py
```diff
@@ -0,0 +1,71 @@
...
+                """
+            )
+            while len(message) + len(wrapped) >= self.size_limit:
+                break
+            message += wrapped
+
+        requests.post(self.url, data=message)
+
+    @classmethod
+    def __to_cleat_text(cls, markdown_text: str) -> str:
```
aantn, line 46: typo - should be *clear*

aantn:
I like that we're doing this a generic webhook and not something specific to NFTY.

@RoiGlinik leaving a real code review to you. I didn't read most of the code.

src/robusta/core/sinks/webhook/webhook_sink.py
arikalon1: fixed, thx

src/robusta/core/sinks/webhook/webhook_sink.py
```diff
@@ -0,0 +1,71 @@
...
+            for block in enrichment.blocks:
+                message_lines.extend(self.__to_unformatted_text(block))
+
+        for line in [line for line in message_lines if line]:
+            wrapped = textwrap.dedent(
+                f"""
+                {line}
+                """
+            )
+            while len(message) + len(wrapped) >= self.size_limit:
```
RoiGlinik, line 39: I think an if will be less confusing here

RoiGlinik (approved):
I had a small comment , Otherwise LG

src/robusta/core/sinks/webhook/webhook_sink.py
arikalon1: Will breaking it to 2 lines make it better?
non_empty_lines = [line for line in message_lines if line]
for line in non_empty_lines: ...

arikalon1: Good catch. Fixed

stephanGarland:
Great timing; I have a "work on what you'd like" day tomorrow, so I'll be installing this and playing around with it then.

arikalon1:
great. Make sure you use the latest release, 0.8.21( will be released soon)
We'd love to hear any feedback you might have

PR #194, opened 2022-02-04
create findings on k8s warning events

arikalon1:


docs/developer-guide/actions/writing-playbooks.rst
```diff
@@ -136,6 +136,52 @@ Sometimes you need to prevent an action from running too often. You can use the
...
+.. code-block:: yaml
+
+   - triggers:
+     - on_deployment_create: {}
+     - on_daemonset_update: {}
+     actions:
+     - resource_babysitter: {}
+     - add_deployment_lines_to_grafana: {}
+
+For every incoming event, Robusta will run any playbook, that *one* of it's ``triggers`` is matched.
```
aantn, line 152: More clear to phrase it like this IMO

> For every incoming event, Robusta will run actions if *any* of the ``triggers`` match.

docs/developer-guide/actions/writing-playbooks.rst
```diff
@@ -136,6 +136,52 @@ Sometimes you need to prevent an action from running too often. You can use the
...
+
+All the playbook's ``actions`` will be executed, according to the order specified in the configuration.
+
+Playbooks execution order is according to the order specified in the configuration as well.
```
aantn, line 156: This sentence "Playbooks execution order is according to the order specified in the configuration as well." seems repetetive and to repeat the information in the previous paragraph.

I know what you're trying to say and I don't think it's getting across. Not sure how to better phrase it. Maybe just give an example and show not tell.

docs/developer-guide/actions/writing-playbooks.rst
```diff
@@ -136,6 +136,52 @@ Sometimes you need to prevent an action from running too often. You can use the
...
+
+There are few mechanisms in place, to stop the processing flow, if needed:
+
+1. Stop handing the current event. No other ``playbooks`` or ``actions`` will be executed
```
aantn, line 160: Don't like "Stop handling the current event" because I don't get what it means. I would replace this with something more specific like "Set `event.stop_processing = True`"

docs/developer-guide/actions/writing-playbooks.rst
```diff
@@ -136,6 +136,52 @@ Sometimes you need to prevent an action from running too often. You can use the
...
+   :emphasize-lines: 5
+
+    @action
+    def event_report(event: EventChangeEvent, action_params: EventErrorReportParams):
+
+       if DONT RUN ANYTHING ELSE ON THIS EVENT:
+           event.stop_processing = True  # no need to run any other playbook or action
+           return
+
+2. For a single ``playbook``, skip the following ``actions``. For this we will set the ``skip_playbook`` attribute
```
aantn, line 172: Again, I know what you mean but I don't think anyone else will. Don't have a great suggestion here.

aantn: Maybe something like this:

Actions can stop other actions in the same playbook from firing using the `event.stop_playbook` variable. For example, if the Python code for `action1` sets `event.stop_playbook = True` then `action2` wont execute in the following example:

```yaml
triggers:
...
actions:
- action1
- action2
```

playbooks/robusta_playbooks/events_enrichments.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+@action
+def event_report(event: EventChangeEvent, action_params: EventErrorReportParams):
+    """
+    Create finding based on the kubernetes event
+    """
+    k8s_obj = event.obj.involvedObject
+
+    # creating the finding before the rate limiter, to use the user the service key for rate limiting
```
aantn, line 15: I don't understand the comment. Specificaly what does it mean "to use the user the service key for rate limiting"

src/robusta/integrations/kubernetes/custom_models.py
```diff
@@ -74,6 +74,15 @@ def does_node_have_taint(node: Node, taint_key: str) -> bool:
     return any(t.key == taint_key for t in node.spec.taints)
 
 
+class RobustaEvent:
```
aantn, line 77: This should inherit from the Hikaru `Event` class for consistency with other classes like `RobustaPod`.

Can we also name the function `get_events_for_object` or something more indicative?

playbooks/robusta_playbooks/events_enrichments.py
```diff
@@ -0,0 +1,56 @@
+from robusta.api import *
+
+
+class EventErrorReportParams(FindingKeyParams, RateLimitParams):
+    pass
+
+
+@action
```
aantn, line 8: There are some weird couplings between this action and `event_resource_events`

1. The behaviour with `stop_playbook` is weird. I think we just want a rate_limit feature on the trigger-level (or playbook-level). This doesn't really belong in the `event_reporting`.

2. The behaviour with `FindingKeyParams` is non intuitive and creates a strange coupling between the two actions. Can we find a better way to do this? Couldn't `event_reporter` just set the default finding and then automatically all enrichments added later will be on that finding?

src/robusta/core/triggers/error_event_trigger.py
```diff
@@ -0,0 +1,24 @@
+from robusta.core.playbooks.base_trigger import TriggerEvent
+from robusta.integrations.kubernetes.autogenerated.triggers import EventAllChangesTrigger, EventChangeEvent
+from robusta.integrations.kubernetes.base_triggers import K8sTriggerEvent
+
+
+class WarningEventTrigger(EventAllChangesTrigger):
```
aantn, line 6: Very nice.

aantn:
Good work getting this done fast.

I left some minor comments on the documentation and larger comments regarding the coupling between the two actions you added.

None of the comments are a blocker from my perspective - I think we should get this released and can discuss improvements afterwards.

I would really like to improve the coupling issues if possible.

playbooks/robusta_playbooks/events_enrichments.py
arikalon1: it's a typo.
it's supposed to be:
creating the finding before the rate limiter, to use the service key for rate limiting

playbooks/robusta_playbooks/events_enrichments.py
```diff
@@ -0,0 +1,56 @@
...
+        action_params.rate_limit,
+    ):
+        event.stop_playbook = True  # no need to run the rest of the playbook actions
+        return
+
+    event.add_finding(finding, finding_key=action_params.finding_key)
+
+
+@action
+def event_resource_events(event: EventChangeEvent, action_params: FindingKeyParams):
```
aantn, line 43: Is there duplicate code here with `pod_events_enricher`?

aantn:
To elaborate on what I wrote earlier, maybe we can even have `event_report` explicitly call `event_resource_events` instead of having them appear in the YAML as two separate actions. That is a better coupling IMO then the two workarounds we have now.

playbooks/robusta_playbooks/events_enrichments.py
arikalon1: 1.
I don't think we want to rate limit on the trigger, but on the action.
A trigger can trigger many different actions, some may be rate limited, while others shouldn't be

2. I added the FindingKeyParams, to allow using this action in other playbooks as well.
Without it, the usage is very limiting, because you only allowed to add to the "default" finding.
This way, this is the default behavior, but you can override in case you need to

playbooks/robusta_playbooks/events_enrichments.py
arikalon1: it's similar. 
I'll combine it

docs/developer-guide/actions/writing-playbooks.rst
arikalon1: cool. fixing

docs/developer-guide/actions/writing-playbooks.rst
arikalon1: Changed the description to: Playbooks execution order, is the same as the configuration file order.
added an example as well

docs/developer-guide/actions/writing-playbooks.rst
arikalon1: fixed

docs/developer-guide/actions/writing-playbooks.rst
arikalon1: fixed

playbooks/robusta_playbooks/events_enrichments.py
arikalon1: fixed

src/robusta/integrations/kubernetes/custom_models.py
arikalon1: Removed it all together

playbooks/robusta_playbooks/events_enrichments.py
arikalon1: fixed

docs/developer-guide/actions/writing-playbooks.rst
```diff
@@ -149,39 +149,39 @@ Every Robusta playbook has a list of :ref:`Triggers` and a list of :ref:`Actions
...
+     actions:
+     - resource_babysitter: {}
+   - triggers:  # playbook B
+     - on_deployment_create: {}
+     actions:
+     - add_deployment_lines_to_grafana: {}
 
-    @action
-    def event_report(event: EventChangeEvent, action_params: EventErrorReportParams):
+On the example above, ``playbook A`` will run before ``playbook B``
```
aantn, line 169: Nice

docs/developer-guide/actions/writing-playbooks.rst
```diff
@@ -149,39 +149,39 @@ Every Robusta playbook has a list of :ref:`Triggers` and a list of :ref:`Actions
...
 
-       if DONT RUN ANYTHING ELSE ON THIS EVENT:
-           event.stop_processing = True  # no need to run any other playbook or action
-           return
+There is a mechanisms in place, to stop the processing flow, if needed:
```
aantn, line 171: Much better. Very clear.

docs/developer-guide/actions/writing-playbooks.rst
```diff
@@ -149,39 +149,39 @@ Every Robusta playbook has a list of :ref:`Triggers` and a list of :ref:`Actions
...
 
-2. For a single ``playbook``, skip the following ``actions``. For this we will set the ``skip_playbook`` attribute
+Set ``event.stop_processing = True``. No other ``playbooks`` or ``actions`` will be executed
 
 .. code-block:: python
    :emphasize-lines: 5
 
     @action
     def event_report(event: EventChangeEvent, action_params: EventErrorReportParams):
```
aantn, line 179: I would give the function here the same name as one of the actions above. (Maybe just name them action1, action2 etc)

src/robusta/core/triggers/error_event_trigger.py
```diff
@@ -1,24 +1,51 @@
...
             return False
 
-        return exec_event.get_event().type == "Warning"
+        if exec_event.get_event().type != "Warning":
+            return False
+
+        # Perform a rate limit for this service key according to the rate_limit parameter
+        name = exec_event.obj.involvedObject.name
+        namespace = exec_event.obj.involvedObject.namespace if exec_event.obj.involvedObject.namespace else ""
+        service_key = TopServiceResolver.guess_service_key(name=name,namespace=namespace)
```
aantn, line 46: What happens is `guess_service_key` fails? Do we still continue?

aantn (approved):
Very nice, love it.

docs/developer-guide/actions/writing-playbooks.rst
arikalon1: fixed

src/robusta/core/triggers/error_event_trigger.py
arikalon1: If no service found, we rate limit by the pod name and namespace.
It has no IO or any way to throw an exception

PR #198, opened 2022-02-05
Simplify telegram sink

arikalon1:


aantn:
The doc changes look good. Haven't had a chance to look over the other parts yet.

src/robusta/core/sinks/telegram/telegram_client.py
```diff
@@ -0,0 +1,46 @@
...
+            file_name = file_name.replace(SVG_SUFFIX, PNG_SUFFIX)
+
+        files = {
+            file_type.lower(): (file_name, contents)
+        }
+        response = requests.post(url, files=files)
+
+        if response.status_code != 200:
+            logging.error(
+                f"Failed to send telegram file: chat_id - {self.chat_id} reason - {response.reason} {response.text}"
```
kotlickya, line 44: i would add response.status code code too here

kotlickya: also above

arikalon1: we have it in the response text

PR #202, opened 2022-02-09
Expiramental java tk

Avi-Robusta:


playbooks/robusta_playbooks/java_pod_troubleshooting.py
```diff
@@ -0,0 +1,150 @@
...
+from robusta.integrations.kubernetes.process_utils import ProcessFinder, ProcessType
+from typing import List
+import traceback
+
+from robusta.integrations.kubernetes.custom_models import JAVA_DEBUGGER_IMAGE
+
+@action
+def java_debugger(event: PodEvent, params: ProcessParams):
+    """
+
```
aantn, line 11: Can you get rid of the leading and trailing whitespace. Should just be.

```
"""
Comment without indentation or blank lines
"""
```

src/robusta/integrations/kubernetes/process_utils.py
```diff
@@ -25,7 +25,7 @@ class ProcessFinder:
     def __init__(
         self, pod: RobustaPod, filters: ProcessParams, process_type: ProcessType
     ):
-        if process_type != ProcessType.PYTHON:
+        if process_type not in {ProcessType.PYTHON, ProcessType.JAVA}:
```
aantn, line 28: The changes here are good and clean.

Note that for Java processes we're still calling `pod.get_processes()` below which uses the standard `debug-toolkit` and not the Java version. 

That is OK! No reason to change it. But does it mean we can remove duplicate code like 
 `pod_ps` from the `java-toolkit` because we don't need it there?

playbooks/robusta_playbooks/java_pod_troubleshooting.py
```diff
@@ -0,0 +1,150 @@
+from robusta.api import *
+from robusta.integrations.kubernetes.process_utils import ProcessFinder, ProcessType
+from typing import List
+import traceback
+
+from robusta.integrations.kubernetes.custom_models import JAVA_DEBUGGER_IMAGE
+
+@action
+def java_debugger(event: PodEvent, params: ProcessParams):
```
aantn, line 9: I'm not crazy about calling this `java_debugger` because that implies we attach an actual debugger here which we don't. (This is what `python_debugger` does so it's confusing if `java_debugger` doesn't.)

Alternative names could be:
* java_ps
* java_inspector
* <your idea here>

Whatever the action is called, it is most similar to the playbook action currently called `advanced_debugging_options`. I suggest we give them the same naming structure. (E.g. rename `advanced_debugging_options` to `python_ps` or `python_inspector`.) 

I'm not just nit-picking here. The consistency is important for making these playbooks self explanatory and easy to use for end users. So please update all the messages in both playbooks according to the names you choose. (E.g. "Java debugging session on pod" -> "Java Inspector for pod...")

aantn (commented):
Left a few comments, but all in all LGTM.

Well done ;)

src/robusta/integrations/kubernetes/process_utils.py
Avi-Robusta: yea added issue to remove duplicate code and combine the toolkits

src/robusta/integrations/kubernetes/custom_models.py
```diff
@@ -107,6 +109,16 @@ def get_logs(self, container=None, previous=None, tail_lines=None) -> str:
             tail_lines,
         )
 
+    @staticmethod
+    def create_java_debugger_pod(
```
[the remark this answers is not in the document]
Avi-Robusta: i feel like run_java_toolkit_command is different then exec in java pod, 
one is a specific command structure and another is any exec
I renamed to exec_in_java_pod and used this function in run_java_toolkit_command which specifies a java toolkit command within the java pod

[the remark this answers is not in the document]
Avi-Robusta: wdyt

playbooks/robusta_playbooks/java_pod_troubleshooting.py
Avi-Robusta: im with you

Avi-Robusta: java_process_inspector? python_process_inspector?

playbooks/robusta_playbooks/java_pod_troubleshooting.py
```diff
@@ -91,6 +87,13 @@ def run_jdk_command_on_pid(event: PodEvent, params: ProcessParams, cmd: str, agg
             pod.metadata.namespace,
         ),
     )
+
+    if not params.pid:
+        process_finder = ProcessFinder(pod, params, ProcessType.JAVA)
+        process = process_finder.get_match_or_report_error(finding, "Debug", retrigger_action, java_process_inspector)
```
aantn, line 93: Is there something more accurate than "Debug" we can use here?

aantn (approved):
LGTM

playbooks/robusta_playbooks/java_pod_troubleshooting.py
aantn: I think java_inspector is good. We can leave out the word precess.

PR #212, opened 2022-02-17
nodes discovery

arikalon1:


src/robusta/core/model/nodes.py
```diff
@@ -0,0 +1,94 @@
...
+    allocated_cpu: float
+    pods_count: int
+    pods: str
+    node_info: Dict
+    deleted: bool = False
+
+    def __compare_node_info(self, other_node_info):
+        return self.node_info == other_node_info
+
+    def __eq__(self, other):
```
aantn, line 26: As far as I can tell, you're overriding the Pydantic default just to exclude `deleted` from the comparison. It's currently a little fragile because if someone adds a new field they could forget to add it to __eq__ and cause bugs. 

Could you instead implement __eq__ as follow:

```python
def __eq__(self, other):
  IGNORED_FIELDS = ["deleted"]
  self_no_deleted = {k: v for k,v in self.dict() if k not in ignore_keys} 
  other_no_deleted = {k: v for k,v in other.dict() if k not in ignore_keys} 
  return self_no_deleted == other_no_deleted
```

src/robusta/core/sinks/robusta/dal/supabase_dal.py
```diff
@@ -277,6 +279,80 @@ def get_active_services(self) -> List[ServiceInfo]:
...
+            .execute()
+        )
+        if res.get("status_code") not in [200]:
+            msg = f"Failed to get existing nodes (supabase) error: {res.get('data')}"
+            logging.error(msg)
+            self.handle_supabase_error()
+            raise Exception(msg)
+
+        return [
+            NodeInfo(
```
aantn, line 302: Can you just write something like `NodeInfo(**node)` instead of explicitly listing each field? I think that should work if the names are the same.

src/robusta/core/sinks/robusta/dal/supabase_dal.py
```diff
@@ -277,6 +279,80 @@ def get_active_services(self) -> List[ServiceInfo]:
...
+                pods_count=node["pods_count"],
+                pods=node["pods"],
+                internal_ip=node["internal_ip"],
+                external_ip=node["external_ip"],
+                node_info=json.loads(node["node_info"]),
+            )
+            for node in res.get("data")
+        ]
+
+    def __to_db_node(self, node: NodeInfo) -> Dict[Any, Any]:
```
aantn, line 322: Here to you can use `self.dict()` if the names are the same.

src/robusta/core/model/nodes.py
```diff
@@ -0,0 +1,94 @@
+from pydantic.main import BaseModel
+from typing import List, Dict
+
+
+class NodeInfo(BaseModel):
```
aantn, line 5: This file is only used in the Robusta sink so I don't think it belongs in `core/model/`. Best to reduce the scope and simplify the cross module dependencies.

src/robusta/core/model/nodes.py
```diff
@@ -0,0 +1,94 @@
...
+    name: str
+    node_creation_time: str
+    internal_ip: str
+    external_ip: str
+    taints: str
+    conditions: str
+    memory_capacity_mb: int
+    allocatable_memory_mb: int
```
aantn, line 13: This is a little nit-picky but I would prefer names like `memory_allocatable` and `memory_allocated` etc such that everything related to memory starts with the prefix `memory_`. I think its more readable and works better in autocomplete.

Same for CPU

src/robusta/core/sinks/robusta/dal/supabase_dal.py
```diff
@@ -277,6 +279,80 @@ def get_active_services(self) -> List[ServiceInfo]:
...
+            )
+            .filter("account_id", "eq", self.account_id)
+            .filter("cluster_id", "eq", self.cluster)
+            .filter("deleted", "eq", False)
+            .execute()
+        )
+        if res.get("status_code") not in [200]:
+            msg = f"Failed to get existing nodes (supabase) error: {res.get('data')}"
+            logging.error(msg)
+            self.handle_supabase_error()
```
aantn, line 298: Can we upgrade to a newer supabase-py version? I think the refresh error has been fixed and we can stop using `self.handle_supabase_error()` in newer versions: https://github.com/supabase-community/gotrue-py/issues/9

src/robusta/core/model/nodes.py
```diff
@@ -0,0 +1,94 @@
...
+    allocated_memory_mb: int
+    cpu_capacity: float
+    allocatable_cpu: float
+    allocated_cpu: float
```
aantn, line 17: I would consider storing this in a more generic format, like:

```
resources["cpu"] = ResourceStats(allocated, allocatable, capacity)
```

It's not a must, but this is likely to expand in the future.

src/robusta/core/sinks/robusta/robusta_sink.py
```diff
@@ -44,23 +48,29 @@ def __init__(
...
             signing_key,
         )
-        # start service discovery
+        # start cluster discovery
         self.__active = True
         self.__discovery_period_sec = DISCOVERY_PERIOD_SEC
         self.__services_cache: Dict[str, ServiceInfo] = {}
-        self.__thread = threading.Thread(target=self.__discover_services)
+        self.__nodes_cache: Dict[str, NodeInfo] = {}
+        self.__init_node_cache()
```
aantn, line 56: The node cache seems to work differently than the services cache, in that the services cache isn't initialized here in the constructor. Can we keep the code consistent and have them work similarly?

Furthermore, if this throws (e.g. due to the supabase refresh bug) then service discovery will be entirely down until the runner starts.

aantn (commented):
Left some comments.

I'm also curious about your thoughts on the data model in Supabase.

Right now all the node data in supabase is denormalized and stored in one table. For example, we're storing the names of all pods in the Node table as one flattened field. Same for tains, conditions, etc.

I think we should consider an alternative data model such as:

1. Nodes table
2. Pods table

Maybe even a Resources or Conditions table as well.

It doesn't make a difference for this specific feature, but if we used a separate Nodes and Pods table it would make future features in the UI much easier to implement. In fact, we'd already be able to implement some interesting features in the UI without changing the runner at all and the total amount of data sent (pod names + which node they're on) would be more or less the same.

So I would strongly recommend considering a normalized data model instead of a denormalized one.

src/robusta/core/sinks/robusta/robusta_sink.py
arikalon1: I opened another task to fix the service cache, so it will behave similar to the node cache.
The service cache calls read from the remote DB on every cycle which is redundant.

Re an error on the init, the refresh bug isn't an option here, but there might be another issue.
The discovery should be able to recover from that.
I'll fix it

arikalon1:
> Left some comments.
> 
> I'm also curious about your thoughts on the data model in Supabase.
> 
> Right now all the node data in supabase is denormalized and stored in one table. For example, we're storing the names of all pods in the Node table as one flattened field. Same for tains, conditions, etc.
> 
> I think we should consider an alternative data model such as:
> 
> 1. Nodes table
> 2. Pods table
> 
> Maybe even a Resources or Conditions table as well.
> 
> It doesn't make a difference for this specific feature, but if we used a separate Nodes and Pods table it would make future features in the UI much easier to implement. In fact, we'd already be able to implement some interesting features in the UI without changing the runner at all and the total amount of data sent (pod names + which node they're on) would be more or less the same.
> 
> So I would strongly recommend considering a normalized data model instead of a denormalized one.

I disagree on that.
It doesn't add anything, both on queries, and it just increases the number of insert/update queries.
Re the pods, we might store those in a separate table if we had more data on it, but it's too early to do it now.
I'm not sure we want that table at all, because it might be VERY big and updated VERY often. There are serious performance implications on doing that. I doubt the way we work today will support that

src/robusta/core/sinks/robusta/dal/supabase_dal.py
arikalon1: they aren't

src/robusta/core/model/nodes.py
arikalon1: I don't see any reason to change that

src/robusta/core/sinks/robusta/dal/supabase_dal.py
arikalon1: Opening a separate issue for that, it's unrelated to this specific task

src/robusta/core/model/nodes.py
arikalon1: Not sure why it matters, but ok

aantn:
> I disagree on that. It doesn't add anything, both on queries, and it just increases the number of insert/update queries. Re the pods, we might store those in a separate table if we had more data on it, but it's too early to do it now. I'm not sure we want that table at all, because it might be VERY big and updated VERY often. There are serious performance implications on doing that. I doubt the way we work today will support that

Is the issue here the amount of data, the number of queries, or the number of rows?

As far as I can tell, the amount of data should be the same both ways.

If it is the number of queries, can we use [batch inserts](https://supabase.com/docs/reference/javascript/insert#upsert) to work around that?

If it is the number of rows then you're right. But keep in mind that today every change to a pod requires updating a large row in the Nodes table. So we'll be doing more read/writes to a Pods table but each one will be small compared to heavier queries to the Nodes table.

To give an example of what the normalized model adds, we'd be able show a list of all the pods in each microservice while keeping the amount of stored data more or less the same. (We might need to add a little more data to make it possible, but fundamentally the number of writes and amount of data written wouldn't change.)

 I know writing all the pods seems heavy, but essentially doing that anyway, but in a format that it impossible to query for other purposes.

src/robusta/core/model/nodes.py
aantn: No problem, keep it as is.

src/robusta/core/sinks/robusta/dal/supabase_dal.py
aantn: Yep, thanks

src/robusta/core/sinks/robusta/robusta_sink.py
aantn: Sounds good.

arikalon1:
> > I disagree on that. It doesn't add anything, both on queries, and it just increases the number of insert/update queries. Re the pods, we might store those in a separate table if we had more data on it, but it's too early to do it now. I'm not sure we want that table at all, because it might be VERY big and updated VERY often. There are serious performance implications on doing that. I doubt the way we work today will support that
> 
> Is the issue here the amount of data, the number of queries, or the number of rows?
> 
> As far as I can tell, the amount of data should be the same both ways.
> 
> If it is the number of queries, can we use [batch inserts](https://supabase.com/docs/reference/javascript/insert#upsert) to work around that?
> 
> If it is the number of rows then you're right. But keep in mind that today every change to a pod requires updating a large row in the Nodes table. So we'll be doing more read/writes to a Pods table but each one will be small compared to heavier queries to the Nodes table.
> 
> To give an example of what the normalized model adds, we'd be able show a list of all the pods in each microservice while keeping the amount of stored data more or less the same. (We might need to add a little more data to make it possible, but fundamentally the number of writes and amount of data written wouldn't change.)
> 
> I know writing all the pods seems heavy, but essentially doing that anyway, but in a format that it impossible to query for other purposes.

The issue here is the number row indeed.
The node item isn't that big at all, and it's not updated too often.
I suggest to change to a more complex db model once we need it, and not in advance when there's no requirement for it

arikalon1:
> > > I disagree on that. It doesn't add anything, both on queries, and it just increases the number of insert/update queries. Re the pods, we might store those in a separate table if we had more data on it, but it's too early to do it now. I'm not sure we want that table at all, because it might be VERY big and updated VERY often. There are serious performance implications on doing that. I doubt the way we work today will support that
> > 
> > 
> > Is the issue here the amount of data, the number of queries, or the number of rows?
> > As far as I can tell, the amount of data should be the same both ways.
> > If it is the number of queries, can we use [batch inserts](https://supabase.com/docs/reference/javascript/insert#upsert) to work around that?
> > If it is the number of rows then you're right. But keep in mind that today every change to a pod requires updating a large row in the Nodes table. So we'll be doing more read/writes to a Pods table but each one will be small compared to heavier queries to the Nodes table.
> > To give an example of what the normalized model adds, we'd be able show a list of all the pods in each microservice while keeping the amount of stored data more or less the same. (We might need to add a little more data to make it possible, but fundamentally the number of writes and amount of data written wouldn't change.)
> > I know writing all the pods seems heavy, but essentially doing that anyway, but in a format that it impossible to query for other purposes.
> 
> The issue here is the number row indeed. The node item isn't that big at all, and it's not updated too often. I suggest to change to a more complex db model once we need it, and not in advance when there's no requirement for it

The problem is the number of updates, and it doesn't scale well on RDBMS, so we better try to minimize the number of updates we create. Because when we'll get to that limit, we'll have to re-design our data pipeline to something scalable

src/robusta/core/model/nodes.py
arikalon1: I think it's fine to have it here.
The core/model is the place to have the data model classes.
It might be used somewhere else as well.
And the service info is already defined there too

aantn: K

src/robusta/core/model/nodes.py
arikalon1: updated it

src/robusta/core/model/nodes.py
arikalon1: I'm skipping this change, because cpu stats are float and mem is int

src/robusta/core/sinks/robusta/robusta_sink.py
arikalon1: fixed

src/robusta/core/model/nodes.py
arikalon1: fixed

src/robusta/core/sinks/robusta/dal/supabase_dal.py
arikalon1: we need json.loads(node_info)

src/robusta/core/sinks/robusta/dal/supabase_dal.py
arikalon1: fixed

aantn (approved):
LGTM

PR #233, opened 2022-02-21
Scheduled backups

aantn:


playbooks/robusta_playbooks/pvc_snapshots.py
```diff
@@ -0,0 +1,55 @@
+from pydantic import SecretStr
```
arikalon1, line 1: I think this is redundant

playbooks/robusta_playbooks/pvc_snapshots.py
```diff
@@ -0,0 +1,55 @@
...
+    :var namespace: namespace of the PersistentVolumeClaim to backup
+    :example name: some_pvc_name
+    """
+
+    name: str
+    namespace: str = "default"
+
+
+@action
+def create_pvc_snapshot(event: ScheduledExecutionEvent, params: VolumeSnapshotParams):
```
arikalon1, line 18: Wouldn't it be better to use ExecutionBaseEvent.
This way we can trigger it manually by the CLI or UI.
We use the ScheduledExecutionEvent data only to add the recurrence to the name. Maybe the date is enough

playbooks/robusta_playbooks/pvc_snapshots.py
```diff
@@ -0,0 +1,55 @@
...
+    """
+    Backup a PersistentVolume by creating a VolumeSnapshot
+    """
+
+    pvc = PersistentVolumeClaim().read(name=params.name, namespace=params.namespace)
```
arikalon1, line 23: Do we really need to read it?
We use only the name & namespace we have in the params.
Is this an implicit check that the PVC exists (if it doesn't an exception is thrown, right?)
Should we handle the case it doesn't exist cleaner?

playbooks/robusta_playbooks/pvc_snapshots.py
```diff
@@ -0,0 +1,55 @@
...
+
+    # both the kubernetes client and hikaru don't have proper support for snapshots in the autogenerated classes
+    # hence the workarounds here
+    # see https://stackoverflow.com/questions/67157702/can-i-take-a-volume-snapshot-with-the-k8s-python-client
+    snapshot_body = {
+        "apiVersion": "snapshot.storage.k8s.io/v1beta1",
+        "kind": "VolumeSnapshot",
+        "metadata": {"name": snapshot_name, "namespace": pvc.metadata.namespace},
+        "spec": {
+            # "volumeSnapshotClassName": "example-snapshot-class",
```
arikalon1, line 35: Did you leave the comment on purpose ?

playbooks/robusta_playbooks/pvc_snapshots.py
```diff
@@ -0,0 +1,55 @@
...
+            "source": {"persistentVolumeClaimName": pvc.metadata.name}
+        },
+    }
+    crd_api = client.CustomObjectsApi()
+    crd_api.create_namespaced_custom_object(
+        group="snapshot.storage.k8s.io",
+        version="v1beta1",
+        namespace="default",
```
arikalon1, line 43: Why is it the `default` namespace, and not the snapshot_body namespace?

playbooks/robusta_playbooks/pvc_snapshots.py
```diff
@@ -0,0 +1,55 @@
...
+        plural="volumesnapshots",
+        body=snapshot_body,
+    )
+
+    finding = Finding(
+        title=f"Created VolumeSnapshot {snapshot_name}",
+        aggregation_key="snapshot",
```
arikalon1, line 50: Should this be `PVC Snapshot`, or this is the only snapshot we're going to have?

src/robusta/core/schedule/scheduler.py
```diff
@@ -13,7 +13,12 @@
     SchedulingInfo,
 )
 
-INITIAL_SCHEDULE_DELAY_SEC = os.environ.get("INITIAL_SCHEDULE_DELAY_SEC", 5)
+# this initial delay is important for when the robusta-runner version is updated
+# at the same time a scheduled playbook is added to the configuration
+# for a short time there are two runners running (old+new version) and they both
+# see the new scheduled playbook. we add a delay so that only the new runner ends up
+# running the scheduled playbook
+INITIAL_SCHEDULE_DELAY_SEC = os.environ.get("INITIAL_SCHEDULE_DELAY_SEC", 120)
```
arikalon1, line 21: I think we need to add a note about it in the docs as well, in case someone is configuring a scheduled trigger with a smaller delay. Let them know the first execution is only after 2 min

arikalon1:
nice work!
Left few minor comments

playbooks/robusta_playbooks/pvc_snapshots.py
aantn: Fixed.

playbooks/robusta_playbooks/pvc_snapshots.py
aantn: Yes, I would much rather use a `PersistentVolumeClaimEvent` here. Is it possible though? Is there a way I can define a `scheduled_trigger` in the YAML with something like that?

playbooks/robusta_playbooks/pvc_snapshots.py
aantn: Yes, it's only to check that it exists. How do you recommend handling it?

playbooks/robusta_playbooks/pvc_snapshots.py
aantn: Yes, we're not specifying a snapshot class right now but it's very possible we'll want to in the future.

playbooks/robusta_playbooks/pvc_snapshots.py
aantn: Updated to `volume_snapshot`

src/robusta/core/schedule/scheduler.py
aantn: Added. Lmk what you think.

aantn:
Thanks, updated. Please re-review.

playbooks/robusta_playbooks/pvc_snapshots.py
arikalon1: I don't think we support `scheduled_trigger` and `PersistentVolumeClaimEvent`
But using `ExecutionBaseEvent` will allow it to be triggered both manually and by a scheduled trigger.

playbooks/robusta_playbooks/pvc_snapshots.py
arikalon1: Maybe we should catch the exception and fail nicely.
Should we notify the user the backup has failed using a finding?

src/robusta/core/schedule/scheduler.py
arikalon1: LGTM

playbooks/robusta_playbooks/pvc_snapshots.py
aantn: Fixed.

playbooks/robusta_playbooks/pvc_snapshots.py
aantn: OK, switched the recurrence number with a timestamp and started using `BaseExecutionEvent`.

playbooks/robusta_playbooks/pvc_snapshots.py
arikalon1: If the any action fails will add some generic finding saying there was an error.
Here, you can send something more user friendly, saying "Backup failed because volume xyz doesn't exist"
Maybe even add the same aggregation key.
Not a must obviously, but it might make it easier to look for such errors in the UI

aantn: Pushed a fix.

arikalon1:
Awesome. Merging

PR #237, opened 2022-02-23
Node attributes to all findings

kotlickya:


src/robusta/core/reporting/finding_subjects.py
```diff
@@ -0,0 +1,52 @@
+from typing import Union, Optional
+from ...integrations.kubernetes.autogenerated.events import RobustaDeployment, RobustaJob, RobustaPod, v1ClusterRole, v1ClusterRoleBinding, v1DaemonSet, v1Event, v1HorizontalPodAutoscaler, v1Namespace, v1Node, v1ReplicaSet, v1Service, v1ServiceAccount, v1StatefulSet, v2beta1ClusterRole, v2beta1ClusterRoleBinding, v2beta1DaemonSet, v2beta1Event, v2beta1HorizontalPodAutoscaler, v2beta1Namespace, v2beta1Node, v2beta1ReplicaSet, v2beta1Service, v2beta1ServiceAccount, v2beta1StatefulSet, v2beta2ClusterRole, v2beta2ClusterRoleBinding, v2beta2DaemonSet, v2beta2Event, v2beta2HorizontalPodAutoscaler, v2beta2Namespace, v2beta2Node, v2beta2ReplicaSet, v2beta2Service, v2beta2ServiceAccount, v2beta2StatefulSet
+from .base import FindingSubject
+from .consts import FindingSubjectType
+
+
+class KubeObjFindingSubject(FindingSubject):
+    UnionKubernetesObjects = Optional[Union[
```
arikalon1, line 8: The list of items here isn't constant.
On the events it's auto generated.
I don't think we get that much from this typed object.
I think it adds too much complexity for the benefits it provides

src/robusta/core/reporting/finding_subjects.py
```diff
@@ -0,0 +1,52 @@
...
+        super(KubeObjFindingSubject, self).__init__(
+            obj.metadata.name,
+            finding_subject_type,
+            obj.metadata.namespace,
+            KubeObjFindingSubject.get_node_name(obj)
+        )
+
+    @staticmethod
+    def get_node_name(obj: UnionKubernetesObjects):
+        try:
```
arikalon1, line 28: don't think we need try/except.
how can an exception be thrown?

src/robusta/core/reporting/finding_subjects.py
```diff
@@ -0,0 +1,52 @@
...
+            if obj.kind == "Node":
+                return obj.metadata.name
+            elif (
+                    obj
```
arikalon1, line 32: if there's no obj, we already have an error in if obj.kind ==...

playbooks/robusta_playbooks/babysitter.py
```diff
@@ -63,11 +65,7 @@ def resource_babysitter(event: KubernetesAnyChangeEvent, config: BabysitterConfi
         finding_type=FindingType.CONF_CHANGE,
         failure=False,
         aggregation_key=f"ConfigurationChange/KubernetesResource/Change",
-        subject=FindingSubject(
-            event.obj.metadata.name,
-            FindingSubjectType.from_kind(event.obj.kind),
-            event.obj.metadata.namespace,
-        ),
+        subject=KubeObjFindingSubject(event.obj),
```
arikalon1, line 68: I think in configuration change, we would like to connect it to the node, only if the change is on the node

playbooks/robusta_playbooks/grafana_enrichment.py
```diff
@@ -101,11 +103,7 @@ def report_image_changes(event: KubernetesAnyChangeEvent):
         title=f"{FindingSubjectType.TYPE_DEPLOYMENT.value} {event.obj.metadata.name} updated in namespace {event.obj.metadata.namespace}",
         source=FindingSource.KUBERNETES_API_SERVER,
         aggregation_key="report_image_changes",
-        subject=FindingSubject(
-            event.obj.metadata.name,
-            FindingSubjectType.TYPE_DEPLOYMENT,
-            event.obj.metadata.namespace,
-        ),
+        subject=KubeObjFindingSubject(event.obj, FindingSubjectType.TYPE_DEPLOYMENT),
```
arikalon1, line 106: I think image changes aren't related to a Node. I would just leave the node_name empty here.
In addition, I think I see some old bug here.
It's not always a deployment, it can be any k8s resource.
I think the SubjectType should be taken from the event, and not always be DEPLOYMENT

playbooks/robusta_playbooks/node_enrichments.py
```diff
@@ -121,6 +123,7 @@ def node_health_watcher(event: NodeChangeEvent):
         source=FindingSource.KUBERNETES_API_SERVER,
         aggregation_key="node_not_ready",
         severity=FindingSeverity.MEDIUM,
+        subject=KubeObjFindingSubject(event.obj)
```
arikalon1, line 126: Maybe it's just better to have NodeFindingSubject

src/robusta/core/reporting/finding_subjects.py
```diff
@@ -0,0 +1,52 @@
...
+                return obj.spec.nodeName
+        except:
+            pass
+        return None
+
+
+class PodFindingSubject(FindingSubject):
+    def __init__(
+        self,
+        pod: RobustaPod = None
```
arikalon1, line 45: I would use just `Pod` here.
`RobustaPod` is a wrapper with some utilities, but we really only need the base here

src/robusta/integrations/prometheus/models.py
```diff
@@ -133,8 +134,10 @@ def __get_alert_subject(self) -> FindingSubject:
         elif self.node:
             subject_type = FindingSubjectType.TYPE_NODE
             name = self.node.metadata.name
+        if self.node:
```
arikalon1, line 137: why didn't you put it under the last `elif`? (elif self.node)?

arikalon1:
nice work!
Left few minor comments

playbooks/robusta_playbooks/node_enrichments.py
Avi-Robusta: NodeFindingSubject is too specific and would technically only be used in one place, 
KubeObjFindingSubject also gets the kind and metadata from the object to make the finding subject so its more diverse

src/robusta/integrations/prometheus/models.py
Avi-Robusta: was unsure about the constructor for PrometheusKubernetesAlert and where node, pod, deployment are assigned,
If it was possible for multiple of them to be assigned I wanted to still get the node from the PrometheusKubernetesAlert but from what i see now it just inherits it from only one other event

playbooks/robusta_playbooks/node_enrichments.py
arikalon1: ok

PR #240, opened 2022-02-24
support https links to playbook git repos

kotlickya:


arikalon1:
nice, can you please update the docs as well?

docs/user-guide/configuration.rst
```diff
@@ -234,7 +234,19 @@ Playbook actions are loaded into Robusta using the ``playbookRepos`` Helm value.
...
 
-You can load extra playbook actions from git repositories:
+You can load extra playbook actions in two different ways from git repositories.
+
+1) Loading a git playbook by HTTPS:
+
+.. code-block:: yaml
+
+    playbookRepos:
+      # we're adding the robusta chaos-engineering playbooks here
```
aantn, line 244: I would add a link in the comment to https://github.com/robusta-dev/robusta-chaos

aantn (approved):
You're missing an explanation of *when* you should use each method. I would add an explanation relatively early like "use https for public repos".

Other than that, good.

PR #242, opened 2022-02-24
update to latest version of prometheus stack add cpu limit and request .

RoiGlinik:
update upgrading guide as well


docs/upgrade.rst
```diff
@@ -28,4 +30,48 @@ Notes
...
+
+    robusta version
+
+2. We suggest removing and installing for the smoothest experience, start by uninstalling Robusta:
+
+.. code-block:: bash
+
+    helm uninstall robusta
+
+3. Manually remove the instlled CRDs (`for more info read here <https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack#uninstall-chart>`_):
```
aantn, line 60: Typo: should be `installed`

docs/upgrade.rst
```diff
@@ -3,6 +3,8 @@ Upgrade Guide
 
 Robusta is installed with Helm, so Robusta upgrades are just Helm upgrades.
 
+.. warning:: Upgrading an existing Release to a new major version might require manual actions. :ref:`read here <Upgrading to a new major version>`
```
aantn, line 6: Capital `R` in `Read here`

docs/upgrade.rst
```diff
@@ -28,4 +30,48 @@ Notes
...
+
+.. code-block:: bash
+
+    helm uninstall robusta
+
+3. Manually remove the instlled CRDs (`for more info read here <https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack#uninstall-chart>`_):
+
+.. code-block:: bash
+
+    kubectl delete crd alertmanagerconfigs.monitoring.coreos.com
```
aantn, line 64: I think there is an edge case here that could be dangerous for users:

1. You have an existing Prometheus Operator running in your cluster
2. You install Robusta with/without the Promtheus feature. (If you install it with, you now have two Prometheus instances running in your cluster as part of different Helm releases. This should work in principal and might be unusual but it's possible.)
3. Someone upgrades Robusta and follows these instructions.
4. Because CRDs are global, this breaks their existing Prometheus install which expects older CRDs.

A few thoughts about solutions:
1. If we can do an upgrade for the CRDs without deleting them first, that's probably safer
2. We should probably warn the user to think this through if they have a separate existing Prometheus install. E.g. put a warning on this whole section like "If you have an existing Prometheus Operator installed independently of Robusta then be very careful! Upgrading CRDs will impact all Prometheus Operators in your cluster."

aantn (commented):
I think the headers and flow on this page needs a little improvement. E.g. there is a section called "Helm Upgrade" but everything on this page is a Helm upgrade.  This is contrasted with a section called "Upgrading to a new major version". Is the difference here a major upgrade vs a minor upgrade? Is it a simple upgrade vs a full upgrade? Is it an "Upgrade for Robusta with Prometheus" vs "Upgrade for Robusta without Prometheus"?

helm/robusta/values.yaml
```diff
@@ -188,3 +188,12 @@ kube-prometheus-stack:
             resources:
               requests:
                 storage: 100Gi
+  prometheus-node-exporter:
```
aantn, line 191: One last comment I forgot to add in the review: can you check all the other Kubernetes resources in the `kube-prometheus-stack` Helm chart and make sure we didn't forget to set/unset limits/requests for any of those?

aantn:
@RoiGlinik let me know when I should re-review

docs/upgrade.rst
RoiGlinik: I updated the upgrading guide. i think the with/without prometheus is probably better.
added the info about the edge cases. tried them a bit, and installation seems to fail in these cases. I would say its a very rare edge case

docs/upgrade.rst
```diff
@@ -28,4 +30,50 @@ Notes
...
+From 0.8.x to >= 0.9.x 
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+1. Check robusta version, look under Robusta:
+
+.. code-block:: bash
+
+    helm list
+
+2. Due to the upgrade of the dependency, kube-state-metrics chart, removal of its deployment/stateful needs to be done manually prior to upgrading:
```
aantn, line 54: This sentence is a little confusing. Can we just write:

```
The kube-state-metrics chart needs to be manually removed prior to upgrading:
```

aantn (commented):
Nice. I've left some comments, mostly grammar nitpicks and recommendations for changing certain CPU requests.

docs/upgrade.rst
aantn: Thanks, looks good.

RoiGlinik:
@aantn  you sure about the requests, they add up to quite a bit. like around 1.5. my Kind cluster have 5 CPU so its ok i guess. but my sample GKE has 2 with 1 reserved to system. so it wont be able to schedule it anymore(from what I understand).
I will fix those. might need to think about it a bit

helm/robusta/values.yaml
```diff
@@ -145,6 +147,7 @@ runner:
   sentry_dsn:
   resources:
     requests:
+      cpu: 250m
```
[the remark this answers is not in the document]
arikalon1: @aantn I saw this. but I think 0.5 CPU is A LOT.
Since it's only bursts, and we're just using CPU request, I think 0.25 is really enough.

helm/robusta/values.yaml
```diff
@@ -172,10 +175,17 @@ kube-prometheus-stack:
...
+      resources:
+        requests:
+          cpu: 50m
   kubeProxy:
     enabled: false
   prometheus:
     prometheusSpec:
+      resources:
+        requests:
+          cpu: 50m
```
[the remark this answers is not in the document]
arikalon1: 500m seems way too much IMO.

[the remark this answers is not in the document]
aantn: @arikalon1 how much is Prometheus using on your cluster?

aantn:
No hard opinion on this this. Choose the values you guys think are best!

PR #245, opened 2022-02-25
Telegram tables

arikalon1:


src/robusta/core/playbooks/playbooks_event_handler_impl.py
```diff
@@ -168,7 +170,13 @@ def __run_playbook_actions(
                     execution_event.response = self.__error_resp(msg)
                     continue
 
-                registered_action.func(execution_event, params)
+                try:
+                    registered_action.func(execution_event, params)
+                except Exception:
+                    logging.error(f"Failed to execute action {action.action_name} {action_params}", exc_info=True)
+                    execution_event.add_enrichment(
+                        [MarkdownBlock(text=f"Oops.. Error processing {action.action_name}")]
```
aantn, line 178: English nitpick: should be `...` and not `..` IMO

aantn: Can we add the exception here as a FileBlock? It's the first thing I would want to see if this happened.

src/robusta/core/playbooks/playbooks_event_handler_impl.py
```diff
@@ -182,8 +190,8 @@ def __get_fired_trigger(
         return None
 
     def __handle_findings(self, execution_event: ExecutionBaseEvent):
-        for finding in execution_event.findings:
-            for sink_name in execution_event.named_sinks:
+        for sink_name in execution_event.named_sinks:
+            for finding in execution_event.sink_findings[sink_name]:
```
aantn, line 194: I would update the comment a few lines below about deep copy. I think we do need a deep copy because enrichments are shared between diff. sinks but I would emphasize in the comment that the problem is the enrichments, not the finding itself which is no longer shared.

src/robusta/core/model/events.py
```diff
@@ -28,7 +29,7 @@ class ExecutionEventBaseParams(BaseModel):
 # (note that we need to integrate with dataclasses because of hikaru)
 @dataclass
 class ExecutionBaseEvent:
-    findings: List[Finding] = field(default_factory=lambda: [])
+    sink_findings: Dict[str, List[Finding]] = field(default_factory=lambda: defaultdict(list))
```
aantn, line 32: Looks good, but I would update the variable names here and add some documentation about non-obvious stuff. I think the current behaviour is:

`sink_findings` - this is something like `shared_findings` - it includes all Findings generated so far on the current trigger-event and it is shared between all ExecutionBaseEvents of different types. It can contain findings for sinks that the current ExecutionBaseEvent doesn't care about.

`named_sinks` - this is something like `my_sinks` - it contains a subset of `sink_findings.keys()` which the current ExecutionBaseEvent actually cares about and is configured for.

Just to emphasize, the behaviour here looks good. I would just document more.

aantn (approved):
LGTM. Left some minor comments.

src/robusta/core/playbooks/playbooks_event_handler_impl.py
arikalon1: regarding the ..., sure
I don't think this is the place to add the exception.
This is good for the Robusta developer, but for most users, they don't care about that exception. 
Most of the times it's an SRE or a developer, who doesn't care about Robusta.
It will only confuse them IMO. 
WDYT?

aantn: I think we should still add it in some way. It's the first thing we're going to need when this occurs.

What if we make it a FileBlock so it doesn't show up in the message by default? (We could give it an extension like .log and not .txt if it will prevent slack from showing a preview.) In other words, we can try to hide it from being displayed prominently, but I think it should be there.

src/robusta/core/model/events.py
arikalon1: I think the name `sink_findings` is good, because it's the findings collected to be sent to each sink.
I'll add some more docs about the behavior which isn't self explanatory

PR #249, opened 2022-02-27
start fixing triggers in autogenerated docs

aantn:
Sometimes the triggers in the autogenerated docs don't make much sense and don't match the playbook action.

This allows passing the Sphinx directive a recommended trigger for each action which will be shown in the docs.

It isn't perfect, but it is better than what we're doing today and lets us fix some really obvious mistakes.


aantn:
@arikalon1 I've updated this and it's ready for review. There are two main changes:

1. Make it possible to control for each action which trigger appears in the autogenerated YAML. This is important to prevent confusing autogenerated examples.
2. Link each trigger in the autogenerated docs to the relevant trigger page in the docs. @saiyam1814 suggested this and it's a great idea. It makes it much easier to read the docs and see the relevant documentation for each trigger.

Internally, I've made one more change: I moved all Sphinx RST generation out of robusta.core and into our autorobusta Sphinx extension. I think it makes more sense that robusta.core doesn't know anything about Sphinx and/or RST.

aantn:
Attaching below a screenshot of what the autogenerated docs now look like, inline with @saiyam1814's suggestion

Screen Shot 2022-03-01 at 10 57 09

PR #251, opened 2022-02-28
Send periodic cluster status data to supabase

RoiGlinik:


src/robusta/core/model/cluster_status.py
```diff
@@ -0,0 +1,10 @@
+from datetime import datetime
+from pydantic.main import BaseModel, Optional
+
+class ClusterStatus(BaseModel):
+    created_at: Optional[datetime]
```
arikalon1, line 5: I think we don't need created_at and updated_at here.
created_at is updated by the DB to now() the when the row is created
updated_at is always sent to the DB as now(), so no point having a member for it

src/robusta/core/sinks/robusta/dal/supabase_dal.py
```diff
@@ -349,3 +351,22 @@ def handle_supabase_error(self):
             self.sign_in()
         except Exception as e:
             logging.error(f"Failed to sign in on error", exc_info=True)
+
+    def to_db_cluster_status(self, data: ClusterStatus) -> Dict[str, Any]:
+        db_cluster_status = data.dict()
+        del db_cluster_status['created_at']
```
arikalon1, line 357: Following the comment above, no need to remove it as well

src/robusta/core/sinks/robusta/robusta_sink.py
```diff
@@ -30,6 +30,7 @@ def __init__(
         super().__init__(sink_config.robusta_sink)
         self.token = sink_config.robusta_sink.token
         self.cluster_name = cluster_name
+        self.account_id = account_id
```
arikalon1, line 33: I would use the account_id from the token.
It's safer to assume the token has the correct one.

src/robusta/core/sinks/robusta/robusta_sink.py
```diff
@@ -231,8 +232,27 @@ def __discover_nodes(self):
                 exc_info=True,
             )
 
+    def __update_cluster_status(self):
+        if self.telemetry is None:
```
arikalon1, line 236: I think telemetry should never be None, and this check should be removed

src/robusta/core/sinks/sink_base.py
```diff
@@ -15,3 +16,6 @@ def write_finding(self, finding: Finding, platform_enabled: bool):
         raise NotImplementedError(
             f"write_finding not implemented for sink {self.sink_name}"
         )
+
+    def inject_telemetry(self,  tel):
```
arikalon1, line 20: This is never called I think

src/robusta/core/sinks/sink_base.py
```diff
@@ -7,6 +7,7 @@ def __init__(self, sink_params: SinkBaseParams):
         self.sink_name = sink_params.name
         self.params = sink_params
         self.default = sink_params.default
+        self.telemetry = None
```
arikalon1, line 10: I like the idea to save it on the sink.
But I think telemetry is never initialized.
I would save a reference of `registry` rather than telemetry on the sink.
I think it's more generic, and might be used for other stuff by the sinks later on
I would pass the registry and save it on the sink in the constructor.
And I would add the telemetry to the registry

src/robusta/runner/telemetry.py
```diff
@@ -0,0 +1,18 @@
+import logging
+from ..model.config import Registry
+
+class Telemetry:
+    def __init__(self, registry: Registry, runner_version: str):
+        self._data = dict()
+        self._data["runner_version"] = runner_version
+
+        for k,v in registry.get_sinks().get_all().items():
```
arikalon1, line 9: This is hard to find.
I would just make the Telemetry class extend BaseModel, with the telemetry attributes, and save it in the registry

src/robusta/runner/web.py
```diff
@@ -50,6 +53,8 @@ def handle_alert_event():
             Web.alerts_queue.add_task(
                 Web.event_handler.handle_trigger, PrometheusTriggerEvent(alert=alert)
             )
+
+        Web.telemetry.set_data({ "last_alert_at": datetime.now()})
```
arikalon1, line 57: would access the telemetry from the registry on the event_handler

arikalon1:
Nice job.
Liked how you solved it.
Left some comments

src/robusta/core/sinks/robusta/robusta_sink.py
```diff
@@ -231,8 +232,27 @@ def __discover_nodes(self):
...
+            return
+        try:
+            cluster_status = ClusterStatus(
+                cluster_id= self.cluster_name,
+                version= self.telemetry.get("runner_version"),
+                last_alert_at= self.telemetry.get("last_alert_at"),
+                account_id= self.account_id
+            )
+
+            self.dal.publish_cluster_status(cluster_status)
```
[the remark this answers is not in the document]
RoiGlinik: Regarding this issue. I think setting for a larger interval is good. 
Problem is I'm afraid we might miss the first prometheus alert and the UI will show error for a while.
I want to find an elegant general solution to the detect some change in the telemetry.
can we postpone that fix for the next telemetry issue?

RoiGlinik:
also when fixing your comments i saw the runner.global_config, it might be some kind of lighter alternative solution to the problem in the future

src/robusta/core/sinks/robusta/robusta_sink.py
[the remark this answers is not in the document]
arikalon1: hmm
What if we saved first_alert_time as well
And send every short period, only if it's not empty, and only up to 10 min passed since first_alert_time.
Will that work?

src/robusta/core/sinks/sink_factory.py
```diff
@@ -21,12 +21,12 @@
 class SinkFactory:
     @classmethod
     def create_sink(
-            cls, sink_config: SinkConfigBase, account_id: str, cluster_name: str, signing_key: str
+            cls, sink_config: SinkConfigBase, account_id: str, cluster_name: str, signing_key: str, registry
```
arikalon1, line 24: Since account_id, cluster_name, signing_key are all taken from registry.global_config, do you think it will be cleaner to just pass registry, and store it on the SinkBase.
Then, each concrete sink can use whatever params it needs from registry.global_config
WDYT?

src/robusta/core/sinks/robusta/robusta_sink.py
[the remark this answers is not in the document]
RoiGlinik: ok did something in that flavor

src/robusta/core/sinks/datadog/datadog_sink.py
```diff
@@ -28,9 +28,10 @@
 
 
 class DataDogSink(SinkBase):
-    def __init__(self, sink_config: DataDogSinkConfigWrapper, cluster_name: str):
-        super().__init__(sink_config.datadog_sink)
-        self.cluster_name = cluster_name
+    def __init__(self, sink_config: DataDogSinkConfigWrapper, registry):
+        super().__init__(sink_config.datadog_sink, registry)
+        global_config = self.registry.get_global_config()
```
arikalon1, line 33: We're repeating these lines few times.
Wouldn't it be easier to set these 3 attributes in the SinkBase, and access it where ever needed in the inheriting classes?

src/robusta/core/sinks/robusta/robusta_sink.py
```diff
@@ -231,10 +239,39 @@ def __discover_nodes(self):
...
     def __discover_cluster(self):
         while self.__active:
             self.__discover_services()
             self.__discover_nodes()
             time.sleep(self.__discovery_period_sec)
 
         logging.info(f"Service discovery for sink {self.sink_name} ended.")
+
+    def __periodic_cluster_status(self):
+        retry_count = 5
```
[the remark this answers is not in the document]
RoiGlinik: fixed as we disscused

src/robusta/core/sinks/datadog/datadog_sink.py
RoiGlinik: good idea

PR #254, opened 2022-03-01
Add PersistentVolume triggers

arikalon1:


aantn:
Don't forget to update the docs with this too: https://docs.robusta.dev/master/catalog/triggers/kubernetes.html

We need a better solution there that's autogenerated, but for now it has to be updated manually

arikalon1:
> Don't forget to update the docs with this too: https://docs.robusta.dev/master/catalog/triggers/kubernetes.html
> 
> We need a better solution there that's autogenerated, but for now it has to be updated manually

Good catch

PR #255, opened 2022-03-01
track init container crashes by default

aantn:
@arikalon1 feel free to hand off the review to Roi or Avi if you like.

This fixes an issue I heard yesterday from someone in the Slack community. Namely, we don't currently track crashes to init containers.

### DIFF
--- a/playbooks/robusta_playbooks/restart_loop_reporter.py
+++ b/playbooks/robusta_playbooks/restart_loop_reporter.py
@@ -7,21 +7,22 @@ class RestartLoopParams(RateLimitParams):
     :var restart_reason: Limit restart loops for this specific reason. If omitted, all restart reasons will be included.
     """
 
-    restart_reason: str = None
+    restart_reason: List[str] = None
 
 
 def get_crashing_containers(
     status: PodStatus, config: RestartLoopParams
 ) -> [ContainerStatus]:
+    all_statuses = status.containerStatuses + status.initContainerStatuses
     return [
         container_status
-        for container_status in status.containerStatuses
+        for container_status in all_statuses
         if container_status.state.waiting is not None
         and container_status.restartCount
         > 1  # report only after the 2nd restart and get previous logs
         and (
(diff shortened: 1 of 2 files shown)


playbooks/robusta_playbooks/restart_loop_reporter.py
```diff
@@ -7,21 +7,22 @@ class RestartLoopParams(RateLimitParams):
     :var restart_reason: Limit restart loops for this specific reason. If omitted, all restart reasons will be included.
     """
 
-    restart_reason: str = None
+    restart_reason: List[str] = None
```
arikalon1, line 10: This is a breaking change, right?
We need to remember to add it in the release notes

playbooks/robusta_playbooks/restart_loop_reporter.py
```diff
@@ -7,21 +7,22 @@ class RestartLoopParams(RateLimitParams):
...
         container_status
-        for container_status in status.containerStatuses
+        for container_status in all_statuses
         if container_status.state.waiting is not None
         and container_status.restartCount
         > 1  # report only after the 2nd restart and get previous logs
         and (
             config.restart_reason is None
-            or container_status.state.waiting.reason == config.restart_reason
+            or container_status.state.waiting.reason in config.restart_reason
```
arikalon1, line 25: Will it works to just leave the original reason, and check if config.restart_reason in waiting.reason
this way it's a backward compatible change
WDYT?

arikalon1:
nice and clean. 
Minor comments

playbooks/robusta_playbooks/restart_loop_reporter.py
aantn: Yes, very good point.

playbooks/robusta_playbooks/restart_loop_reporter.py
aantn: Very good idea. I'll update it.

PR #256, opened 2022-03-01
Adding events to robusta sink

Avi-Robusta:


src/robusta/core/playbooks/common.py
```diff
@@ -17,7 +17,7 @@ def get_resource_events_table(table_name: str, kind: str, name: str, namespace:
         headers = ["reason", "type", "time", "message"]
         rows = [
             [event.reason, event.type, parse_kubernetes_datetime_to_ms(event.lastTimestamp), event.message]
-            for event in event_list.items
+            for event in event_list.items if event.lastTimestamp
```
Avi-Robusta, line 20: I had an event that had no last timestamp and it threw here, it was weird

src/robusta/core/sinks/robusta/robusta_sink.py
```diff
@@ -110,6 +110,23 @@ def __publish_new_services(self, active_services: List):
         # save the cached services in the resolver.
         TopServiceResolver.store_cached_services(list(self.__services_cache.values()))
 
+    def __get_events_history(self):
```
Avi-Robusta, line 113: I would wrap this function in a try catch so if there is any unexpected errors here we will atleast run the sink

src/robusta/core/playbooks/common.py
arikalon1: yeah, we saw that.
Good catch

arikalon1: Do we really want to remove events without timestamp?
Would it be better to add them with some fixed timestamp?
WDYT?

arikalon1:
nice work.
left few comments

src/robusta/core/playbooks/common.py
aantn: I'll second that thought. I recently got a report from a user about this happening on their environment. 

I think the best solution is to add them to the table and either leave the time column empty or write `time missing`

src/robusta/core/playbooks/internal/discovery_events.py
```diff
@@ -17,3 +19,54 @@ def cluster_discovery_updates(event: KubernetesAnyChangeEvent):
...
+        finding_type=FindingType.ISSUE,
+        aggregation_key=f"Kubernetes {event.type} Event",
+        subject=FindingSubject(
+            k8s_obj.name,
+            FindingSubjectType.from_kind(k8s_obj.kind.lower()),
+            k8s_obj.namespace,
+        ),
+        creation_date=event.metadata.creationTimestamp
+    )
+    finding.service_key = f"{k8s_obj.namespace}/{k8s_obj.kind}/{k8s_obj.name}"
```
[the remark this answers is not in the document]
kotlickya: I could see this being an issue for services that currently aren't running,  so TopServiceResolver.guess_service_key will return "" constantly in those cases, we will still need to make our own service key either way (this is the same format) so it sounds slightly redundant.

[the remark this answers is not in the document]
arikalon1: it's not the same format
here it's the pod name, like robusta_runner_xckjhkdsjhf_dsfsd, and not the service name which is robusta_runner
I think it's better to have events only for existing services.

src/robusta/core/sinks/robusta/robusta_sink.py
```diff
@@ -232,6 +249,7 @@ def __discover_nodes(self):
             )
 
     def __discover_cluster(self):
+        self.__get_events_history()
```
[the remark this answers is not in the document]
kotlickya: yea, didn't catch that

src/robusta/core/playbooks/common.py
kotlickya: the only options are making a magic timestamp that we can understand from the UI is missing a timestamp, or making a column in the issues table for no timestamp, also sounds like a waste for an uncommon event

the checking all three timestamps is easily do-able though

kotlickya: @arikalon1 , Im not sure the previous event would work in this case since there is no guarantee its in chronological order,
also having a bunch of events in the UI with the time of 0 could look buggy

src/robusta/core/playbooks/common.py
```diff
@@ -29,13 +32,15 @@ def get_resource_events_table(table_name: str, kind: str, name: str, namespace:
 
 
 def get_event_timestamp(event: Event):
-    if event.lastTimestamp:
+    if event.metadata.creationTimestamp:
```
arikalon1, line 35: i think the creationTimestamp should be the last option.
The event is being created, then updated.
So the lastTimestamp is probably the most important, and should be first

src/robusta/core/playbooks/common.py
```diff
@@ -16,7 +16,10 @@ def get_resource_events_table(table_name: str, kind: str, name: str, namespace:
     if event_list.items:
         headers = ["reason", "type", "time", "message"]
         rows = [
-            [event.reason, event.type, parse_kubernetes_datetime_to_ms(event.lastTimestamp), event.message]
+            [event.reason,
+             event.type,
+             parse_kubernetes_datetime_to_ms(get_event_timestamp(event)) if get_event_timestamp(event) else "Missing Timestamp",
```
arikalon1, line 21: Did you check this is parsed properly on all sinks?
This column type has a DATETIME renderer, which means the UI (and maybe other sinks) will try to render it as timestamp.
We need to make sure it doesn't break.

src/robusta/core/playbooks/internal/discovery_events.py
```diff
@@ -17,3 +18,59 @@ def cluster_discovery_updates(event: KubernetesAnyChangeEvent):
...
+            k8s_obj.name,
+            FindingSubjectType.from_kind(k8s_obj.kind.lower()),
+            k8s_obj.namespace,
+        ),
+        creation_date= get_event_timestamp(event)
+    )
+    finding.service_key = TopServiceResolver.guess_service_key(name=k8s_obj.name, namespace=k8s_obj.namespace)
+    # non-running services won't be in top service resolver
+    if not finding.service_key:
+        finding.service_key = gen_object_key(k8s_obj)
```
arikalon1, line 71: Service key shouldn't have the pod name.
It can be empty though.
I would either leave it empty, or remove the whole finding

src/robusta/core/playbooks/internal/discovery_events.py
```diff
@@ -17,3 +18,59 @@ def cluster_discovery_updates(event: KubernetesAnyChangeEvent):
...
+        finding = create_event_finding(warning_event)
+        events_table = get_resource_events_table(
+            "Resource events",
+            warning_event.involvedObject.kind,
+            warning_event.involvedObject.name,
+            warning_event.involvedObject.namespace
+        )
+        if events_table:
+            finding.add_enrichment([events_table])
+        event.add_finding(finding)
```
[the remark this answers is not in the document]
kotlickya: not possible to send with no events since there will be at least one warning event to get to this line and the events table will contain that information

src/robusta/core/playbooks/internal/discovery_events.py
kotlickya: for services that are no longer running and jobs it would mean we loose all the data, this way we would be able to connect relevant services in the future.
Additionally these are simmilar parameters passed to and displayed from ServiceInfo to make the service key for events. So i would like to understand how its different here?
`            
          TopServiceResolver.add_cached_service(
            ServiceInfo(
                name=event.obj.metadata.name,
                service_type=event.obj.kind,
                namespace=event.obj.metadata.namespace,
            )
        )
`

src/robusta/core/playbooks/common.py
kotlickya: Screen Shot 2022-03-09 at 15 05 04
in the ui when the string is "Missing timestamp"

src/robusta/core/sinks/robusta/dal/supabase_dal.py
```diff
@@ -282,6 +287,23 @@ def get_active_services(self) -> List[ServiceInfo]:
...
+            .filter("account_id", "eq", self.account_id)
+            .filter("cluster", "eq", self.cluster)
+            .limit(1)
+            .execute()
+        )
+        if res.get("status_code") not in [200]:
+            msg = f"Failed to check cluster issues: {res.get('data')}"
+            logging.error(msg)
+            self.handle_supabase_error()
+            raise Exception(msg)
```
arikalon1, line 303: This is a bit risky.
If there's some error here, we won't get the event history (which might be fine), but we won't start the cluster discovery at all. 
I think we don't want that

src/robusta/runner/web_api.py
```diff
@@ -0,0 +1,28 @@
+import logging
+import time
+
+import requests
+
+
+class WebApi:
+
+    @staticmethod
+    def run_manual_action(action_name: str, params, sinks, retries=1, timeout_delay=1):
```
arikalon1, line 10: can we call it action_params?
Makes it easier to find all usages of action_params.
In addition, I would add type annotation to it.
And default it to None, which is the only value we actually give it for now

src/robusta/runner/web_api.py
```diff
@@ -0,0 +1,28 @@
...
+        MANUAL_ACTION_URL = "http://127.0.0.1:5000/api/trigger"
```
arikalon1, line 11: Please make local variable lower case

src/robusta/runner/web_api.py
```diff
@@ -0,0 +1,28 @@
...
+                MANUAL_ACTION_URL,
+                json=data
+            )
+
+            if response.status_code == 200:
+                return response.status_code
+
+            logging.error(
+                f"Failed to run manual action \naction_name:{action_name}\n"
+                f"params:{params}\nReason: {response.reason}\nStatus Code{response.status_code}"
```
arikalon1, line 24: I wouldn't print the params here, cause it might contain sensitive data.
Or maybe some safe print

src/robusta/runner/web_api.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class WebApi:
+
+    @staticmethod
+    def run_manual_action(action_name: str, params, sinks, retries=1, timeout_delay=1):
+        MANUAL_ACTION_URL = "http://127.0.0.1:5000/api/trigger"
+        data = { "action_name": action_name, "action_params": params, "sinks": sinks }
+        for _ in range(retries):
+            response = requests.post(
```
arikalon1, line 14: Can this throw an exception? If so, what will be the ourcome?

src/robusta/runner/web_api.py
```diff
@@ -0,0 +1,28 @@
...
+            if response.status_code == 200:
+                return response.status_code
+
+            logging.error(
+                f"Failed to run manual action \naction_name:{action_name}\n"
+                f"params:{params}\nReason: {response.reason}\nStatus Code{response.status_code}"
+            )
+            time.sleep(timeout_delay)
+
+        return response.status_code
```
arikalon1, line 28: My ide warns me, that this might not be initialized (I guess if retries is 0)

src/robusta/core/sinks/robusta/robusta_sink.py
```diff
@@ -116,6 +118,25 @@ def __publish_new_services(self, active_services: List):
...
+    def __get_events_history(self):
+        try:
+            # we will need the services in cache before the event history is run to guess service name
+            self.__discover_services()
+            response = WebApi.run_manual_action(
+                action_name="event_history",
+                params=None,
+                sinks=[self.sink_name],
+                retries=4,
+                timeout_delay=5
```
arikalon1, line 130: I think total of 20 seconds might not be enough.
I would choose something bigger

src/robusta/core/sinks/robusta/robusta_sink.py
```diff
@@ -116,6 +118,25 @@ def __publish_new_services(self, active_services: List):
...
+            )
+            if response != 200:
+                logging.info("Error running 'event_history'.")
```
arikalon1, line 133: should probably be error

src/robusta/core/playbooks/common.py
arikalon1: Checkout render_value function (used by other sinks). This is used for rendering tables as strings in Slack etc
Does it fails nicely here as well?

def render_value(renderer: RendererType, value):
    if renderer == RendererType.DATETIME:
        date_value = datetime.fromtimestamp(value / 1000.0)
        return date_value.astimezone(DEFAULT_TIMEZONE).strftime(
            "%b %d, %Y, %I:%M:%S %p"
        )
    raise Exception(f"Unsupported renderer type {renderer}")

src/robusta/core/playbooks/internal/discovery_events.py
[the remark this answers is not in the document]
arikalon1: yes, I know. But we still have that if statement.
If we have it, I think it makes more sense to have the add_finding under it as well

src/robusta/core/playbooks/internal/discovery_events.py
arikalon1: This is just for configuration entities, 
`event.obj.kind in ["Deployment", "ReplicaSet", "DaemonSet", "StatefulSet"]`
Not for jobs/pods which have random names.
The service_key field refers to a top level entity, which is a microservice/Application

src/robusta/core/playbooks/common.py
kotlickya: i got it working with 0

src/robusta/core/playbooks/internal/discovery_events.py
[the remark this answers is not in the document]
kotlickya: kk

src/robusta/core/sinks/robusta/dal/supabase_dal.py
kotlickya: nice catch

src/robusta/core/playbooks/internal/discovery_events.py
arikalon1: I would just add it without one.
We already have such findings

src/robusta/integrations/kubernetes/api_client_utils.py
```diff
@@ -246,5 +246,19 @@ def parse_kubernetes_datetime(k8s_datetime: str) -> datetime.datetime:
     return datetime.datetime.strptime(k8s_datetime, "%Y-%m-%dT%H:%M:%S%z")
 
 
+def parse_kubernetes_datetime_with_ms(k8s_datetime: str) -> datetime.datetime:
```
arikalon1, line 249: I'm not sure I understand the issue, but wouldn't we should just write: (instead of 3 functions)
def parse_kubernetes_datetime_to_ms(k8s_datetime: str) -> float:
    try:
        return parse_kubernetes_datetime_with_ms(k8s_datetime).timestamp() * 1000
    except:
        return parse_kubernetes_datetime(k8s_datetime).timestamp() * 1000

PR #264, opened 2022-03-05
Enable Robusta UI by default in gen-config and handle invalid token format error

arikalon1:
Don't abort on slack channel or ui token errors

### DIFF
--- a/src/robusta/cli/main.py
+++ b/src/robusta/cli/main.py
@@ -160,6 +160,9 @@ def gen_config(
         slack_channel = get_slack_channel()
 
     if slack_api_key and slack_channel:
+        while not verify_slack_channel(slack_api_key, cluster_name, slack_channel, slack_workspace):
+            slack_channel = get_slack_channel()
+
         sinks_config.append(
             SlackSinkConfigWrapper(
                 slack_sink=SlackSinkParams(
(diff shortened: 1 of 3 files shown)


src/robusta/cli/main.py
```diff
@@ -166,9 +169,8 @@ def gen_config(
                 )
             )
         )
-        if not verify_slack_channel(slack_api_key, cluster_name, slack_channel):
-            typer.secho(f"\nInstallation Aborted.", fg=typer.colors.RED)
-            return
+        while not verify_slack_channel(slack_api_key, cluster_name, slack_channel, slack_workspace):
```
aantn, line 172: I think this is a bug. You've already written the slack channel to sinks_config above using the invalid value.

src/robusta/cli/main.py
```diff
@@ -194,13 +196,24 @@ def gen_config(
...
+                        "Please insert your Robusta account token",
+                        default=None,
+                    )
+                    try:
+                        json.loads(base64.b64decode(robusta_api_key))
+                        break
+                    except Exception:
+                        typer.secho(
+                            "Sorry, invalid token format. "
+                            "The token can be found in your generated_values.yaml file, under the robusta_sink",
```
aantn, line 213: Instead of "in your generated_values.yaml file" I think it's more clear to write one of the following (either is fine):

* "in any existing generated_values.yaml file"
* "by running `helm get values robusta`"

aantn (commented):
Left some comments. Other than that looks good.

src/robusta/cli/main.py
arikalon1: Good catch. Fixing

src/robusta/cli/main.py
arikalon1: fixing

PR #267, opened 2022-03-08
Add base telemetry

RoiGlinik:


helm/robusta/templates/runner.yaml
```diff
@@ -29,6 +29,14 @@ spec:
             value: "true"
           - name: ENABLE_COLORED_LOGS
             value: "true"
+          - name: PROMETHEUS_ENABLED
+            value: {{ .Values.enablePrometheusStack | quote}}
+          - name: CLUSTER_ID_HASH
```
arikalon1, line 34: cluster_name and account_id are part of global config, and can change without restart.
We already have it in the runner, might be better to take it from there and hash it

src/robusta/core/model/env_vars.py
```diff
@@ -33,6 +33,7 @@
 PROMETHEUS_REQUEST_TIMEOUT_SECONDS = float(
     os.environ.get("PROMETHEUS_REQUEST_TIMEOUT_SECONDS", 90.0)
 )
+PROMETHEUS_ENABLED = os.environ.get("PROMETHEUS_ENABLED", False)
```
arikalon1, line 36: os.environ.get returns str, and the default value here is bool

Might be safer to do something like:
PROMETHEUS_ENABLED = os.environ.get("PROMETHEUS_ENABLED", "false").lower() == "true"

src/robusta/core/model/env_vars.py
```diff
@@ -54,4 +55,9 @@
 
 ROBUSTA_UI_DOMAIN = os.environ.get("ROBUSTA_UI_DOMAIN", "https://platform.robusta.dev")
 
-SENTRY_TRACES_SAMPLE_RATE = float(os.environ.get("SENTRY_TRACES_SAMPLE_RATE", 0.5))
\ No newline at end of file
+TELEMETRY_ENDPOINT = os.environ.get("TELEMETRY_ENDPOINT", "https://api.robusta.dev/telemetry")
+ENABLE_TELEMETRY = os.environ.get("ENABLE_TELEMETRY", True)
```
arikalon1, line 59: same here

src/robusta/core/playbooks/playbooks_event_handler_impl.py
```diff
@@ -202,7 +202,9 @@ def __get_fired_trigger(
                 return trigger.get()
         return None
 
-    def __handle_findings(self, execution_event: ExecutionBaseEvent):
+    def __handle_findings(self, execution_event: ExecutionBaseEvent): #todo add the code to increase finding count.
```
arikalon1, line 205: Re the comment, isn't it what we do at line 224?

src/robusta/core/sinks/robusta/dal/supabase_dal.py
```diff
@@ -355,10 +355,7 @@ def handle_supabase_error(self):
 
     def to_db_cluster_status(self, data: ClusterStatus) -> Dict[str, Any]:
         db_cluster_status = data.dict()
-        db_cluster_status["updated_at"] = "now()"
-        if data.last_alert_at is not None:
-            db_cluster_status["last_alert_at"] = str(data.last_alert_at)
-            
+        db_cluster_status["updated_at"] = "now()"
```
arikalon1, line 358: We want to add the last_alert_at only if there's a value there, right?
So, in case the runner was restarted, but we haven't received an alert yet, we won't override the old value in the table.
Do we still follow that?

src/robusta/runner/config_loader.py
```diff
@@ -231,6 +231,10 @@ def __reload_playbook_packages(self, change_name):
                 self.registry.set_playbooks(playbooks_registry)
                 self.registry.set_sinks(sinks_registry)
 
+                telemetry = self.registry.get_telemetry()
+                telemetry.sinks_findings_count = { k:0 for (k, _) in sinks_registry.get_all().items() }
```
arikalon1, line 235: We can also use a defaultdict here:
`telemetry.sinks_findings_count = defaultdict(lambda: 0)`

But it will have only sinks which have at least one message

src/robusta/runner/telemetry.py
```diff
@@ -2,5 +2,13 @@
...
+    account_id: str
+    cluster_id: str
     runner_version : str
-    last_alert_at : Optional[datetime]
+    last_alert_at : Optional[str] #ts
+
+    nodes_count: int = 0
+    playbooks_count : int = 0 
+    prometheus_enabled: bool = False
+    sinks_findings_count: dict = dict() # refresh daily, sinks id -> finding count
```
arikalon1, line 13: better to use Dict instead of dict. 
Support for dict was added in later python versions.
We can also initialize to {}

src/robusta/runner/telemetry_service.py
```diff
@@ -0,0 +1,74 @@
...
+        self.periodic_time_sec = periodic_time_sec
+
+        sentry_dsn = os.environ.get("SENTRY_DSN", "")
+        if self.telemetry_level == TelemetryLevel.ERROR and sentry_dsn:
+            logging.info(f"Telemetry set to include error info, Thank you for helping us improve Robusta.")
+            try:
+                sentry_sdk.init(sentry_dsn, traces_sample_rate=float(os.environ.get("SENTRY_TRACES_SAMPLE_RATE", 0.5)))
+            except Exception as e:
+                logging.error(f"Sentry error: {e}")
+                pass
```
arikalon1, line 36: no need for pass

src/robusta/runner/telemetry_service.py
```diff
@@ -0,0 +1,74 @@
...
+        self.registry = registry
+        self.periodic_time_sec = periodic_time_sec
+
+        sentry_dsn = os.environ.get("SENTRY_DSN", "")
+        if self.telemetry_level == TelemetryLevel.ERROR and sentry_dsn:
+            logging.info(f"Telemetry set to include error info, Thank you for helping us improve Robusta.")
+            try:
+                sentry_sdk.init(sentry_dsn, traces_sample_rate=float(os.environ.get("SENTRY_TRACES_SAMPLE_RATE", 0.5)))
+            except Exception as e:
+                logging.error(f"Sentry error: {e}")
```
arikalon1, line 35: would add trace using `exc_info=True`

src/robusta/runner/telemetry_service.py
```diff
@@ -0,0 +1,74 @@
...
+
+        while(True):
+            tele = self.registry.get_telemetry()
+          
+            current_nodes: NodeList = NodeList.listNode().obj
+            tele.nodes_count = len(current_nodes.items)
+
+            self.log(tele)
+
+            tele.sinks_findings_count = {k:0 for k in tele.sinks_findings_count} # periodic data.
```
arikalon1, line 52: consider defaultdict

src/robusta/runner/telemetry_service.py
```diff
@@ -0,0 +1,74 @@
...
+            sleep(self.periodic_time_sec)
+
+
+    def log(self, data: Telemetry):
```
arikalon1, line 56: should this be private?

src/robusta/runner/telemetry_service.py
```diff
@@ -0,0 +1,74 @@
...
+        try:
+            r = requests.post(self.endpoint, data=data.json(), headers={'Content-Type': 'application/json'})
+            if(r.status_code != 201):
+                logging.error(
+                    f"Failed to log telemetry data",
+                    exc_info=True,
```
arikalon1, line 62: no exception here

src/robusta/runner/telemetry_service.py
```diff
@@ -0,0 +1,74 @@
...
+        self.__thread = threading.Thread(target=self.__log_periodic)
+        self.__thread.start()
+
+
+    def __log_periodic(self):
+
+        while(True):
+            tele = self.registry.get_telemetry()
+          
+            current_nodes: NodeList = NodeList.listNode().obj
```
arikalon1, line 47: If an exception is thrown here, the telemetry thread will stop.
I would surround this logic with try/except instead of the one inside the `log` method

src/robusta/core/playbooks/playbooks_event_handler_impl.py
```diff
@@ -218,6 +220,8 @@ def __handle_findings(self, execution_event: ExecutionBaseEvent):
                     sink.write_finding(
                         finding_copy, self.registry.get_sinks().platform_enabled
                     )
+
+                    sinks_findings_dict[sink_name] += 1
```
arikalon1, line 224: Would be nice the capture the sink type as well, because sink_name might hide what it is.
For example, if someone names the sink 'prod' or 'support', we won't know the sink type.
WDYT?

arikalon1:
Nice work.
Left few comments

src/robusta/core/sinks/robusta/dal/supabase_dal.py
RoiGlinik: If the robusta_sink is reconstructed when the runner is restarted,  Then it will send update_cluster_status in the constructor.  It will put null until first alert. (like install scenario)
but, other than that it will behave the same

I changed the type because i think the regular use case wont be needing the datetime methods and I'ts easier than converting it to string in many places.

RoiGlinik:
ready for review again

helm/robusta/values.yaml
```diff
@@ -143,7 +143,8 @@ runner:
   image: us-central1-docker.pkg.dev/genuine-flight-317411/devel/robusta-runner:0.0.0
   imagePullPolicy: IfNotPresent
   log_level: INFO
-  sentry_dsn:
+  sentry_dsn: https://53b627690db14de7b02095407596fa16@o1120648.ingest.sentry.io/6156573
+  sendAdditionalTelemetry: true
```
arikalon1, line 147: I think this should be false by default

src/robusta/core/model/env_vars.py
```diff
@@ -54,4 +55,7 @@
 
 ROBUSTA_UI_DOMAIN = os.environ.get("ROBUSTA_UI_DOMAIN", "https://platform.robusta.dev")
 
-SENTRY_TRACES_SAMPLE_RATE = float(os.environ.get("SENTRY_TRACES_SAMPLE_RATE", 0.5))
\ No newline at end of file
+TELEMETRY_ENDPOINT = os.environ.get("TELEMETRY_ENDPOINT", "https://api.robusta.dev/telemetry")
```
arikalon1, line 58: Following my change re backend profiles, the TELEMETRY_ENDPOINT should be added there as well
see https://github.com/robusta-dev/robusta/pull/270

src/robusta/core/sinks/robusta/dal/supabase_dal.py
arikalon1: I think we don't want that. 
Because on every restart, we will get the "prometheus not connected" notification until the first alert.
I think we should not update that field, unless it has a value which isn't None

src/robusta/runner/config_loader.py
```diff
@@ -231,6 +231,11 @@ def __reload_playbook_packages(self, change_name):
                 self.registry.set_playbooks(playbooks_registry)
                 self.registry.set_sinks(sinks_registry)
 
+                telemetry = self.registry.get_telemetry()
+                telemetry.playbooks_count = len(runner_config.active_playbooks)
+                telemetry.account_id = hashlib.sha256(str(runner_config.global_config.get("account_id")).encode("utf-8")).hexdigest()
```
arikalon1, line 236: I would use some default value, in case account_id/cluster_name isn't configured in global_config

src/robusta/runner/telemetry.py
```diff
@@ -1,6 +1,19 @@
...
 from pydantic.main import BaseModel, Optional
-from datetime import datetime
+from typing import Dict
+
+class SinkInfo(BaseModel):
+    type: str = 'None'
+    findings_count: int = 0
 
 class Telemetry(BaseModel):
+    account_id: str = 'None'
```
arikalon1, line 10: Why do we put that default value here?
I think "None" is a bad default, because it's confusing

arikalon1:
Nice. Added some more comments

src/robusta/runner/telemetry.py
RoiGlinik: good idea. thanks

Issue #266, opened 2022-03-08
Routing messages


zhelyan:
Hi,

thank you for the great tool! 

We might be able to contribute on the feature request below but I wanted to first check if you had any thoughts, design concerns or suggestions:

It would be nice to be able to route the robusta messages similarly to how alertmanger does it.
Here's an example for Slack:

```
- slack_sink:
    name: prod_alerts
    match:
       namespace: prod-*
    slack_channel: prod-issues
 ```

would send messages where `namespace` matches `prod-*` to the `prod-issues` slack channel.
Ideally it should be possible to match against all attributes of the `Finding` [class](https://github.com/robusta-dev/robusta/blob/master/src/robusta/core/reporting/base.py#L60-L65)

thanks!

**Comment by aantn** [comment]
Hey, sounds great to me! We'd love to get a PR for this.

@arikalon1 any notes regarding implementation?

**Comment by arikalon1** [comment]
Thank you for the kind words @zhelyan :)
I like the idea very much as well.

What are your thoughts about the syntax here?
Is it full regex match? only simple wildcard (*)? maybe just contains or simple prefix?
And regarding namespace/kind/name, these are all mapped to the finding subject, right?

**Comment by zhelyan** [comment]
Regex sounds reasonable to me too.
 
I _think_ the general use case here is conditional forwarding, not routing. Routing implies that you have 1 sink and multiple internal destinations but some sinks, e.g datadog don't really work like that - there's one DD account with no option to "route" messages to e.g account members:

```
- datadog_sink:
    name: datadog_sink
    match:
        namespace: prod-.*
    api_key: "datadog api key"
```

arikalon1:
yes, I agree. I think it should probably be a generic filtering layer before sending to each sink

michelesr:
@arikalon1 I have a working implementation that uses the subject finding fields as matchers, see https://github.com/citizensadvice/robusta/blob/master/docs/user-guide/configuration.rst#sink-matchers; I still need to tidy up a little bit but if you like the idea I'll open a PR soon

michelesr:
Actually my implementation currently works only with the fields of the subject, rather than the whole finding attributes. Matching against any arbitrary attribute of the finding is more flexible but requires more care in handling attributes (some of them needs to be converted in the proper string representation before being matched against the regular expression).

arikalon1:
Looks very nice!
Just to make sure, if you specify multiple matchers, the operation is logical AND?
Like, node and namespace?

[...]

PR #268, opened 2022-03-10
Implement regex based filtering for sinks

michelesr:
Implement #266 

Allow to set matchers in the sink configuration to match against the attributes of the finding, for example "namespace".

```
sinksConfig:
  - slack_sink:
      name: main_slack_sink
      slack_channel: alerts
      api_key: mykey
  - slack_sink:
      name: second_slack_sink
      slack_channel: test
      match:
        namespace: test
      api_key: mykey
```


src/robusta/core/reporting/base.py
```diff
@@ -34,6 +35,25 @@ def __str__(self):
...
+        raise NotImplementedError
+
+    def attribute_matches(self, attribute: str, expression: str) -> bool:
+        value = self.attribute_map[attribute]
+        return bool(re.match(expression, value))
+
+    def matches(self, requirements: Dict[str, str]) -> bool:
+        for attribute, expression in requirements.items():
+            if not (
+                attribute in self.attribute_map.keys()
```
arikalon1, line 50: Should we add a warning log, in case the match attribute is illegal? (in case the user had a typo in the sink definition)
It can be here, or a validation when creating the sink.
WDYT?

arikalon1:
Excellent work!
Left one minor comment.

If it's ok with you, we would love to know your names, so we can mention you in the release notes. 
Can we also mention the organization you work for?

src/robusta/core/reporting/base.py
michelesr: Sounds like a good idea!

michelesr:
@arikalon1 I formatted with black, and the last commit picked up some unrelated style changes: do you want me to amend it or leave it like that?

src/robusta/core/reporting/base.py
michelesr: Added the logging in the last commit

michelesr:
> If it's ok with you, we would love to know your names, so we can mention you in the release notes.

It was mainly me, with some help from @zhelyan and @ruthwells, but I'm not sure whether they want to be mentioned in the release notes, maybe they can comment on this conversation if they are happy with it.

> Can we also mention the organization you work for?

Citizens Advice UK

arikalon1:
Thank you @michelesr , it's totally fine.
Merging

PR #269, opened 2022-03-13
fixing bad import

kotlickya:
older versions of pyyaml will block gen-config from working because some of the parameters are missing


aantn:
Looks good to me.

PR #271, opened 2022-03-13
Warning events to debug

Avi-Robusta:


src/robusta/core/playbooks/internal/discovery_events.py
```diff
@@ -55,7 +55,7 @@ def create_event_finding(event: Event):
         title=f"{event.reason} {event.type} for {k8s_obj.kind} {k8s_obj.namespace}/{k8s_obj.name}",
         description=event.message,
         source=FindingSource.KUBERNETES_API_SERVER,
-        severity=FindingSeverity.MEDIUM,
+        severity=FindingSeverity.MEDIUM if event.type.lower() != "warning" else FindingSeverity.DEBUG,
```
arikalon1, line 58: I would just use debug

PR #282, opened 2022-03-27
Custom Graph Enrichers actions

Sheeproid:
Added:
1. node_resource_graph_enricher
2. node_resource_graph_enricher
3. alert_resource_graph_enricher
4. custom_graph_enricher

Using alert_resource_graph_enricher to replace all old graph_enricher (s) in our default values.yaml file

Examples:

robusta playbooks trigger node_resource_graph_enricher name='kind-control-plane' resource_type=Disk

robusta playbooks trigger c name='stress-test-788575bd65-m4fpb' namespace=default resource_type=Memory 

robusta playbooks trigger prometheus_alert alert_name=CPUThrottlingHigh pod_name=robusta-grafana-7dcb4d958b-bg6vt


helm/robusta/values.yaml
```diff
@@ -67,24 +67,35 @@ builtinPlaybooks:
       alert_name: HostHighCpuLoad
   actions:
   - node_cpu_enricher: {}
-  - graph_enricher: {}
+  - alert_resource_graph_enricher:
+      resource_type: CPU
+      item_type: Node
+      graph_duration_minutes: 30
```
arikalon1, line 73: I think 1-2 hours will be better

playbooks/robusta_playbooks/playbooks_utils.py
```diff
@@ -0,0 +1,175 @@
...
+    value_formatters = {
+        ChartValuesFormat.Plain: lambda val: str(val),
+        ChartValuesFormat.Bytes: lambda val: humanize.naturalsize(val, binary=True),
+        ChartValuesFormat.Percentage: lambda val: f'{(100*val):.1f}'
+    }
+    chart_values_format = values_format if values_format else ChartValuesFormat.Plain
+    chart.value_formatter = value_formatters[chart_values_format]
+
+    if chart_title:
+        chart.title = f'{chart_title} starting {humanize.naturaldelta(timedelta(minutes=graph_duration_minutes))} ago'
```
arikalon1, line 92: I'm not crazy about that format.
Can we ask the team what they think the best format is, and get data driven decision?

playbooks/robusta_playbooks/playbooks_utils.py
```diff
@@ -0,0 +1,175 @@
...
+    if not chosen_combination:
+        raise AttributeError(f'The following combination for resource chart is not supported: {combination}')
+    values_format_text = 'Utilization' if chosen_combination.values_format == ChartValuesFormat.Percentage else 'Usage'
+    graph_enrichment = create_graph_enrichment(
+        starts_at,
+        labels,
+        chosen_combination.query,
+        prometheus_url=prometheus_url,
+        graph_duration_minutes=graph_duration_minutes,
+        query_name=f'{resource_type.name} {values_format_text} for this {item_type.name.lower()}',
```
arikalon1, line 172: I would phrase this as: Pod CPU Usage
In addition, I would leave an option to set a custom title for the chart, that comes from the action params.
For example, if I want to write:
OOM Killed pod memory usage
or something ..

playbooks/robusta_playbooks/playbooks_utils.py
```diff
@@ -0,0 +1,175 @@
+from robusta.api import *
```
arikalon1, line 1: Playbooks_utils might be used by other playbooks packages as well, I assume.
So I think it's better not to include it in the default playbooks package. (Other packages don't have access to it)
I would add it as a part of the core code, and add it to robusta.api
This way, it can be imported and used by any playbooks package

src/robusta/core/model/base_params.py
```diff
@@ -47,6 +74,37 @@ class PrometheusParams(ActionParams):
...
+    """
+    :var promql_query: Promql query. See https://prometheus.io/docs/prometheus/latest/querying/basics/
+    :var query_name: A nicer name for the Prometheus query.
+    :var graph_duration_minutes: Graph duration is minutes. Default is 60.
+    :var chart_values_format: one of the ChartValuesFormat. Default is Plain.
+    """
+
+    promql_query: str
+    query_name: Optional[str] = None
+    graph_duration_minutes: Optional[int] = None
```
arikalon1, line 87: I think the default should be defined here

src/robusta/core/model/base_params.py
```diff
@@ -47,6 +74,37 @@ class PrometheusParams(ActionParams):
...
+
+
+class ResourceGraphEnricherParams(PrometheusParams):
+    """
+    :var resource_type: one of ResourceChartResourceType.
+    :var graph_duration_minutes: Graph duration is minutes. Default is 60.
+
+    """
+    resource_type: str
+    graph_duration_minutes: Optional[int] = None
```
arikalon1, line 98: I think the default should be defined here

arikalon1 (commented):
great job!
Left some minor comments.

One small additional note:
Our coding convention is to use double quotes for strings ("..")
I know it's not the case everywhere, but we try to follow it

aantn:
Didn't review the code, but please make sure you update the docs! This probably belongs in either `event-enrichment.rst` or a new page on graph enrichers.

playbooks/robusta_playbooks/playbooks_utils.py
```diff
@@ -0,0 +1,175 @@
...
+        if addr.type == "InternalIP"
+    )
+    return internal_ip
+
+
+def __prepare_promql_query(provided_labels: Dict[Any, Any], promql_query_template: str) -> str:
+    labels = defaultdict(lambda: "<missing>")
+    labels.update(provided_labels)
+    if '$node_internal_ip' in promql_query_template:
+        # TODO: do we already have alert.Node here?
```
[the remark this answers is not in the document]
Sheeproid: I think it is too, however I'm not sure how to recreate this to make sure it actually works

playbooks/robusta_playbooks/alerts_integration.py
```diff
@@ -129,44 +129,50 @@ def graph_enricher(alert: PrometheusKubernetesAlert, params: PrometheusParams):
...
-    result = prom.custom_query_range(
+    chart = playbooks_utils.create_chart_from_prometheus_query(
+        params.prometheus_url,
         promql_query,
-        start_time,
-        end_time,
-        increment,
-        {"timeout": PROMETHEUS_REQUEST_TIMEOUT_SECONDS},
+        alert.alert.startsAt,
+        include_x_axis=False,
```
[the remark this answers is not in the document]
Sheeproid: include_x_axis means should the y axis always start from 0. it kind of gives perspective, for example if the memory fluctuage between 40 to 45 percent it might look very scary if the y axis starts from 40 instead of 0.

playbooks/robusta_playbooks/playbooks_utils.py
Sheeproid: I'll create my own prometheus chart utils

arikalon1:
very nice!

PR #290, opened 2022-04-06
Custom playbook action tutorial

Sheeproid:
This is my attempts at a complete first custom action guide (@arik what I did in Notion a month ago) in our docs format
I think it is a more “complete” read then Writing Playbook Actions section, maybe it can be called a “Quickstart guide” or something (it is not that quick though, and it contains a link to “Custom playbook repositories”)
[Your-first-custom-action.zip](https://github.com/robusta-dev/robusta/files/8424312/Your-first-custom-action.zip)


docs/tutorials/my-first-custom-action.rst
```diff
@@ -0,0 +1,255 @@
...
+
+.. code-block:: bash
+
+    $ kubectl get pods
+
+    NAME                                                     READY   STATUS    RESTARTS   AGE
+
+    nginx                                                    0/1     Pending   0          5h19m
+
+You wait a few minutes, but is remains the same.
```
arikalon1, line 57: Do you think we should mention the Robusta UI here. 
Like, if you're using the Robusta UI you can see it there as well

docs/tutorials/my-first-custom-action.rst
```diff
@@ -0,0 +1,255 @@
...
+        print(f"This print will be shown in the robusta logs={actual_event}")
+
+        if actual_event.type.casefold() == f'Warning'.casefold() and \
+            actual_event.reason.casefold() == f'FailedScheduling'.casefold() and \
+            actual_event.involvedObject.kind.casefold() == f'Pod'.casefold():
+            _report_failed_scheduling(event, actual_event.involvedObject.name, actual_event.message)
+
+    def _report_failed_scheduling(event: EventEvent, pod_name: str, message: str):
+        # this is how you send data to slack or other destinations
+        event.add_enrichment([
```
arikalon1, line 158: I think we need to create a finding here. 
Otherwise, the default finding is used, which has some generic fields

docs/tutorials/my-first-custom-action.rst
```diff
@@ -0,0 +1,255 @@
...
+Before we proceed, we need to enable local playbook repositories in Robusta.
+
+Follow this quick guide to learn how to package your python file for Robusta: :ref:`Custom playbook repositories`
+
+Use this useful debugging commands to make sure your action ( ``report_scheduling_failure``) is loaded:
+
+.. code-block:: bash
+
+    robusta logs # get robusta logs, see errors.
+    robusta playbooks list-dirs  # get see if you custom action was loaded
```
arikalon1, line 171: I think it's more accurate to say: "# get see if you custom action package was loaded"
`list-dirs` display loaded packages I think

docs/tutorials/my-first-custom-action.rst
```diff
@@ -0,0 +1,255 @@
...
+      account_id: XXXX
+    sinksConfig:
+    - slack_sink:
+        name: main_slack_sink
+        slack_channel: '#my-slack-channel'
+        api_key: XXXXX
+    - robusta_sink:
+        name: robusta_ui_sink
+        token: XXXXXX  # generated with `robusta gen-config`
+    clusterName: kind-kind
```
arikalon1, line 202: This is the first time we talk about kind. 
Is another clusterName is better here?

arikalon1, line 202: This is the first time we talk about kind. 
Is another clusterName is better here?

docs/tutorials/my-first-custom-action.rst
```diff
@@ -0,0 +1,255 @@
...
+    enablePrometheusStack: true
```
arikalon1, line 203: Lines 204-215 are relevant only for kind clusters.
Maybe it's easier to follow if only the relevant configuration is present

docs/tutorials/my-first-custom-action.rst
```diff
@@ -0,0 +1,255 @@
...
+
+Great!
+
+Run the scenario from the first section again (creating a bad bad configuration), and you should see this in your slack:
+
+Check our slack channel, and:
+
+.. image:: /images/example_report_scheduling_failure.png
+
+
```
arikalon1, line 251: Should we add cleanup commands. 
Like, how to clean everything we installed for the tutorial

arikalon1:
great great tutorial
left few minor comments

docs/tutorials/my-first-custom-action.rst
Sheeproid: not sure I understand . What do you see in Robusta UI? You won't see the pod (like Lens for example)

docs/tutorials/my-first-custom-action.rst
Sheeproid: what generic fields bug you? I think if we can keep the code short and simple it is best

docs/tutorials/my-first-custom-action.rst
arikalon1: Won't you see the warning event on the pod in the timeline?

docs/tutorials/my-first-custom-action.rst
arikalon1: The finding title would be: "Robusta notification"
The aggregation key (which is used for grouping in the timeline) would be: "Generic finding key"

This would make the example a little longer though:
`   
    finding = Finding(
        title=f"Pod {pod_name} Failed Scheduling",
        aggregation_key="Pod Failed Scheduling",
    )
    finding.add_enrichment([...])
    event.add_finding(finding)
`

docs/tutorials/my-first-custom-action.rst
```diff
@@ -0,0 +1,256 @@
+Your first custom action!
+######################################################
+
+Learn how to automate everything with Robusta, in *15 minutes* or less :)
```
aantn, line 4: Perhaps be more specific than "everything"

e.g: "Learn how to automate _Kubernetes error handling_ with Robusta..." or some other detailed description

docs/tutorials/my-first-custom-action.rst
```diff
@@ -0,0 +1,256 @@
...
+
+Goals
+---------------------------------------
+We’re gonna learn how to create a custom playbook using a short and made-up (but realistic) scenario.
```
aantn, line 8: "going to" not "gonna".

"gonna" is slang

aantn: I would also cut out the unecessary words. E.g:

_We’re going to a custom playbook_ and not
_We’re going learn how to create a custom playbook_

docs/tutorials/my-first-custom-action.rst
```diff
@@ -0,0 +1,256 @@
...
+
+First, we’ll describe the the scenario, and then we’ll implement a Robusta automation (Playbook) that detect this issue immediately and recommends a fix!
```
aantn, line 10: Lower case `p` in `Playbook` and `detects` not `detect`

docs/tutorials/my-first-custom-action.rst
```diff
@@ -0,0 +1,256 @@
...
+
+.. note::
+
+    It is recommended to read :ref:`Example Playbook` before starting this guide.
+
+The scenario
+---------------------------------------
+You want to create a new pod with the image “ngnix”!
+
+Being the smart person that you are, you decide save some time by copying and pasting an existing YAML file you have in your computer, and change the pod name and image to ngnix.
```
aantn, line 22: `changing` not `change`. It needs to match the tense in `copying` and `pasting`

docs/tutorials/my-first-custom-action.rst
```diff
@@ -0,0 +1,256 @@
...
+        spoiler: alert
+
+Now you start the pod by running the following command:
+
+.. code-block:: bash
+
+    $ kubectl apply -f nginx-pod.yaml
+    pod/nginx created
+
+but for some reason it it not starting (note it’s “Pending” Status):
```
aantn, line 47: Best not to start a sentence with `but` or a lowercase letter.

I would write:

_For some reason the pod doesn't start (note it’s “Pending” Status):_

docs/tutorials/my-first-custom-action.rst
```diff
@@ -0,0 +1,256 @@
...
+----
+What we need to do?
+----
+
+.. note::
+    Make sure to clean up the pod from the last section by running ``kubectl delete pod nginx``
+
+A playbook consists of two things:
+
+- Trigger - We’re gonna use a built in trigger
```
aantn, line 112: `Going to` not `gonna` here and in the next line

docs/tutorials/my-first-custom-action.rst
```diff
@@ -0,0 +1,256 @@
...
+- Action - We’re gonna write our own action!
+
+
+Finding the correct trigger
+----
+So what is the correct trigger for the job?
```
aantn, line 118: No need for `So`

docs/tutorials/my-first-custom-action.rst
```diff
@@ -0,0 +1,256 @@
...
+
+- Creation of a new pod (because we create a new pod, ‘ngnix’)
+- A Kubernetes Event is fired (because we ran kubectl get event to find out the scheduling error)
+
+Let’s look at the Trigger section about :ref:`Kubernetes (API Server)`, and try to find out triggers for both.
+Go ahead and try to find them!
+
+Okay! We find ``on_pod_create`` and ``on_event_create``
+
+We’ll use ``on_event_create``, because in this case ``on_pod_create`` is not even called because the pod is not created.
```
aantn, line 129: Are you sure about this? I would expect `on_pod_create` to be called because the Kubernetes pod resource is created... It even shows up in `kubectl get pods`.

The pod's status is pending and the pod was never scheduled, but the pod object itself was still created.

docs/tutorials/my-first-custom-action.rst
```diff
@@ -0,0 +1,256 @@
...
+
+Writing the action
+----
+
+Now we need to write code that checks this event and reports it. To find out the correct event class that matches our trigger ``on_event_create``. please take a look at :ref:`Event Hierarchy`.
```
aantn, line 134: `To find the correct...` not `To find out`...

docs/tutorials/my-first-custom-action.rst
Sheeproid: done

docs/tutorials/my-first-custom-action.rst
Sheeproid: I think it makes it more complicated, so I won't use it.

docs/tutorials/my-first-custom-action.rst
```diff
@@ -0,0 +1,256 @@
...
+    robusta playbooks delete <PLAYBOOK_FOLDER> # remove the playbook we just added from Robusta
+
+    # Remove "customPlaybooks" and "playbooksPersistentVolume" from you config, and then run helm upgrade
+    helm upgrade robusta robusta/robusta --values=generated_values.yaml
+
+
+Summary
+-------------------------------------
+
+We learned how to solve a real problem (pod not scheduling) only once and have Robusta automate it in the future for all our happy co-workers (and future us) to enjoy.
```
aantn, line 257: Should we add something here like:

This example of an unschedulable pod is actually covered by Robusta out of the box (if you enable the builtin Prometheus stack) but you can see how easy it is to track any error you like and send it to a notifications system with extra data.

aantn (commented):
Strongly like. Leaving some comments.

docs/tutorials/my-first-custom-action.rst
arikalon1: Your call obviously.
But I think calling only event.add_enrichment() isn't the correct use of our API.
This approach should be used only when chaining few playbooks using the same finding, like in alerts enrichers.
There, the created finding has meaning.

In general, if one needs to create a finding, he needs to create both the finding and the enrichment.
@aantn your thoughts about this?

aantn: @arikalon1 you're correct (and we should document it somewhere) but for the sake of the tutorial I think either is fine.

docs/tutorials/my-first-custom-action.rst
Sheeproid: When I tried it, it did not work. Maybe we can try together If you'd like. I don't want to change the rest of the tutorial, so maybe we can choose on_event_create for other reason, for example because it contains the information we need. What do you think?

aantn: Sounds good to me. Happy to look at this together at some point.

For now, you can just write something like:

_We'll use `on_event_create` in this tutorial because it will be easier to identify scheduling issues by looking at the event._

PR #292, opened 2022-04-07
added Context flag

### LINKED ISSUE #277: Add --context flag to all robusta cli commands to run them on a given kube config context
E.g. `robusta gen-config --context=other-cluster`

AvineshTripathi:
fixes #277

### DIFF
--- a/src/robusta/cli/main.py
+++ b/src/robusta/cli/main.py
@@ -87,11 +87,16 @@ class HelmValues(BaseModel):
         )
 
 
-def guess_cluster_name():
+def guess_cluster_name(context):
     with click_spinner.spinner():
         try:
             all_contexts, current_context = config.list_kube_config_contexts()
+            if context is not None:
+                for i in range(len(all_contexts)):
+                    if all_contexts[i].get('name') == context:
+                        return context
             if current_context and current_context.get("name"):
+                typer.echo("default")
                 return current_context.get("name")
         except Exception:  # this happens, for example, if you don't have a kubeconfig file
             typer.echo("Error reading kubeconfig to generate cluster name")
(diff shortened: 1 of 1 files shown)


AvineshTripathi:
@kingsathurthi

src/robusta/cli/main.py
```diff
@@ -87,11 +87,16 @@ def set_pod_configs_for_small_clusters(self):
...
 
-def guess_cluster_name():
+def guess_cluster_name(context):
     with click_spinner.spinner():
         try:
             all_contexts, current_context = config.list_kube_config_contexts()
+            if context is not None:
+                for i in range(len(all_contexts)):
+                    if all_contexts[i].get('name') == context:
+                        return context
```
Avi-Robusta, line 97: this returns the context name, not the cluster name (they can be different based on how the context is defined)
for example `kubectl config --kubeconfig=config-demo set-context dev-frontend --cluster=development --namespace=frontend --user=developer`
to return the actual cluster name you would want:
`all_contexts[i].get('context').get('cluster')` 
to return the cluster with that name

src/robusta/cli/main.py
```diff
@@ -87,11 +87,16 @@ def set_pod_configs_for_small_clusters(self):
...
             if current_context and current_context.get("name"):
+                typer.echo("default")
```
Avi-Robusta, line 99: no need to add this log here

src/robusta/cli/main.py
```diff
@@ -87,11 +87,16 @@ def set_pod_configs_for_small_clusters(self):
         )
 
 
-def guess_cluster_name():
+def guess_cluster_name(context):
     with click_spinner.spinner():
         try:
             all_contexts, current_context = config.list_kube_config_contexts()
+            if context is not None:
+                for i in range(len(all_contexts)):
```
Avi-Robusta, line 95: if there is no context with the context name given to the function
i would have typer print something like, no context exists with the name '{context}', your current context is ... 
so that way the user has an idea why the context he inserted didnt display the cluster he wanted

Avi-Robusta:
Great job!!! Needs a few small changes but over all excellent work!

src/robusta/cli/main.py
AvineshTripathi: opps! I was trying to debug so added this that time

src/robusta/cli/main.py
AvineshTripathi: currently for default case without ```--context``` flag we return the ```current_context.get('name')``` so are we suppose to change that as well into returning the name of cluster

Avi-Robusta: yea, it looks like a bug

AvineshTripathi:
done with the fixes @Avi-Robusta

Avi-Robusta:
great job

PR #294, opened 2022-04-17
fixed formatting issues. added logs of all events in case the bug in …

djarv1337:
…https://www.notion.so/ImagePullBackoff-enhancement-c5ab846506f7470cbc83ff4de808fa3e is reproduced in the future, so we can understand why it occurs


arikalon1:
LGTM

PR #295, opened 2022-04-20
renamed, and updated docs

Avi-Robusta:


playbooks/robusta_playbooks/pod_troubleshooting.py
```diff
@@ -297,15 +297,15 @@ def debugger_stack_trace(event: PodEvent, params: DebuggerParams):
 def python_process_inspector(event: PodEvent, params: DebuggerParams):
     """
 
-    Create a finding with alternative debugging options for received processes ; i.e. Stack-trace or Memory-trace.
+    Prompts the user with debugging options for all python processes currently running in the pod.
```
aantn, line 300: Excellent

aantn:
@Avi-Robusta looks good.

Can you also add a screenshot for each playbook to `playbooks/robusta_playbooks/images`? 

The image will automatically be added to the docs if it has the same name as the playbook function.

aantn (changes requested):
Please add the images. Other than that looks good.

aantn:
@Avi-Robusta hmm, sorry to nitpick but after looking at the images I'm slightly confused. Does `python_process_inspector` let you choose between the different debugging options (like we do in `java_process_inspector`) or is it just a wrapper for the stack trace playbook which asks you to choose a process first?

I'd like to keep it as consistent as possible with the other one to avoid confusion.

Avi-Robusta:
currently it is just stack trace but it used to have memory too, i think it was removed because it was determined to be less useful but it can be added again

PR #296, opened 2022-04-22
Related pods action

arikalon1:
Bug fix - Node creation time may change on node upgrade - add it to the diff


src/robusta/core/model/nodes.py
```diff
@@ -27,56 +27,7 @@ def __eq__(self, other):
         if not isinstance(other, NodeInfo):
             return NotImplemented
 
-        ignored_fields = ["deleted", "node_creation_time"]  # node_creation_time never changes
+        ignored_fields = ["deleted"]
```
Avi-Robusta, line 30: any reason why we stopped ignoring this? i think it should be in nodeInfo but i was wondering if we saw the time changes.

src/robusta/core/sinks/robusta/robusta_sink.py
```diff
@@ -191,7 +192,7 @@ def __to_node_info(cls, node: Node) -> Dict:
 
     @classmethod
     def __from_api_server_node(
-        cls, api_server_node: Node, pod_requests: List[PodRequests]
+        cls, api_server_node: Node, pod_requests: List[PodResources]
```
Avi-Robusta, line 195: confusing name for pod_requests list and pod_requests function we import

Avi-Robusta: if we didnt use both on this sinc i would say its no issue, would just rename to pod_requests_list

playbooks/robusta_playbooks/k8s_resource_enrichments.py
```diff
@@ -0,0 +1,52 @@
...
+        return
+
+    if resource.kind == "Pod":
+        pods = [resource]
+    else:
+        selector = build_selector_query(resource.spec.selector)
+        pods = PodList.listNamespacedPod(namespace=resource.metadata.namespace, label_selector=selector).obj.items
+
+    rows = []
+    for pod in pods:
```
Avi-Robusta, line 32: looks good, i personally would wrap everything in a function `generate_related_pod_row` but its fine like this

src/robusta/core/playbooks/playbooks_event_handler_impl.py
```diff
@@ -79,6 +79,10 @@ def run_actions(
```
Avi-Robusta, line 79: unrelated to your code i had a question about line 39
`if execution_event:  # might not exist for unsupported k8s types`
which ones do we not support?

src/robusta/integrations/slack/sender.py
```diff
@@ -189,6 +189,15 @@ def __send_blocks_to_slack(
         )
         other_blocks = [b for b in report_blocks if not isinstance(b, FileBlock)]
 
+        # wide tables aren't displayed properly on slack. looks better in a text file
+        table_blocks = [b for b in other_blocks if isinstance(b, TableBlock)]
+        for table_block in table_blocks:
+            if len(table_block.headers) > 4:
```
Avi-Robusta, line 195: I would make the 4 something like MAX_HEADERS_AMNT_FOR_SLACK_TABLE=4 at the top of the file to make it easier to find and change in the future

Avi-Robusta:
Looks good, nothing critical to change

src/robusta/core/model/nodes.py
arikalon1: Yeah, the time changes when you upgrade the node.
I wrote it in the comment

src/robusta/integrations/slack/sender.py
arikalon1: fixed

src/robusta/core/playbooks/playbooks_event_handler_impl.py
arikalon1: We don't support CronJobs, Secrets, Ingresses
k8s has many resource types. You can run `kubectl api-resources` to see
In addition, we don't support CRDs yet
The list of our supported types is in `generate_kubernetes_code.py`

playbooks/robusta_playbooks/k8s_resource_enrichments.py
arikalon1: fixed

src/robusta/core/sinks/robusta/robusta_sink.py
arikalon1: fixed

PR #298, opened 2022-04-24
Slack feedback message

Sheeproid:
Feedback messages can be scheduled to be sent in Slack after configuring the sink in the CLI.
The configuration of when and what to send in those messages is saved in https://docs.robusta.dev/extra/feedback_messages.json

A sample configuration may look like this:
```
{
    "heads_up_message": "We're going to send a feedback message. We love you.",
    "messages": [
    {
        "minutes_from_now": 1,
        "title": ":raising_hand: Help the open-source improve and grow!",
        "other_sections": [
            "How do you feel about Robusta so far? Do you like it or not?",
            "We value your opinion and experience.",
            "Please help us improve Robusta with this short survey:",
            "<https://dx4xwuqz6wy.typeform.com/to/EGebO4rn#source=slack1&id=$ACCOUNT_ID|Fill Survey>",
            "Thanks! Natan from Robusta"
            ]
        }
    ]
}
```

The message then will be sent to the slack sync:
Screen Shot 2022-04-24 at 16 10 56

The configurable heads up message in the CLI will appear only if a feedback message was scheduled. It will look like this:
Screen Shot 2022-04-24 at 16 12 12

Just to clarify, the messages are scheduled from the cli. We do **not** save the Slack token on our end.

Furthermore, the account_id can be included in the message but if you didn't sign up for the UI then it's just a random ID.

### DIFF
--- a/src/robusta/cli/main.py
+++ b/src/robusta/cli/main.py
@@ -9,6 +9,7 @@ import click_spinner
 from distutils.version import StrictVersion
 from typing import Optional, List, Union, Dict
 from zipfile import ZipFile
+import traceback
 
 import requests
 import typer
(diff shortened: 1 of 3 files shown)


aantn (approved):
Looks good.

PR #301, opened 2022-04-27
update docs related to triggers and events

aantn:
@arikalon1 this still needs work. Please feel free to suggest changes.

The main problem I'm trying to solve is that `PrometheusKubernetesAlert` and `on_prometheus_alert` are under-documented and its not clear how they interact with playbook actions, what they're compatible with, etc.

I don't think I've solved that problem entirely, but that was the motivation for working on this.

### DIFF
--- a/docs/developer-guide/actions/triggers-and-events.rst
+++ b/docs/developer-guide/actions/triggers-and-events.rst
@@ -1,10 +1,10 @@
-Event Hierarchy
+Events and Triggers
 =====================
 
 When configuring Robusta as a user, you define :ref:`triggers <Triggers>` in ``values.yaml`` but when writing playbook
 actions you deal with events.
 
-This page explains the connection between the two.
+The basic idea is that **triggers** generate **event** objects and those events are passed to **action** functions.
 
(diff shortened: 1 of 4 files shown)


docs/developer-guide/actions/triggers-and-events.rst
```diff
@@ -1,33 +1,79 @@
...
     .. inheritance-diagram2:: robusta.api.ExecutionBaseEvent
         :parts: 1
         :include-subclasses:
 
+
+.. note::
+
+    Note that ``PrometheusKubernetesAlert`` inherits from many Kubernetes event classes, so you can use
+    ``on_prometheus_alert`` with many types of playbook actions. For example, ``on_prometheus_alert`` is
+    compatible with any playbook that accepts ``PodEvent``, ``NodeEvent``, ``DeploymentEvent``, and so on.
```
arikalon1, line 75: I would add that when doing it, you need to verify that the alert contains the actual resource.
Because, for example, a Node alert might not have a Pod

arikalon1:
LGTM
Left one minor comment

PR #303, opened 2022-04-30
Support list of exact values in sink matchers

arikalon1:


aantn (approved):
Looks great, but please take the opportunity to make it more obvious from other places in the docs too. E.g. perhaps from the main sinks page link here. We know people have had trouble finding these docs in the past.

PR #304, opened 2022-04-30
Improve gen-config signup for UI

aantn:


aantn:
This PR changes three things. 

1. It fixes the prints during UI signup so they contain color as intended.

2. It reverses the order so we first ask for an email and then an account name. In practice we've seen a number of people enter their email as the account ID so reversing the order will hopefully decrease confusion around this.

3. It modifies the printed text to warn the user about known edge cases

Before:

Screen Shot 2022-04-30 at 12 47 45

After:

Screen Shot 2022-04-30 at 12 47 27


This PR also contains some random autoformatting changes that Black made.

arikalon1:
Good idea.
Look good, no comments

PR #313, opened 2022-05-03
reduce the number of duplicate sphinx labels

aantn:


aantn:
@arikalon1 no need to review - already merged. Just FYI

PR #314, opened 2022-05-03
customise finding action

arikalon1:


playbooks/robusta_playbooks/common_actions.py
```diff
@@ -0,0 +1,28 @@
...
+    :var title: Overriding finding title.
+    :var description: Overriding finding description.
+    :var severity: Overriding finding severity. Allowed values: DEBUG, INFO, LOW, MEDIUM, HIGH
+    """
+
+
+@action
+def customise_finding(event: ExecutionBaseEvent, params: FindingOverrides):
+    """
+    Overrides the created finding attribute with the provided attributes overrides.
```
aantn, line 19: I think this is easier to read as:

Overrides a finding attribute with the provided value

playbooks/robusta_playbooks/common_actions.py
```diff
@@ -0,0 +1,28 @@
...
+
+    This action does not create a new Finding, it just override the attributes of an existing Finding.
```
aantn, line 21: Should be `overrides` (plura) not `override` (singular)

src/robusta/core/model/events.py
```diff
@@ -91,6 +91,16 @@ def add_finding(self, finding: Finding):
 
             self.sink_findings[sink].insert(0, finding)
 
+    def override_finding_attributes(self, title: str = "", description: str = "", severity: FindingSeverity = None):
```
aantn, line 94: I think it's more correct here to pass the title and description as `None` by default not an empty string. Then you can check below `if title is not None` etc.

This way someone can override the description with a blank string if they like.

playbooks/robusta_playbooks/common_actions.py
```diff
@@ -0,0 +1,28 @@
+from robusta.api import *
+
+
+class FindingOverrides(ActionParams):
+    title: str = ""
+    description: str = ""
+    severity: str = ""
```
aantn, line 7: Is it possible to define this as a FindingSeverity? E.g. `severity: Optional[FindingSeverity] = None` or something like that?

docs/catalog/actions/customising.rst
```diff
@@ -0,0 +1,9 @@
+Customising
```
aantn, line 1: I'm not sure this belongs in it's own category. I would rather put it in `Miscellaneous` for now.

If we are going to put it in a separate category, then `Customizing` is confusing IMO. Perhaps we can call it `Message Formatting` or something like that.

playbooks/robusta_playbooks/common_actions.py
aantn: Perhaps add a sentence explaining what Findings are. E.g. `All messages from Robusta are represented as a Finding object. This action lets you override Finding fields to change that messages Robusta sends. This lets you modify messages created by other actions without needing to rewrite those actions.`

aantn (approved):
LGTM. Left minor comments

playbooks/robusta_playbooks/common_actions.py
arikalon1: done

playbooks/robusta_playbooks/common_actions.py
arikalon1: fixed

src/robusta/core/model/events.py
arikalon1: I'll fix as you suggested, but it makes it very fragile and bug prone

playbooks/robusta_playbooks/common_actions.py
arikalon1: no easy way to initialize it by enum key

docs/catalog/actions/customising.rst
arikalon1: I think it will disappear under Miscellaneous
I'll change to Message Formatting

src/robusta/core/model/events.py
arikalon1: while testing it, I saw that using an empty title introduces other issues.
It's not worth the time fixing it

PR #318, opened 2022-05-11
Add services labels and images

arikalon1:


src/robusta/integrations/kubernetes/custom_models.py
```diff
@@ -86,6 +86,27 @@ def extract_images(k8s_obj: HikaruDocumentBase) -> Optional[Dict[str, str]]:
...
+            "containers",
+        ],  # deployment, replica set, daemon set, stateful set, job
+        ["spec", "containers"],  # pod
+    ]
+    images = []
+    for path in containers_paths:
+        try:
+            for container in k8s_obj.object_at_path(path):
+                images.append(container.image)
+        except Exception:  # Path not found on object, not a real error
```
Sheeproid, line 104: Can you catch the exact error instead of all exception?

src/robusta/integrations/kubernetes/custom_models.py
```diff
@@ -86,6 +86,27 @@ def extract_images(k8s_obj: HikaruDocumentBase) -> Optional[Dict[str, str]]:
     return None
 
 
+def extract_image_list(k8s_obj: HikaruDocumentBase) -> List[str]:
```
Sheeproid, line 89: the containers is code duplicated, can you define them in one place?

Sheeproid: actually, maybe make extract_images call extract_image_list and then just run a "map" on the data?

src/robusta/core/sinks/robusta/dal/supabase_dal.py
```diff
@@ -163,6 +165,8 @@ def get_active_services(self) -> List[ServiceInfo]:
                 service_type=service["type"],
                 namespace=service["namespace"],
                 classification=service["classification"],
+                images=service["images"] if service["images"] is not None else [],
```
Sheeproid, line 168: why can it be None? because of older versions? Then you can say instead `service["labels"] if "labels" in service else {}`

Sheeproid: also in that's the case please add a comment that you check this to support older versions of Robusta

src/robusta/core/model/services.py
```diff
@@ -21,4 +24,7 @@ def __eq__(self, other):
             and self.namespace == other.namespace
             and self.classification == other.classification
             and self.deleted == other.deleted
+            and sorted(self.images) == sorted(other.images)
```
Sheeproid, line 27: what about old data that don't contain the images?

Sheeproid: same for the labels

Sheeproid (commented):
looks good, I have some questions because I am not that familiar with the code.

src/robusta/core/sinks/robusta/dal/supabase_dal.py
arikalon1: yes, it's for backward compatibility
For old stored services the "labels" key exists, it's just contains the value None

That's why I explicitly checked if the value is None

src/robusta/core/model/services.py
arikalon1: It works good.
Because on the old data it will have empty array, on the new it will have the image list.
The equals method is used to check if the service changed, and if so it should be persisted to the DB.

So here, an old service will be persisted again with the images

src/robusta/integrations/kubernetes/custom_models.py
arikalon1: There are many type of errors that can be thrown here.
I don't want to fail because I failed to extract the image

src/robusta/integrations/kubernetes/custom_models.py
arikalon1: yes, you're right
Fixing

PR #319, opened 2022-05-12
Update LICENSE

rishavmehra:
change> 2021 to 2022

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2022 Robusta Dev Ltd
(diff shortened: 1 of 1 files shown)


aantn:
Hey, I *think* it should read `2021-2022` instead of a single year `2022`.

At least that's how I interpret https://opensource.stackexchange.com/questions/5778/why-do-licenses-such-as-the-mit-license-specify-a-single-year#:~:text=The%20copyright%20notice%20should%20include,until%201999%2C%20use%201998).

rishavmehra (commented):
New changes

PR #321, opened 2022-05-13
Added a playbook to restart a deployment on certain prometheus alert

NikhilMaheshwari:


docs/catalog/actions/deployment-restart.rst
```diff
@@ -0,0 +1,8 @@
+Restart Based on prometheus alert
+########################
+
+These actions are useful for automatic restart of deployment based on certain prometheus alert.
```
arikalon1, line 4: Can you emphasize that we restart the deployment by the configured name, and not necessarily the deployment related to the alert

playbooks/robusta_playbooks/deployment_restart.py
```diff
@@ -0,0 +1,34 @@
...
+    """
+    :var name: name of deployment that needs to restart.
+    :var namespace: namespace of deployment that needs restart.
+    """
+    name: str
+    namespace: str
+
+
+@action
+def deployment_restart(alert: PrometheusKubernetesAlert, params: DeploymentParams):
```
arikalon1, line 16: Do you mind naming it `deployment_restart_by_name`.
We might add another action in the future, that will do the same but using the deployment in the alert (according to labels). So we want to save a name like this for that one

playbooks/robusta_playbooks/deployment_restart.py
```diff
@@ -0,0 +1,34 @@
...
+    now = datetime.datetime.utcnow()
+    now = str(now.isoformat("T") + "Z")
+    my_deployment: RobustaDeployment = RobustaDeployment.readNamespacedDeployment(params.name, params.namespace).obj
+
+    if not my_deployment:
+        logging.error(
+            f"cannot run deployment_restart on alert with no deployment: {alert}"
+        )
+        return
+    my_deployment.spec.template.metadata.annotations['kubectl.kubernetes.io/restartedAt'] = now
```
arikalon1, line 33: Is kubernetes.io reserved? (See 
 https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#:~:text=The%20kubernetes.io/%20and%20k8s.io/%20prefixes%20are%20reserved%20for%20Kubernetes%20core%20components.)

maybe 'robusta.dev/restartAt' ?

playbooks/robusta_playbooks/deployment_restart.py
```diff
@@ -0,0 +1,34 @@
...
+            "deployment_restart", params.name + params.namespace, params.rate_limit
+    ):
+        return
+    now = datetime.datetime.utcnow()
+    now = str(now.isoformat("T") + "Z")
+    my_deployment: RobustaDeployment = RobustaDeployment.readNamespacedDeployment(params.name, params.namespace).obj
+
+    if not my_deployment:
+        logging.error(
+            f"cannot run deployment_restart on alert with no deployment: {alert}"
```
arikalon1, line 30: This will happen if the deployment doesn't exist.
It might be useful to add the deployment name and namespace to the error log

arikalon1 (commented):
Awesome work. Left a few minor comments, mostly around naming. Code itself is super clean, bug free.

RoiGlinik:
We now support this logic so closing

PR #327, opened 2022-05-18
Adding a log sanitizer for logs_enricher and report_crash_loop using a regex, with a chosen sanitizing format

Sheeproid:


playbooks/robusta_playbooks/alerts_integration.py
```diff
@@ -211,12 +212,23 @@ def template_enricher(alert: PrometheusKubernetesAlert, params: TemplateParams):
...
+
 class LogEnricherParams(ActionParams):
     """
     :var warn_on_missing_label: Send a warning if the alert doesn't have a pod label
+    :var regex_replacer_pattern: a regex pattern to replace text for example for security reasons
+    :var regex_replacement_style: one of SameLengthAsterisks or Redacted (See RegexReplacementStyle)
     """
-
     warn_on_missing_label: bool = False
+    regex_replacer_pattern: Optional[str] = None
```
arikalon1, line 230: I think we might need a list of regex matchers
The customer request said they have IPs, and other PIIs and will want to sanitize all
I assume that they probably able to do it using a single regex, but probably easier to do it with a list of matchers

playbooks/robusta_playbooks/alerts_integration.py
```diff
@@ -238,11 +250,22 @@ def logs_enricher(event: PodEvent, params: LogEnricherParams):
...
-    if log_data:
-        event.add_enrichment(
-            [FileBlock(f"{pod.metadata.name}.log", log_data.encode())],
-        )
+    if not log_data:
+        return
+
+    if params.regex_replacer_pattern:
+        regex_replacement_style = RegexReplacementStyle[params.regex_replacement_style]
+        logging.info('Sanitizing log data with the provided regex pattern')
```
arikalon1, line 260: I would add the filtering regex

arikalon1 (commented):
Nice work!
Left couple of small comments
I think we need to add this functionality to the `report_crash_loop` action (this is the new restart loop reporter action, we don't need it on the deprecated action `restart_loop_reporter` though)

playbooks/robusta_playbooks/alerts_integration.py
Sheeproid: My original code had this list of matchers, even with a functionality that replaces them in a cool way like [IP] for ip, [PORT] for port, etc..
However there are edge cases here, that can happened depending on the order of the regex definition. So I removed it.

playbooks/robusta_playbooks/alerts_integration.py
Sheeproid: My original code had this too, but the filtering regex may contain sensitive information.

playbooks/robusta_playbooks/alerts_integration.py
Sheeproid: I kept a snippet of the original code, I ran it quickly to show you how it looks like:

for
'Hello 123456 xx12Yo\nGoodbye, 007 agent', when there is a regex to replace a number and a text to replace agent names,
It will print:
Hello [NUMBER] xx[NUMBER]Yo
Goodbye, [AGENT NAME]

Sheeproid: we can quickly chat about the edge case on Sunday

arikalon1 (approved):
Looks good

PR #332, opened 2022-05-21
Onboarding improvements

aantn:


src/robusta/cli/integrations_cmd.py
```diff
@@ -58,13 +58,56 @@ def get_slack_key() -> (str, str):
...
+        if not try_again:
+            return ""
+
+
+@app.command()
+def ui():
+    ui_key = get_ui_key()
+    if ui_key:
+        log_title(
+            f"Your UI key is:\n{ui_key}\nAdd it to the Robusta sink configuration"
```
arikalon1, line 108: I would say 'add it as `token` to the Robusta sink configuration'

src/robusta/cli/main.py
```diff
@@ -126,11 +127,7 @@ def write_values_file(output_path: str, values: HelmValues):
     with open(output_path, "w") as output_file:
         yaml.safe_dump(values.dict(exclude_defaults=True), output_file, sort_keys=False)
         typer.secho(
-            f"Saved configuration to {output_path}",
-            fg="green",
-        )
-        typer.secho(
-            f"Save this file for future use. It contains your account credentials",
+            f"Saved configuration to {output_path} - save this file for future use!",
             fg="red",
```
arikalon1, line 131: Why red?
I think it hints there's an error

In all other places we print red only on error

src/robusta/cli/main.py
```diff
@@ -380,14 +319,17 @@ def gen_config(
...
         ]
-    
-
 
     write_values_file(output_path, values)
 
     if robusta_api_key:
         typer.secho(
-            f"Finish the Helm install and then login to Robusta UI at {backend_profile.robusta_ui_domain}\n",
+            f"Finish installing with Helm (see the Robusta docs). Then login to Robusta UI at https://platform.robusta.dev\n",
```
arikalon1, line 327: We need to take the url from the backend profile

arikalon1 (approved):
Looks good
Left few minor comments

src/robusta/cli/main.py
aantn: I tried doing them both in green but this and the next message just merged together visually. I think it's good this way and it's red as it requires action on your part. It's warning you to save the file.

src/robusta/cli/integrations_cmd.py
aantn: I updated it so that it just prints the actual YAML you need. Is that OK?

aantn:
@arikalon1 replied to your comments and pushed an improvement. If it's OK, I'd like to merge.

arikalon1 (approved):

PR #342, opened 2022-05-24
Fix playbook order

Avi-Robusta:


helm/robusta/values.yaml
```diff
@@ -31,6 +31,15 @@ disableCloudRouting: false
 playbooksPersistentVolume: false
 playbooksPersistentVolumeSize: 4Gi
 
+# priority builtin playbooks for running before custom playbooks
```
arikalon1, line 34: I would say:
priority builtin playbooks for running before all playbooks

arikalon1 (approved):
Looks good
Add 2 minor comments

PR #347, opened 2022-05-24
Trim TableBlocks properly for the Slack sink

aantn:
If tables are too long for a single slack message, we should trim them in a way that doesn't break the whole table

Before:
Screen Shot 2022-05-25 at 0 11 23

After:
Screen Shot 2022-05-25 at 0 11 33

### DIFF
--- a/src/robusta/core/reporting/blocks.py
+++ b/src/robusta/core/reporting/blocks.py
@@ -119,7 +119,7 @@ class KubernetesDiffBlock(BaseBlock):
         old: Optional[HikaruDocumentBase],
         new: Optional[HikaruDocumentBase],
         name: str,
-        namespace: str = None
+        namespace: str = None,
     ):
         """
         :param interesting_diffs: parts of the diff to emphasize - some sinks will only show these to save space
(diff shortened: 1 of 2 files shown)


src/robusta/core/reporting/blocks.py
```diff
@@ -230,18 +243,51 @@ def __calc_max_width(cls, headers, rendered_rows, table_max_width: int) -> List[
...
+        lines_to_include = 0
+        for l in lines:
+            new_length = length_so_far + len("\n") + len(l)
+            if new_length > max_chars:
+                break
+            else:
+                length_so_far = new_length
+                lines_to_include += 1
+
+        return "\n".join(lines[:lines_to_include]) + truncator
```
arikalon1, line 268: We add the `truncator` even in the case we didn't remove any line
Should we?

aantn: I don't think so. In that case we already left the function in line 252

arikalon1: got it

arikalon1 (approved):
LGTM

PR #348, opened 2022-05-24
Make the disk enricher show less information by default

aantn:
The disk enricher currently shoves a wall of text at you containing both pod-level and container-level details:

Screen Shot 2022-05-25 at 0 30 10

Note that the above screenshot is actually only part of the data, as the message was so long I couldn't get a full screenshot.

In this PR I've removed the container level details by default (while giving the user the option to re-enable them) which results in a much shorter message that doesn't lose much troubleshooting quality:

Screen Shot 2022-05-25 at 0 30 15

### DIFF
--- a/playbooks/robusta_playbooks/node_disk_analysis.py
+++ b/playbooks/robusta_playbooks/node_disk_analysis.py
@@ -6,8 +6,13 @@ import re
 import humanize
 
 
+class DiskAnalyzerParams(ActionParams):
+    show_pod_level_details: bool = True
+    show_container_level_details: bool = False
+
+
 @action
-def node_disk_analyzer(event: NodeEvent):
+def node_disk_analyzer(event: NodeEvent, params: DiskAnalyzerParams):
     """
     Provides relevant disk information for troubleshooting disk issues.
     Currently, the following information is provided:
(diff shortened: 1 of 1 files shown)


playbooks/robusta_playbooks/node_disk_analysis.py
```diff
@@ -6,8 +6,13 @@
 import humanize
 
 
+class DiskAnalyzerParams(ActionParams):
+    show_pod_level_details: bool = True
```
arikalon1, line 10: I would shorten it to: 
pod_level, container_level
or
pod_level_details, container_level_details

playbooks/robusta_playbooks/node_disk_analysis.py
```diff
@@ -6,8 +6,13 @@
...
+    show_container_level_details: bool = False
+
+
 @action
-def node_disk_analyzer(event: NodeEvent):
+def node_disk_analyzer(event: NodeEvent, params: DiskAnalyzerParams):
```
arikalon1, line 15: Would update the docs about the configuration options, and what's chosen by default

arikalon1 (approved):
Looks good.
Add 2 minor comments

playbooks/robusta_playbooks/node_disk_analysis.py
aantn: @arikalon1 is `show_pods` and `show_containers` OK? I prefer having a verb (`show`) in the setting name.

arikalon1: k

PR #351, opened 2022-05-26
Improve docs for customise_finding action

aantn:
1. Use `on_pod_crash_loop` in the autogenerated example
2. Fix a bug in the docs that caused them to not show up in sphinx


aantn:
For reference, old is below:
Screen Shot 2022-05-26 at 8 52 00
Screen Shot 2022-05-26 at 8 51 50

Here is new:
Screen Shot 2022-05-26 at 8 53 09
Screen Shot 2022-05-25 at 22 02 46

arikalon1 (approved):
nice

PR #352, opened 2022-05-26
Show playbook action names in docs

aantn:
Today if you search for a playbook action in the docs (e.g. `alert_graph_enricher`) then you often find no results because the docs display a "humanized" version of the name (e.g. `Alert Graph Enricher`) and the actual action name never appears in the docs. This means people don't find things when they use the search.

This fixes that.

@arikalon1 FYI

### DIFF
--- a/docs/advanced/robusta-ui-triggers.rst
+++ b/docs/advanced/robusta-ui-triggers.rst
@@ -21,5 +21,6 @@ To use this feature on multiple clusters, make sure you install Robusta with the
     - :code:`globalConfig.account_id`
     - :code:`globalConfig.signing_key`
     - :code:`rsa`
+
 | These values are inside the :code:`values.yaml` file you use when you run :code:`helm install robusta...`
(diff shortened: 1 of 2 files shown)


docs/advanced/robusta-ui-triggers.rst
```diff
@@ -21,5 +21,6 @@ To use this feature on multiple clusters, make sure you install Robusta with the
     - :code:`globalConfig.account_id`
     - :code:`globalConfig.signing_key`
     - :code:`rsa`
+
 | These values are inside the :code:`values.yaml` file you use when you run :code:`helm install robusta...`
-| See `Installation Guide <https://docs.robusta.dev/master/installation.html>`_ for more details.
\ No newline at end of file
+| See `Installation Guide <https://docs.robusta.dev/master/installation.html>`_ for more details.
```
aantn, line 26: FYI this change is unrelated, but removes a sphinx build error due to missing newlines

arikalon1 (approved):
looks good

PR #353, opened 2022-05-26
Update configuration.rst

aantn:
@arikalon1 I tried to clarify some things we got asked about in a call. I don't feel strongly about this change, so let me know if you think it makes things better or worse.


arikalon1 (approved):
looks good

PR #354, opened 2022-05-27
Additional tweek to sink matcher docs

aantn:
Made some additional changes here to try and clarify even more. Thoughts?


aantn:
Whoops, seem to have mixed things up with the PRs. Sorry

PR #355, opened 2022-05-27
More docs about sink matchers

aantn:


aantn:
@arikalon1 added some more docs here about identifier matchers. Wdyt?

arikalon1 (approved):
looks good

PR #357, opened 2022-05-29
use poetry lock --check feature from alpha

RoiGlinik:
Simple action to test the lock file. Uses an alpha version of poetry. 
We will have to update using:
poetry self update --preview

Check is not aligned with older versions.


aantn:
@RoiGlinik when you say "we will have to update" do you mean us developers on our local machines?

RoiGlinik:
@aantn  yea, for running the check command and the sync command ourselfs. 
Also i seem to have a problem with the test

aantn:
@RoiGlinik is the error because they're running an old version of poetry and not installing stuff correctly? Something else?

RoiGlinik:
@aantn  Yea i had to update it at the docker file and the test action as well . now they all use the new version

aantn:
Looks good to me, just make sure you document and announce to everyone that they need to update poetry.

1. I would do a search on the whole project for references to "poetry" and see if we need to update any docs, etc
2. Send a message to everyone in the slack community (in the #general channel) with the exact command they'll need to run to avoid breaking their local setup

PR #366, opened 2022-06-08
Alerts resolution support

arikalon1:
Supporting alert resolution on the Robusta UI
Added 3 new columns: starts_at, updated_at, ends_at - all nullable
Added the alert fingerprint to the finding (this column already exists)
Try to validate the change is indeed backward compatible


src/robusta/core/reporting/base.py
```diff
@@ -94,6 +95,9 @@ def __init__(
         finding_type: FindingType = FindingType.ISSUE,
         failure: bool = True,
         creation_date: str = None,
+        fingerprint: str = None,
+        starts_at: datetime = datetime.now(),
```
aantn, line 99: There is a bug here. Default parameters in Python are evaluated only once, so `starts_at` will always be equal to the time when this function was first imported, not the time of the function call.

You can try this for yourself (see below):
Screen Shot 2022-06-09 at 20 14 30

aantn: The common solution is to set `start_at=None` and in the function itself if it is None then you set to `datetime.now()`

src/robusta/utils/parsing.py
```diff
@@ -13,3 +14,10 @@ def load_json(s: Union[str, bytes]) -> Any:
         sep = "----------------"
         logging.error(f"could not parse json:\n{sep}\n{s}\n{sep}\n")
         raise
+
+
+def to_datetime_str(datetime_obj: datetime.datetime) -> str:
```
aantn, line 19: I would give this a slightly more descriptive name like `datetime_to_db_str`. The comment is very good, but I would make the name a little better.

aantn (approved):
Found a bug (I think). Other than that looks good.

src/robusta/core/reporting/base.py
arikalon1: oops. Good catch!
Fixing

src/robusta/utils/parsing.py
arikalon1: fixing

PR #368, opened 2022-06-10
Fix/incorrect alert names in values yaml

aantn:


arikalon1 (approved):
Nice
Do we know when it was renamed?

aantn:
@arikalon1 no clue. Tried to find it in history and couldn't find the old name for some reason. 🤷

PR #369, opened 2022-06-11
add pods to top-level services and fix possible bug with replicasets

aantn:
note that at least for pods, when there is no owner reference then the value is [] and not None. This might be the case for replicasets too, hence the change there


arikalon1 (approved):
LGTM

PR #372, opened 2022-06-12
fix an exception in the msteams sink

aantn:
Note that there are a number of changes due to autoformatting, but only a single line was actually changed. I'll point out which line in the diff.

### DIFF
--- a/src/robusta/integrations/msteams/msteams_elements/msteams_text_block.py
+++ b/src/robusta/integrations/msteams/msteams_elements/msteams_text_block.py
@@ -3,34 +3,46 @@ from ..msteams_mark_down_fix_url import MsTeamsMarkDownFixUrl
 
 
 class MsTeamsTextBlock(MsTeamsBase):
-    def __init__(self,
-                 text : str,
-                 is_subtle : bool = None,
-                 wrap: bool = None,
-                 weight: str = None,
-                 is_visible : bool = True,
-                 separator : bool = False,
-                 font_size : str = 'medium',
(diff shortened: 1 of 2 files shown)


src/robusta/integrations/msteams/msteams_elements/msteams_text_block.py
```diff
@@ -3,34 +3,46 @@
...
+                is_visible,
+                separator,
+                font_size,
+                horizontal_alignment,
+            )
+        )
 
     def __to_msteams_text(self, text: str) -> str:
-        teams_text = MsTeamsMarkDownFixUrl().fix_text(text)
+        teams_text = MsTeamsMarkDownFixUrl().fix_text(str(text))
```
aantn, line 31: This is the actual fix. Sometimes text was a number and it caused an exception down the call stack.

src/robusta/integrations/msteams/msteams_mark_down_fix_url.py
```diff
@@ -1,26 +1,23 @@
-
+import logging
```
arikalon1, line 1: Why did you add it?

arikalon1 (approved):
Looks good.
Left 1 minor comment

src/robusta/integrations/msteams/msteams_mark_down_fix_url.py
aantn: Whoops, fixed it. (It was from an earlier version when I was debugging the issue)

PR #373, opened 2022-06-13
Update event-enrichment.rst

aantn:
Today the autogenerated docs for the `logs_enricher` action aren't great.

Screen Shot 2022-06-13 at 18 35 38

While technically the `logs_enricher` action is compatible with the `on_pod_delete` trigger, it wouldn't make much sense to configure and it probably wouldn't work because the pod was deleted.

This tiny PR improves the autogenerated docs by explicitly instructing our Sphinx extension to use the `on_pod_crash_loop` trigger in the examples.


arikalon1 (approved):
nice

PR #374, opened 2022-06-14
refactor and improve event enrichments + improve default playbooks

aantn:
@arikalon1 Assigning to you, but feel free to delegate to someone else.

A few notes on the PR:

1. I've moved playbooks around a little. I feel strongly that it's better to organize playbooks into files by logical topics (e.g. playbooks related to kubernetes events) rather than by kubernetes resources (e.g. playbooks that act on pods) because the former puts related playbooks next to one another and the latter puts totally unrelated playbooks in the same file. But if someone objects to this, I can undo that organizational change.

2. I'm not entirely happy with the fact that we have so many separate playbooks here - pod_events_enricher, deployment_events_enricher, and event_resource_events as opposed to one master playbook `events_enricher`. @arikalon1 we discussed a little the current limitations that lead to this (namely we can't write a generic playbook at the moment that works well with both `on_kubernetes_any_resource_*` and `on_prometheus_alert`) so I'm proceeding as is. But I'm open to suggestions for improving this now or down the road.


playbooks/robusta_playbooks/event_enrichments.py
```diff
@@ -0,0 +1,128 @@
...
+        )
+        return
+    obj = event.obj.involvedObject
+    events_table = get_resource_events_table(
+        f"*Related Events*",
+        obj.kind,
+        obj.name,
+        obj.namespace,
+        included_types=params.included_types,
+        max_events=params.max_events,
```
arikalon1, line 61: Before this change we would bring all events.
This playbook only sends it to the UI by default.
Why are we limiting it?

src/robusta/core/playbooks/common.py
```diff
@@ -1,34 +1,74 @@
...
 from ..reporting.custom_rendering import RendererType
 from ...integrations.kubernetes.api_client_utils import parse_kubernetes_datetime_to_ms
 
 
-def get_resource_events_table(table_name: str, kind: str, name: str, namespace: str = None) -> Optional[TableBlock]:
-    field_selector = f"involvedObject.kind={kind},involvedObject.name={name}"
+def filter_event(
+    ev: Event, name_substring_filter: str, included_types: Optional[List[str]]
+) -> bool:
+    if name_substring_filter not in ev.involvedObject.name:
```
arikalon1, line 12: I didn't see any such usage, but if this is called with None, what would be the result?
Should we be extra careful and handle it?

playbooks/robusta_playbooks/event_enrichments.py
```diff
@@ -0,0 +1,128 @@
...
+        )
+        return
+
+    if params.dependent_pod_mode:
+        events_table_block = get_resource_events_table(
+            "*Related pod events:*",
+            "Pod",
+            None,
+            dep.metadata.namespace,
+            name_substring=dep.metadata.name,
```
arikalon1, line 113: Since we're using the deployment name, and a pod type, I think this will match all the deployment pods together.
I'm afraid it might be spammy (ignoring the max_events for a sec), with a deployment with many pods.

With the replicas mismatch, with many replicas, chances are we won't get the problematic pod events, right?

That might also be the cause for the duplications

Maybe we should just get the matching non-running pods, and get their events?

arikalon1 (approved):
Nice work!
Added some comments

playbooks/robusta_playbooks/event_enrichments.py
aantn: Yeah, we can undo it. I think you're right

playbooks/robusta_playbooks/event_enrichments.py
arikalon1: I think there will be cases where all pod crashes, and cases where only some (like with lack of enough resources, or affinity limitations etc)

Maybe we need a little more logic to decide what to show

arikalon1: Maybe get related non-running pods, and choose max 2 pods, to display the pod events

src/robusta/core/playbooks/common.py
aantn: Sure.

arikalon1: fixed

playbooks/robusta_playbooks/event_enrichments.py
arikalon1: fixed

playbooks/robusta_playbooks/event_enrichments.py
arikalon1: fixed

PR #375, opened 2022-06-14
By default, only trigger on_prometheus_alert for firing alerts

aantn:
This changes the default behaviour for `on_prometheus_alert` so that it only triggers for currently firing alerts, *unless configured otherwise*.

### Rationale
When configuring playbooks with `on_prometheus_alert`, most users don't want to run the action on resolved alerts. The previous default behave led to surprising behaviour. We know this because we received questions from users on why their playbook actions were failing and it was because they were unknowingly running the actions on resolved alerts. The same happened to me when I did some work on Prometheus alerts recently. 

### Docs
This change will need to be communicated in the release notes. I suggest the following text:

**Breaking: `on_prometheus_alert` will no longer trigger by default on resolved alerts. This can be overriden if you need. See the [docs](https://docs.robusta.dev/master/catalog/triggers/prometheus.html) for details**

### Other
This PR also includes a tiny fix for Sphinx documentation. The trigger classes didn't inherit from DocumentedModel so the following output was *not* shown properly in the docs:

Screen Shot 2022-06-14 at 4 02 11

Now it's fixed.

### DIFF
--- a/helm/robusta/values.yaml
+++ b/helm/robusta/values.yaml
@@ -35,7 +35,8 @@ playbooksPersistentVolumeSize: 4Gi
 priorityBuiltinPlaybooks:
 # playbooks for prometheus silencing
 - triggers:
-  - on_prometheus_alert: {}
+  - on_prometheus_alert:
+      status: "all"
   actions:
   - name_silencer:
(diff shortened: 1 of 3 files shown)


aantn:
CC @Sheeproid as I know @arikalon1 discussed this with you too.

PR #377, opened 2022-06-14
changeable namespace

RoiGlinik:


aantn:
Re-assigning to @arikalon1!
I unfortunately don't have the time right now

RoiGlinik:
its not a high priority. also i forgot a testing file there i will remote it

helm/robusta/kubewatch.yaml
```diff
@@ -1,4 +1,3 @@
-namespace: ""
```
arikalon1, line 1: Why did you remove it?
What was the purpose of that?

helm/robusta/templates/auth-config.yaml
```diff
@@ -3,6 +3,7 @@ apiVersion: v1
 kind: Secret
 metadata:
   name: robusta-auth-config-secret
+  namespace: {{ .Release.Namespace }}
```
arikalon1, line 6: Does .Release.Namespace always has a value?
It's either the current namespace, or the one you pass to the helm command?

helm/robusta/kubewatch.yaml
RoiGlinik: i moved it to the kubewatch-configmap.yaml file so i can use the release value, the content of the file was used for the config map.

helm/robusta/templates/auth-config.yaml
RoiGlinik: from my test its set to default when i dont use  something.

PR #378, opened 2022-06-15
Install default playbooks at build time in the dockerfile

aantn:
This also updates the typer dependency to fix https://github.com/tiangolo/typer/issues/377


arikalon1 (approved):
LGTM

PR #384, opened 2022-06-18
Add poetry-core as dependency

SamAlex0808:
* Add poetry-core as dependency to enable support default build-system


SamAlex0808:
@aantn FYI

aantn:
@SamAlex0808 regarding the lock file, I'll fix it. It's likely because we're using a newer beta version of Poetry.

No need to do anything, updating it now.

PR #386, opened 2022-06-18
Warning event trigger exclude

arikalon1:
Added exclude and include filters to warning_events_trigger


src/robusta/core/triggers/error_event_trigger.py
```diff
@@ -52,6 +58,17 @@ def should_fire(self, event: TriggerEvent, playbook_id: str):
...
 
+        event_content = f"{exec_event.obj.reason}{exec_event.obj.message}".lower()
+        # exclude if any of the exclusions is found in the event content
+        for exclusion in self.exclude:
+            if exclusion.lower() in event_content:
+                return False
+
+        # exclude if any of the inclusions is NOT found in the event content
+        for inclusion in self.include:
+            if inclusion.lower() not in event_content:
```
Sheeproid, line 69: something here is confusing for me.

if I want to include two events: include: ["ImagePullBackOff", "NodeSysctlChange"] - then it will never work, because they BOTH have to be inside the event_content

Sheeproid (changes requested):
please look at the comment

src/robusta/core/triggers/error_event_trigger.py
arikalon1: You're right.
Fixing it.
Nice catch!

Sheeproid (approved):

PR #391, opened 2022-06-20
Fix docstring for PodCrashLoopTrigger

aantn:
@arikalon1 note that for the multiline string to be extracted by sphinx as a docstring, it needs to appear directly after the class definition. It can't appear after other stuff (see what I'm fixing if this doesn't make sense)


src/robusta/core/triggers/pod_crash_loop_trigger.py
```diff
@@ -1,29 +1,32 @@
...
 
 
 class PodCrashLoopTrigger(PodUpdateTrigger):
-    rate_limit: int = 3600
-    restart_reason: str = None
-    restart_count: int = 2
-
     """
-    :var rate_limit: Limit firing to once every `rate_limit` seconds 
+    :var rate_limit: Limit firing to once every `rate_limit` seconds
```
aantn, line 13: Moving this docstring to the top of the class is the only real change here. Everything else is just black autoformating.

arikalon1 (approved):

PR #403, opened 2022-06-28
add checksum and add some change to the config

RoiGlinik:


RoiGlinik:
@arikalon1  FYI,

arikalon1:
@RoiGlinik why do we need the checksum?

RoiGlinik:
@arikalon1 , It makes sure that the forwarder is restared if the configmap is changed. For example, if you upgrade it will be restared and reread the config map with the changes.

PR #406, opened 2022-07-01
Enhancement of docs

Shubh28698:
Hi @aantn ,

I have added some cool emojis before headings in docs. It's my attempt to make robusta docs more cool & attractive, So that user likes to read it & not get bombarded with text only. If this modifications are not fine, No worries!

Thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,10 +39,10 @@
   </ol>
 </details>
 
-##💻 About the project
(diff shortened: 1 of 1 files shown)


arikalon1 (approved):
Looks great @Shubh28698 
Thanks for the contribution!

PR #408, opened 2022-07-02
Minor fixes for sink errors

aantn:


aantn:
Nothing urgent here, but I was going through some local changes and found this. It fixes various errors I encountered in edge cases when there was a blank svg file.

arikalon1 (approved):
nice

PR #410, opened 2022-07-02
Fix babysitter messages for creations and deletions

aantn:
This should fix the misleading message on the timeline for creations/deletions (see below):

Screen Shot 2022-03-24 at 17 12 12
m

I haven't tested this properly. Please do not merge without testing creations, deletions, and modifications and seeing that each one shows up properly.


arikalon1 (approved):
LGTM

PR #423, opened 2022-07-09
Alert handling job

arikalon1:


playbooks/robusta_playbooks/job_actions.py
```diff
@@ -0,0 +1,92 @@
...
+
+    job.create()
+
+    if params.notify:
+        event.add_enrichment([
+            MarkdownBlock(f"Alert handling job *{job_name}* created.")
+        ])
+
+
+def alert_env_vars(event: PrometheusKubernetesAlert) -> List[EnvVar]:
```
Sheeproid, line 79: I think usually python private functions starts with an underline, like _alert_env_vars or __alert_env_vars

Sheeproid: I'd change the name to get_alert_env_vars

playbooks/robusta_playbooks/job_actions.py
```diff
@@ -0,0 +1,92 @@
...
+                        )
+                    ],
+                    restartPolicy=params.restart_policy
+                )
+            ),
+        )
+    )
+
+    if params.job_ttl_after_finished:
+        job.spec.ttlSecondsAfterFinished = params.job_ttl_after_finished
```
Sheeproid, line 69: ttlSecondsAfterFinished can be None (it is optional) so you can skip the "if" and do it on the spot when you define the JobSpec, I think

playbooks/robusta_playbooks/job_actions.py
```diff
@@ -0,0 +1,92 @@
...
+    :var namespace: The created job namespace.
+    :var restart_policy: Job container restart policy
+    :var job_ttl_after_finished: Delete finished job ttl (seconds). If omitted, jobs will not be deleted automatically.
+    :var notify: Add a notification for creating the job.
+
+    :example command: ["perl",  "-Mbignum=bpi", "-wle", "print bpi(2000)"]
+    """
+    image: str
+    command: List[str]
+    name: str = "robusta-action-job"
```
Sheeproid, line 18: I think that requiring an explicit name from the user is much preferred (removing the default "robusta-action-job"). Will make it searchable in the values file and will make sure that the user understands he creates a job.

playbooks/robusta_playbooks/job_actions.py
```diff
@@ -0,0 +1,92 @@
...
+
+    ALERT_NAME
+
+    ALERT_STATUS
+
+    ALERT_OBJ_KIND - oneof pod/deployment/node/job/daemonset or None in case it's unknown
+
+    ALERT_OBJ_NAME
+
+    ALERT_OBJ_NAMESPACE (Optional)
```
Sheeproid, line 40: I think that instead of **(Optional)** it is better to say something like **(If present)**

playbooks/robusta_playbooks/job_actions.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+@action
+def alert_handling_job(event: PrometheusKubernetesAlert, params: JobParams):
+    """
+    Create a kubernetes job with the specified parameters
+
+    In addition, the job pod receives the following alert parameters as environment variables
+
+    ALERT_NAME
```
Sheeproid, line 32: silly question - did you verified that the job actually gets all the env vars?

playbooks/robusta_playbooks/job_actions.py
```diff
@@ -0,0 +1,92 @@
...
+    command: List[str]
+    name: str = "robusta-action-job"
+    namespace: str = "default"
+    restart_policy: str = "OnFailure"
+    job_ttl_after_finished: int = None
+    notify: bool = False
+
+
+@action
+def alert_handling_job(event: PrometheusKubernetesAlert, params: JobParams):
```
Sheeproid, line 26: Do you think we should also add a non-alert related job creation action?

src/robusta/integrations/prometheus/models.py
```diff
@@ -109,7 +109,7 @@ def get_description(self) -> str:
             )
         return clean_description
 
-    def __get_alert_subject(self) -> FindingSubject:
+    def get_alert_subject(self) -> FindingSubject:
```
Sheeproid, line 112: same comment about __ prefix but in reverse. why removing it? it is a private function

playbooks/robusta_playbooks/job_actions.py
```diff
@@ -0,0 +1,92 @@
...
+
+class JobParams(ActionParams):
+    """
+    :var image: The job image.
+    :var command: The job command as array of strings
+    :var name: Custom name for the job and job container.
+    :var namespace: The created job namespace.
+    :var restart_policy: Job container restart policy
+    :var job_ttl_after_finished: Delete finished job ttl (seconds). If omitted, jobs will not be deleted automatically.
+    :var notify: Add a notification for creating the job.
```
Sheeproid, line 12: you are missing the "backoffLimit" parameter form the example at https://kubernetes.io/docs/concepts/workloads/controllers/job/#running-an-example-job.

Also more parameters are missing (e.g activeDeadlineSeconds). I don't know how important they are to the OSS users and specifically for the one that opened the Issue.

Sheeproid (commented):
Nice work. See comments.

playbooks/robusta_playbooks/job_actions.py
arikalon1: fixing

playbooks/robusta_playbooks/job_actions.py
arikalon1: I rather leave the default name, so the user can have less parameters to configure

playbooks/robusta_playbooks/job_actions.py
arikalon1: fixing

playbooks/robusta_playbooks/job_actions.py
arikalon1: yes, I used that job that prints all the env vars to stdout

src/robusta/integrations/prometheus/models.py
arikalon1: I'm now calling it from the new playbook, so it can't be private

playbooks/robusta_playbooks/job_actions.py
arikalon1: I wouldn't add it in advance.
If someone will ask, we can surely add it

playbooks/robusta_playbooks/job_actions.py
arikalon1: I re-checked, and you're right.
Fixed

playbooks/robusta_playbooks/job_actions.py
arikalon1: I'm not sure as well how important these are, but I'll add those 2, because I can think of some alert handling jobs these will be required.

playbooks/robusta_playbooks/job_actions.py
```diff
@@ -0,0 +1,98 @@
...
+        metadata=ObjectMeta(
+            name=job_name,
+            namespace=params.namespace
+        ),
+        spec=JobSpec(
+            template=PodTemplateSpec(
+                spec=PodSpec(
+                    containers=[
+                        Container(
+                            name=params.name,
```
SamAlex0808, line 62: Even here `serviceAccountName` matters right?

arikalon1: yes, it does matter.
Would you like to use the runner's service account, or provide one using the action parameters? Or both?

PR #427, opened 2022-07-11
Oomkill trigger and enricher

Avi-Robusta:


Avi-Robusta:
This is how it looks
Screen Shot 2022-07-11 at 16 27 45

arikalon1:
I have a few comments
Let's have a short call and discuss it

playbooks/robusta_playbooks/oom_killer.py
```diff
@@ -8,6 +8,8 @@
     K8sMemoryTransformer,
 )
 
+from .alerts_integration import LogEnricherParams
```
Avi-Robusta, line 11: base params move

playbooks/robusta_playbooks/oom_killer.py
```diff
@@ -31,6 +33,82 @@ class OomKillerEnricherParams(ActionParams):
...
+    event: PodEvent,
+    params: LogEnricherParams
+):
+    pod = event.get_pod()
+    node: Node = Node.readNode(pod.spec.nodeName).obj
+    if not node:
+        logging.warning(
+            f"Node {pod.spec.nodeName} not found for OOMKilled for pod {pod.metadata.name}"
+        )
+        return
```
Avi-Robusta, line 47: create finding without node info

playbooks/robusta_playbooks/oom_killer.py
Avi-Robusta: remove

helm/robusta/values.yaml
```diff
@@ -58,6 +58,18 @@ builtinPlaybooks:
...
 
+# playbooks for non-prometheus based monitoring that use prometheus for enrichment
+- triggers:
+  - on_pod_oom_killed: {}
+  actions:
+  - pod_oom_killer_enricher: {}
+  - pod_graph_enricher:
+      resource_type: Memory
+      display_limits: true
+  - node_graph_enricher_for_pod:
```
arikalon1, line 69: I would name this pod_node_graph_enricher

playbooks/robusta_playbooks/node_enrichments.py
```diff
@@ -1,4 +1,5 @@
 from robusta.api import *
+from robusta.core.playbooks.prometheus_enrichment_utils import XAxisLine
```
arikalon1, line 2: If you need to import something from the runner code, add it to robusta.api

We don't want to assume it's the same code base

playbooks/robusta_playbooks/node_enrichments.py
```diff
@@ -135,10 +136,14 @@ def node_graph_enricher(node_event: NodeEvent, params: ResourceGraphEnricherPara
...
     """
-    start_at = datetime.now()
-    labels = {'node': node_event.get_node().metadata.name}
-
     node = node_event.get_node()
+    graph_enrichment = create_node_graph_enrichment(params, node)
+    node_event.add_enrichment([graph_enrichment])
+
+
+def create_node_graph_enrichment(params: ResourceGraphEnricherParams, node: Node):
```
arikalon1, line 144: Maybe it would be better to add it in prometheus_enrichment_utils, like create_graph_enrichment

playbooks/robusta_playbooks/oom_killer.py
```diff
@@ -31,6 +31,65 @@ class OomKillerEnricherParams(ActionParams):
 NODE_MEMORY_THRESHOLD = 0.95
 
 
+@action
+def pod_oom_killer_enricher(
+    event: PodEvent,
+):
+    pod = event.get_pod()
```
arikalon1, line 38: Just to be on the safe side, I would add a check if the pod object exists.
This action requires PodEvent.
Alert is a pod event, but might not have a pod object for some alerts

playbooks/robusta_playbooks/oom_killer.py
```diff
@@ -31,6 +31,65 @@ class OomKillerEnricherParams(ActionParams):
...
+            oom_killed_status = container_status.lastState.terminated
+
+    labels = [("Pod", pod.metadata.name), ("Namespace", pod.metadata.namespace),
+              ("Memory limit", "None" if not resource_limits.memory else f"{resource_limits.memory}MB"),
+              ("Memory requests", "None" if not resource_requests.memory else f"{resource_requests.memory}MB"),
+              ("Node Name", pod.spec.nodeName),
+              ]
+    node: Node = Node.readNode(pod.spec.nodeName).obj
+    if node:
+        logging.warning(
```
arikalon1, line 63: I think this should be on the else, if the node is not found

playbooks/robusta_playbooks/oom_killer.py
```diff
@@ -31,6 +31,65 @@ class OomKillerEnricherParams(ActionParams):
...
+            f"Node {pod.spec.nodeName} not found for OOMKilled pod {pod.metadata.name}"
+        )
+        allocatable_memory = PodResources.parse_mem(node.status.allocatable.get("memory", "0Mi"))
+        capacity_memory = PodResources.parse_mem(node.status.capacity.get("memory", "0Mi"))
+        node_labels = [("Node memory allocatable (free)", f"{allocatable_memory}MB"),
```
arikalon1, line 68: I would remove the "(free)" 
Allocatable isn't free. It's capacity - reserved for k8s

playbooks/robusta_playbooks/oom_killer.py
```diff
@@ -31,6 +31,65 @@ class OomKillerEnricherParams(ActionParams):
...
+              ("Node memory capacity", f"{capacity_memory}MB"),
+              ("Node memory precent requested", f"{(capacity_memory - allocatable_memory) * 100 / capacity_memory}%")]
```
arikalon1, line 70: Maybe this would be better phrased as allocated

playbooks/robusta_playbooks/oom_killer.py
```diff
@@ -31,6 +31,65 @@ class OomKillerEnricherParams(ActionParams):
...
+
+    resource_requests = pod_requests(pod)
+    resource_limits = pod_limits(pod)
+
+    finding = Finding(
+        title=f"Pod {pod.metadata.name} in namespace {pod.metadata.namespace} OOMKilled results",
+        aggregation_key="pod_oom_killer_enricher",
+        severity=FindingSeverity.HIGH
+    )
+    logs_blocks: List[BaseBlock] = []
```
arikalon1, line 48: It doesn't have any logs.
Maybe just blocks?

playbooks/robusta_playbooks/oom_killer.py
```diff
@@ -31,6 +31,65 @@ class OomKillerEnricherParams(ActionParams):
...
+    oom_killed_status = None
+    for container_status in pod.status.containerStatuses:
```
arikalon1, line 50: checkout get_oom_kills_from_pod
It does a lot of similar work

It has for example: get_oom_killed_state
You could probably use it

In addition, the code there takes only recent oom kills.
Can there be a situation that a container has old oomkills?

playbooks/robusta_playbooks/oom_killer.py
```diff
@@ -31,6 +31,65 @@ class OomKillerEnricherParams(ActionParams):
...
+              ("Node memory capacity", f"{capacity_memory}MB"),
+              ("Node memory precent requested", f"{(capacity_memory - allocatable_memory) * 100 / capacity_memory}%")]
+        labels.extend(node_labels)
+
+    if not oom_killed_status:
+        logging.error(
+            f"could not find OOMKilled status in pod {pod.metadata.name}"
+        )
+    else:
+        labels.append(("Container started at",oom_killed_status.startedAt))
```
arikalon1, line 78: startsAt and finishedAt are both optional
Should we check if it actually exists?

playbooks/robusta_playbooks/pod_enrichments.py
```diff
@@ -2,9 +2,11 @@
 
 from datetime import datetime
 
+from robusta.core.playbooks.prometheus_enrichment_utils import XAxisLine
```
arikalon1, line 5: We don't want to assume it's the same code base as the runner.
If we need something from core, we need to add it to the robusta.api

playbooks/robusta_playbooks/pod_enrichments.py
```diff
@@ -13,12 +15,39 @@ def pod_graph_enricher(pod_event: PodEvent, params: ResourceGraphEnricherParams)
         "pod": pod_event.get_pod().metadata.name,
         "namespace": pod_event.get_pod().metadata.namespace,
     }
+    pod = pod_event.get_pod()
```
arikalon1, line 18: I would add it at the beginning of the function, and use it in the lines above as well

playbooks/robusta_playbooks/pod_enrichments.py
```diff
@@ -13,12 +15,39 @@ def pod_graph_enricher(pod_event: PodEvent, params: ResourceGraphEnricherParams)
...
+    limit_lines = []
+    resource_limits = pod_limits(pod)
```
arikalon1, line 20: Would it make more sense to do it:

if params.display_limits:
  resource_limits = pod_limits
  if params.resource_type == "CPU" ...

playbooks/robusta_playbooks/pod_enrichments.py
```diff
@@ -13,12 +15,39 @@ def pod_graph_enricher(pod_event: PodEvent, params: ResourceGraphEnricherParams)
...
     )
     pod_event.add_enrichment([graph_enrichment])
+
+
+@action
+def node_graph_enricher_for_pod(pod_event: PodEvent, params: ResourceGraphEnricherParams):
+    """
+    Get a graph of a specific resource for the node the pod resides on.
+    """
+    pod = pod_event.get_pod()
```
arikalon1, line 45: This can be triggered by alert.
So we need to verify the alert actually has a pod

src/robusta/core/playbooks/prometheus_enrichment_utils.py
```diff
@@ -82,13 +87,23 @@ def create_chart_from_prometheus_query(
...
+    min_time = None
+    max_time = None
     for series in result:
         label = "\n".join([v for v in series["metric"].values()])
         values = [
             (timestamp, round(float(val), FLOAT_PRECISION_LIMIT))
             for (timestamp, val) in series["values"]
         ]
+        times = [timestamp for (timestamp, _) in series["values"]]
+        if not min_time or min(times) < min_time:
```
arikalon1, line 99: I would initialize min_time to big number, and max_time to 0
then:
min_time = min(min_time, min(times))
max_time = max(max_time, max(times))

no "ifs" is better imo

src/robusta/core/triggers/pod_oom_killed_trigger.py
```diff
@@ -0,0 +1,74 @@
...
+from ...integrations.kubernetes.base_triggers import K8sTriggerEvent
+from ...utils.rate_limiter import RateLimiter
+
+
+class PodOOMKilledTrigger(PodUpdateTrigger):
+    """
+    :var rate_limit: Limit firing to once every `rate_limit` seconds
+    """
+
+    rate_limit: int = 3600
```
arikalon1, line 15: Maybe on oom kills, the default should be to report on all oom killed.
So maybe the rate limit should be defaulted to 0

src/robusta/core/triggers/pod_oom_killed_trigger.py
```diff
@@ -0,0 +1,74 @@
...
+        )
+        namespace = pod.metadata.namespace
+        return RateLimiter.mark_and_test(
+            f"PodOOMKilledTrigger_{playbook_id}",
+            namespace + ":" + name,
+            self.rate_limit,
+        )
+
+    @staticmethod
+    def is_oom_killed_state(container_state):
```
arikalon1, line 70: I think this is duplicated with get_oom_killed_state, or at least very similar.
We have 3 places we're doing the same logic

arikalon1 (commented):
Nice work
Added few comments

playbooks/robusta_playbooks/pod_enrichments.py
Avi-Robusta: do you mean:
```
if params.display_limits:
     resource_limits = pod_limits
     if params.resource_type == "CPU" and resource_limits.cpu > 0:
```

arikalon1: yeap

playbooks/robusta_playbooks/node_enrichments.py
Avi-Robusta: I get it since `pod_enrichments` uses it but personally since this is node specific and this is just a small wrapper function I feel like its makes more sense location wise in the `node_enrichments`.

playbooks/robusta_playbooks/oom_killer.py
Avi-Robusta: Got it,
Is it possible to have an OOMKill event without a pod object? 
if so then i need to change this enricher to support that.

playbooks/robusta_playbooks/oom_killer.py
Avi-Robusta: yea, changing it to get most recent oomkill

arikalon1 (approved):
Great work!

PR #429, opened 2022-07-12
can generate auth token with a signing key in env var

Sheeproid:


src/robusta/cli/utils.py
```diff
@@ -57,6 +57,26 @@ def exec_in_robusta_runner(
     return subprocess.check_call(cmd)
 
 
+def exec_in_robusta_runner_output(command, namespace) -> Optional[bytes]:
+    exec_cmd = [
```
arikalon1, line 61: to avoid code duplication, I would extract the command building to one function (lines 61-73).

arikalon1 (approved):
Looks good.
Added 2 minor comments

PR #434, opened 2022-07-18
fail robusta installation when no sinks are defined

Avi-Robusta:
also slackApiKey and robustaApiKey for backward compatibility


Avi-Robusta:
the values slackApiKey and robustaApiKey can be used to define a sink in older versions so we support them in this check too

Sheeproid (approved):

PR #442, opened 2022-07-23
add error codes for different errors

arikalon1:


src/robusta/integrations/receiver.py
```diff
@@ -189,7 +200,7 @@ def on_open(self, ws):
         )
         ws.send(json.dumps(open_payload))
 
-    def __validate_request(self, action_request: ExternalActionRequest, validate_timestamp: bool) -> bool:
+    def __validate_request(self, action_request: ExternalActionRequest, validate_timestamp: bool) -> ValidationResponse:
```
Sheeproid, line 203: if makes more sense to me that validate request will throw an exception in case it fails (and return None), as most of the time the request should be valid. "ValidationResponse" looks like an exception class.

src/robusta/integrations/receiver.py
```diff
@@ -36,6 +37,12 @@
 )
 
 
+class ValidationResponse(BaseModel):
+    http_code: int = 200
+    error_code: Optional[int] = None
```
Sheeproid, line 42: error code-> error message looks like a python enum

src/robusta/core/playbooks/playbooks_event_handler_impl.py
```diff
@@ -189,12 +189,12 @@ def __run_playbook_actions(
                 not registered_action
             ):  # Might happen if manually trying to trigger incorrect action
                 msg = f"action {action.action_name} not found. Skipping for event {type(execution_event)}"
-                execution_event.response = self.__error_resp(msg)
+                execution_event.response = self.__error_resp(msg, 4600)
```
Sheeproid, line 192: makes more sense to me to throw an exception when an error happens

Sheeproid (approved):
style comments about exception and enum

src/robusta/integrations/receiver.py
```diff
@@ -209,50 +220,53 @@ def __validate_request(self, action_request: ExternalActionRequest, validate_tim
...
 
         # First auth protocol option, based on signature only
         signature = action_request.signature
         if signature:
             generated_signature = sign_action_request(body, signing_key)
-            return hmac.compare_digest(generated_signature, signature)
+            if hmac.compare_digest(generated_signature, signature):
+                return ValidationResponse()
+            else:
+                return ValidationResponse(http_code=500, error_code=4502,  error_msg="Signature missmatch")
```
Sheeproid, line 238: type in "missmatch", should be "mismatch"

Sheeproid (approved):

src/robusta/integrations/receiver.py
```diff
@@ -209,50 +220,53 @@ def __validate_request(self, action_request: ExternalActionRequest, validate_tim
...
 
         # Second auth protocol option, based on public key
         partial_auth_a = action_request.partial_auth_a
         partial_auth_b = action_request.partial_auth_b
         if not partial_auth_a or not partial_auth_b:
             logging.error(f"Insufficient authentication data. Cannot verify request {body}")
-            return False
+            return ValidationResponse(http_code=500, error_code=4503, error_msg="Missing auth input")
```
Sheeproid, line 245: are those human readable message suppose to be 100% clear to users? Maybe add a section in the docs on how to respond to this errors. The link to the docs can appear in the UI as well.

src/robusta/integrations/receiver.py
arikalon1: I want to return different permutations of http_code and error_code
I think a return value is better here

src/robusta/core/playbooks/playbooks_event_handler_impl.py
arikalon1: same

src/robusta/integrations/receiver.py
arikalon1: fixing

src/robusta/integrations/receiver.py
arikalon1: These are internal errors. 
I'm not sure the user has anything to do but contact us.
This error for example, doesn't have anything to do with what the user is doing

src/robusta/integrations/receiver.py
arikalon1: you're right, changing to enum

PR #446, opened 2022-07-26
add demo/ folder with examples

aantn:


aantn:
@arikalon1 note that I didn't document all the demos yet. There is still some work to do here.

demo/README.md
```diff
@@ -0,0 +1,55 @@
...
+1. You can get Slack notifications about issues in your cluster 
+2. Notifications contain the exact data you need to debug
+3. You see more troubleshooting data in the Robusta UI, including recent changes to Kubernetes objects
+
+# OOM Kill Demo
+
+## Running it
+
+1. Run `kubectl apply -f ./oomkill_demo/pod_oomkill.yaml`
+2. Check your Slack channel for a notification about the crashing pod. The message includes details you need to debug the OOM
```
arikalon1, line 26: Is this the crashing pod or oom killed pod?

demo/custom_playbook.yaml
```diff
@@ -0,0 +1,11 @@
+
```
arikalon1, line 1: Is this file used anywhere here?

arikalon1 (approved):
Nice work
Left 2 minor comments

demo/README.md
aantn: What do you mean?

demo/custom_playbook.yaml
aantn: No, removed it.

demo/README.md
arikalon1: It's oom kill demo, and the notification should be oom kill enricher, right?
The text says "for a notification about the crashing pod"
Should this be "for a notification about the oom killed pod" ?

PR #449, opened 2022-07-27
Alert subject mapping

arikalon1:


aantn (approved):
LGTM. As discussed, lets test extensively.

PR #454, opened 2022-07-30
add page explaining about the saas ui and comparing it with OSS

aantn:
As soon as this is merged, I also need to update the link from the FAQ at home.robusta.dev (we moved the page in the docs, so the old link needs to be fixed)


arikalon1 (approved):
LGTM

PR #460, opened 2022-08-01
Feature/discord sink

anFatum:


src/robusta/integrations/discord/sender.py
```diff
@@ -0,0 +1,281 @@
+import requests
+from typing import Tuple
+from ...core.reporting.base import *
+from ...core.reporting.blocks import *
+from ...core.reporting.utils import add_pngs_for_all_svgs
+from ...core.model.env_vars import DISCORD_TABLE_COLUMNS_LIMIT, ROBUSTA_LOGO_URL
+
+ACTION_TRIGGER_PLAYBOOK = "trigger_playbook"
```
arikalon1, line 8: I think this is not used

src/robusta/core/reporting/blocks.py
```diff
@@ -283,8 +283,9 @@ def __to_strings_rows(cls, rows):
         # This is just to assert all row column values are strings. Tabulate might fail on other types
         return [list(map(lambda column_value: str(column_value), row)) for row in rows]
 
-    def to_markdown(self, max_chars=None) -> MarkdownBlock:
+    def to_markdown(self, max_chars=None, apply_table_name: bool = True) -> MarkdownBlock:
```
arikalon1, line 286: Do you think `add_table_header` would be a better name to the parameter?

src/robusta/integrations/discord/sender.py
```diff
@@ -0,0 +1,281 @@
...
+    FindingSeverity.LOW: ":yellow_circle:",
+    FindingSeverity.INFO: ":green_circle:",
+}
+SEVERITY_COLOR_MAP = {
+    FindingSeverity.HIGH: "14495556",
+    FindingSeverity.MEDIUM: "16027661",
+    FindingSeverity.LOW: "16632664",
+    FindingSeverity.INFO: "7909721",
+}
+MAX_BLOCK_CHARS = 2000
```
arikalon1, line 22: Are these Discord's API limits?

src/robusta/integrations/discord/sender.py
```diff
@@ -0,0 +1,281 @@
...
+            self, url: str, cluster_name: str
+    ):
+        """
+        Set the Discord webhook url.
+        """
+        self.url = url
+        self.cluster_name = cluster_name
+
+    @classmethod
+    def __add_log_level_icon(cls, title: str, severity: FindingSeverity) -> str:
```
arikalon1, line 61: Do you think `__add_severity_icon` would be a better name here?

src/robusta/integrations/discord/sender.py
```diff
@@ -0,0 +1,281 @@
...
+        icon = SEVERITY_EMOJI_MAP.get(severity, "")
+        return f"{icon} {severity.name} - {title}"
+
+    @staticmethod
+    def apply_length_limit(msg: str, max_length: int = MAX_FIELD_CHARS):
```
arikalon1, line 66: We have an identical function in slack.
Can we extract it to a common location instead of having the same function twice?

src/robusta/integrations/discord/sender.py
```diff
@@ -0,0 +1,281 @@
...
+        if len(msg) <= max_length:
+            return msg
+        truncator = "..."
+        return msg[: max_length - len(truncator)] + truncator
+
+    @staticmethod
+    def __extract_markdown_name(block: MarkdownBlock):
```
arikalon1, line 73: Can you add a return type here?
What is the purpose of this function?
Why are we using the BLACK_CHAR everywhere and not just ""?

src/robusta/integrations/discord/sender.py
```diff
@@ -0,0 +1,281 @@
...
+        text = block.text
+        regex = re.compile(r"\*.+\*")
+        match = re.match(regex, block.text)
+        if match:
+            title = text[match.span()[0]:match.span()[1]]
+            text = text[match.span()[1]:]
+        return title, DiscordSender.__transform_markdown_links(text) or BLANK_CHAR
+
+    @staticmethod
+    def __transform_markdown_links(text: str):
```
arikalon1, line 84: Is this the same as `to_github_markdown` on transformer.py ?

docs/catalog/sinks/discord.rst
```diff
@@ -0,0 +1,51 @@
...
+      :align: center
+
+3. Select Integrations.
+4. If there are no existing webhooks, select Create Webhook. Otherwise, select View Webhooks then New Webhook.
+
+    .. image:: /images/discord_create_webhook.png
+      :width: 1000
+      :align: center
+
+5. Optional. You can enter the name of the bot and the avatar to post the message, but Robusta will change them if allowed.
```
arikalon1, line 29: Do we really enable that?
I didn't see a way to configure it in the sink.
And when sending the message, the name seems to always be `Robusta`

src/robusta/integrations/discord/sender.py
```diff
@@ -0,0 +1,281 @@
...
+            cropped_url = find.replace("<", "").replace(">", "")
+            url, url_text = cropped_url.split("|")
+            new_url = f"[{url_text}]({url})"
+            text = text.replace(find, new_url)
+        return text
+
+    @staticmethod
+    def __format_final_message(discord_blocks: List[DiscordBlock],
+                               header_block: DiscordBlock, msg_color: Union[str, int]) -> Dict:
+        fields = [block for block in discord_blocks if 'value' in block]
```
arikalon1, line 97: Is this check is to check if it's a DiscordFieldBlock?
If so, it would be better to use isinstance(block, DiscordFieldBlock)

src/robusta/integrations/discord/sender.py
```diff
@@ -0,0 +1,281 @@
...
+        discord_msg = {
+            "username": "Robusta",
+            "avatar_url": ROBUSTA_LOGO_URL,
+            "embeds": [
+                *[_add_color_to_block(block, msg_color) for block in discord_blocks if "description" in block]
```
arikalon1, line 102: Same here, just for DiscordDescriptionBlock

src/robusta/integrations/discord/sender.py
```diff
@@ -0,0 +1,281 @@
...
+            ],
+            **_add_color_to_block(header_block, msg_color)
+        }
+        if fields:
+            discord_msg['embeds'].append({"fields": fields, "color": msg_color})
```
arikalon1, line 107: We try to always use double quotes everywhere (")

src/robusta/integrations/discord/sender.py
```diff
@@ -0,0 +1,281 @@
...
+                    name=block.table_name,
+                    value=block.to_markdown(max_chars=MAX_BLOCK_CHARS, apply_table_name=False).text
+                ), sink_name
+            )
+        elif isinstance(block, ListBlock):
+            return self.__to_discord(block.to_markdown(), sink_name)
+        elif isinstance(block, KubernetesDiffBlock):
+            return self.__to_discord_diff(block, sink_name)
+        else:
+            logging.error(
```
arikalon1, line 177: I would change it to warning.
There are few blocks that we don't support here, but can still be used, like JsonBlock and CallbackBlock

src/robusta/integrations/discord/sender.py
```diff
@@ -0,0 +1,281 @@
...
+            f"discord_msg: {discord_msg}\n"
+            f"attachment_blocks: {attachment_blocks}\n"
+        )
+
+        try:
+            response = requests.post(self.url, json=discord_msg)
+            response.raise_for_status()
+            if attachment_blocks:
+                response = requests.post(self.url, data={
+                    "username": discord_msg['username'],
```
arikalon1, line 224: please use double quotes

src/robusta/core/sinks/discord/discord_sink_params.py
```diff
@@ -0,0 +1,15 @@
+from ..sink_config import SinkConfigBase
+from ..sink_base_params import SinkBaseParams
+
+
+class DiscordSinkParams(SinkBaseParams):
+    url: str
+    size_limit: int = 2000
```
arikalon1, line 7: Is this something configurable, or it's a Discord limitation?

src/robusta/integrations/discord/sender.py
arikalon1: Ok, after playing with it a little I understood the purpose of this function...

arikalon1 (approved):
Great work!
Left few minor comments

Can you please add to the docs page of the sink, and image that shows how it looks on Discord?
you can use this one:
Screen Shot 2022-08-06 at 5 36 42 PM

src/robusta/integrations/discord/sender.py
anFatum: The Discord API does not allow us to send the empty string, so this is kind of workaround for that

src/robusta/integrations/discord/sender.py
anFatum: They are all converted into DiscordBlock in __to_discord method, so here isinstance will fail (as both DiscordField and Description blocks are inherited from BaseBlock and are not the instances of DiscordBlock) 
That's why I'm checking by key present in DiscordBlock - that is a dictionary basically

src/robusta/integrations/discord/sender.py
arikalon1: thx

src/robusta/integrations/discord/sender.py
arikalon1: I think this is a little bug prone, if someone will add another block with 'value' or 'description' field
Can we pass the original blocks into the `__format_final_message` function, and then, do the __to_dicsord inside?

It will allow us to check using isInstance
WDYT?

anFatum: Now it should be possible to check type of block using `isinstance` method

PR #461, opened 2022-08-01
fix duplicated enrichments

arikalon1:
don't load kube-state-metrics alert resources

### DIFF
--- a/playbooks/robusta_playbooks/pod_troubleshooting.py
+++ b/playbooks/robusta_playbooks/pod_troubleshooting.py
@@ -164,6 +164,7 @@ def python_memory(event: PodEvent, params: MemoryTraceParams):
         finding, "Profile", python_memory, python_process_inspector
     )
     if process is None:
+        event.add_finding(finding)
         return
 
     cmd = f"debug-toolkit memory --seconds={params.seconds} {process.pid}"
@@ -319,7 +320,7 @@ def python_process_inspector(event: PodEvent, params: DebuggerParams):
         ERROR_MESSAGE = f"No relevant processes found for advanced debugging."
         logging.info(ERROR_MESSAGE)
         finding.add_enrichment([MarkdownBlock(ERROR_MESSAGE)])
-    elif params.interactive:
+    else:
         finding.add_enrichment(
             [MarkdownBlock(f"Please select an advanced debugging choice:")]
         )
--- a/playbooks/robusta_playbooks/pod_troubleshooting.py
+++ b/playbooks/robusta_playbooks/pod_troubleshooting.py
@@ -159,7 +159,6 @@ def python_memory(event: PodEvent, params: MemoryTraceParams):
         finding_type=FindingType.REPORT,
         failure=False,
     )
-    event.add_finding(finding)
     process_finder = ProcessFinder(pod, params, ProcessType.PYTHON)
     process = process_finder.get_match_or_report_error(
         finding, "Profile", python_memory, python_process_inspector
@@ -187,6 +186,7 @@ def python_memory(event: PodEvent, params: MemoryTraceParams):
         MarkdownBlock(f"*Other unfreed memory:* {snapshot.other_data.to_markdown()}")
     )
     finding.add_enrichment(blocks, annotations={SlackAnnotations.ATTACHMENT: True})
+    event.add_finding(finding)
 
 
 class DebuggerParams(ProcessParams):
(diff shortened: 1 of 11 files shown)


src/robusta/core/reporting/base.py
```diff
@@ -147,6 +147,7 @@ def __init__(
         self.fingerprint = fingerprint
         self.starts_at = starts_at if starts_at else datetime.now()
         self.ends_at = ends_at
+        self.dirty = False
```
aantn, line 150: As far as I can tell, you never set this to `True`

src/robusta/core/playbooks/internal/discovery_events.py
```diff
@@ -45,7 +45,7 @@ def event_history(event: ExecutionBaseEvent):
         )
         if events_table:
             finding.add_enrichment([events_table])
-            event.add_finding(finding)
+            event.add_finding(finding, True)
```
aantn, line 48: Why is this supression necessary here?

aantn (approved):
Nice work. Lets make sure we do extensive testing on a demo cluster before releasing. It might also make sense to test a few different setups like a single Slack sink and multiple Slack sinks.

I think you're missing a call to `event.add_finding()` in the following places:

1. `java_pod_troubleshooting.py` line 40 and line 111
2. `deployment_status_report.py` line 54

There might be others - I started going over every call to `add_enrichment` in the code but didn't finish. I decided it was best to submit the review anyway. If you can, do another check to see if anything is missing.

src/robusta/integrations/kubernetes/process_utils.py
```diff
@@ -116,12 +116,14 @@ def __get_matches(
     def __get_error_blocks(
         self, processes: List[Process], text: str, action: Callable, debug_action: Callable
     ) -> List[BaseBlock]:
-        blocks = [
-            TableBlock(
+        blocks = []
```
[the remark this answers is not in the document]
arikalon1: Adding a markdown for that

playbooks/robusta_playbooks/pod_troubleshooting.py
```diff
@@ -187,6 +186,7 @@ def python_memory(event: PodEvent, params: MemoryTraceParams):
         MarkdownBlock(f"*Other unfreed memory:* {snapshot.other_data.to_markdown()}")
     )
     finding.add_enrichment(blocks, annotations={SlackAnnotations.ATTACHMENT: True})
+    event.add_finding(finding)
```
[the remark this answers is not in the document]
arikalon1: yes, you're right.
The code here is a little obfuscated

src/robusta/core/reporting/base.py
arikalon1: You're right.
Adding it to event.add_finding

src/robusta/core/playbooks/internal/discovery_events.py
arikalon1: This is the get events history scenario.
We're creating multiple findings in this playbook, we're reading all warning events, and create a new finding for each.
That's ok in this scenario.
I've seen this warning for all new customers, on the first installation. this suppression hides it

playbooks/robusta_playbooks/pod_troubleshooting.py
```diff
@@ -312,20 +312,17 @@ def python_process_inspector(event: PodEvent, params: DebuggerParams):
...
         ERROR_MESSAGE = f"No relevant processes found for advanced debugging."
         logging.info(ERROR_MESSAGE)
         finding.add_enrichment([MarkdownBlock(ERROR_MESSAGE)])
-        return
-
-    finding.add_enrichment(
-        [MarkdownBlock(f"Please select an advanced debugging choice:")]
-    )
-    if params.interactive:
+    elif params.interactive:
```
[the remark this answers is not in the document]
arikalon1: I'm changing it. This code is too obfuscated.

It does hurt the user, in case he's using a sink that doesn't support callbacks.
he will see just:
"Please select an advanced debugging choice:"
"*After clicking a button please wait up to 120 seconds for a response*"

The choices clocks aren't there

arikalon1:
> Nice work. Lets make sure we do extensive testing on a demo cluster before releasing. It might also make sense to test a few different setups like a single Slack sink and multiple Slack sinks.
> 
> I think you're missing a call to `event.add_finding()` in the following places:
> 
> 1. `java_pod_troubleshooting.py` line 40 and line 111
> 2. `deployment_status_report.py` line 54
> 
> There might be others - I started going over every call to `add_enrichment` in the code but didn't finish. I decided it was best to submit the review anyway. If you can, do another check to see if anything is missing.

I fixed those, but it's unrelated to this change 
Thanks for the great review

PR #466, opened 2022-08-03
Use k8s python client on discovery

arikalon1:
Fix bug on Telemetry - endless loop with no sleep on exception


src/robusta/core/sinks/robusta/robusta_sink.py
```diff
@@ -147,30 +136,55 @@ def __get_events_history(self):
...
+        return ServiceInfo(
+            name=meta.name,
+            namespace=meta.namespace,
+            service_type=kind,
+            images=images,
+            labels=meta.labels or {},
+        )
+
     def __discover_services(self):
         try:
```
Sheeproid, line 150: The fact that K8ss python client is faster then Hikaru is knows to us but not clear to the reader. Just add a comment in the K8s client util file, __discover_services and __discover_nodes that says why we use this instead of Hikaru

src/robusta/core/discovery/utils.py
```diff
@@ -0,0 +1,59 @@
+from typing import List
+from kubernetes.client import V1Deployment, V1DaemonSet, V1StatefulSet, V1Job, V1Pod, V1ResourceRequirements
+
+from ...core.model.pods import PodResources, ResourceAttributes
+
+# This file contains utility related to k8s python api objects (rather than hikaru)
+
```
Sheeproid, line 7: I think that instead of calling it util.py, it is better to just make it a Discovery class and then move _create_service_info,  discover_services, __to_node_info, and discover_nodes here. (__publish_new_services and __publish_new_nodes should stay where they are)

src/robusta/core/sinks/robusta/robusta_sink.py
```diff
@@ -214,13 +228,14 @@ def __from_api_server_node(
             address for address in addresses if "internalip" in address.type.lower()
         ]
         internal_ip = ",".join([addr.address for addr in internal_addresses])
+        node_taints = api_server_node.spec.taints or []
         taints = ",".join(
-            [cls.__to_taint_str(taint) for taint in api_server_node.spec.taints]
+            [cls.__to_taint_str(taint) for taint in node_taints]
```
Sheeproid, line 233: I see .__to_taint_str still get a Hikaru Taint. does it work?

Sheeproid: (same for __to_active_conditions_str)

src/robusta/core/discovery/utils.py
```diff
@@ -0,0 +1,59 @@
...
+                or isinstance(resource, V1DaemonSet) \
+                or isinstance(resource, V1StatefulSet) \
+                or isinstance(resource, V1Job):
+            containers = resource.spec.template.spec.containers
+        elif isinstance(resource, V1Pod):
+            containers = resource.spec.containers
+
+        return [container.image for container in containers]
+    except Exception:
+        pass
```
Sheeproid, line 24: please add a comment here as well of why it can fail

( In general, why not add logs in those cases?)

src/robusta/core/discovery/utils.py
```diff
@@ -0,0 +1,59 @@
...
+    return []
+
+
+def k8s_pod_requests(pod: V1Pod) -> PodResources:
+    """Extract requests from k8s python api pod (not hikaru)"""
+    return pod_resources(pod, ResourceAttributes.requests)
+
+
+def k8s_pod_limits(pod: V1Pod) -> PodResources:
```
Sheeproid, line 33: could not find a use for this function for some reason

src/robusta/core/discovery/utils.py
```diff
@@ -0,0 +1,59 @@
...
+    """Extract limits from k8s python api pod (not hikaru)"""
+    return pod_resources(pod, ResourceAttributes.limits)
+
+
+def pod_resources(pod: V1Pod, resource_attribute: ResourceAttributes) -> PodResources:
```
Sheeproid, line 38: this function can be private

src/robusta/core/discovery/utils.py
```diff
@@ -0,0 +1,59 @@
...
+            if resources:
+                resource_spec = getattr(resources, resource_attribute.name) or {}  # requests or limits
+                pod_cpu_req += PodResources.parse_cpu(
+                    resource_spec.get("cpu", 0.0)
+                )
+                pod_mem_req += PodResources.parse_mem(
+                    resource_spec.get("memory", "0Mi")
+                )
+        except Exception:
+            pass  # no requests on container
```
Sheeproid, line 53: I think throwing exception in a part of a happy flow/non edge case is bad practice. I'm not sure why this exception is not preventable you can just check if the attribute exists by code like you already did?

Sheeproid (approved):
nice work

src/robusta/core/sinks/robusta/robusta_sink.py
arikalon1: fixing

src/robusta/core/discovery/utils.py
arikalon1: Fixing

src/robusta/core/sinks/robusta/robusta_sink.py
arikalon1: Probably a bug. Nice catch

src/robusta/core/discovery/utils.py
arikalon1: removing

src/robusta/core/discovery/utils.py
arikalon1: fixing

src/robusta/core/discovery/utils.py
arikalon1: removed the try-except

PR #467, opened 2022-08-04
Oomkill enhanced

Avi-Robusta:


Avi-Robusta:
docs  example of config
https://docs.robusta.dev/oomkill-enhanced/catalog/triggers/smart.html#pod-oomkilled

playbooks/robusta_playbooks/oom_killer.py
```diff
@@ -30,6 +30,19 @@ class OomKillerEnricherParams(ActionParams):
...
+@action
+def oomkilled_container_graph_enricher(event: PodEvent, params: ResourceGraphEnricherParams):
+    """
+    Get a graph of a specific resource for this pod. Note: "Disk" Resource is not supported.
+    """
+    pod = event.get_pod()
+    if not pod:
+        logging.error(f"cannot run pod_oom_killer_enricher on event with no pod: {event}")
+        return
+    oomkilled_container = pod_most_recent_oom_killed_container(pod)
```
Avi-Robusta, line 42: check container isnt None and return if it is

src/robusta/core/triggers/container_oom_killed_trigger.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def should_fire(self, event: TriggerEvent, playbook_id: str):
+        return super().should_fire(event, playbook_id)
+
+    def get_relevant_oomkilled_containers(self, pod: Pod):
+        containers = pod.status.containerStatuses + pod.status.initContainerStatuses
+        if not self.exclude:
+            return containers
+        for selector in self.exclude:
+            namespace = None if "namespace" not in selector else selector["namespace"]
```
arikalon1, line 34: This is just `selector.namespace`, isn't it?

src/robusta/core/triggers/container_oom_killed_trigger.py
```diff
@@ -0,0 +1,50 @@
...
+            namespace_prefix=namespace_prefix,
+            labels_selector=labels_selector,
+            rate_limit=rate_limit,
+            exclude=exclude,
+        )
+
+    def should_fire(self, event: TriggerEvent, playbook_id: str):
+        return super().should_fire(event, playbook_id)
+
+    def get_relevant_oomkilled_containers(self, pod: Pod):
```
arikalon1, line 29: Please add return type

src/robusta/core/triggers/container_oom_killed_trigger.py
```diff
@@ -0,0 +1,50 @@
...
+                continue
+            namespace_match = namespace and namespace == pod.metadata.namespace
+            if namespace and not namespace_match:
+                # this selector isnt the current namespace
+                continue
+            if namespace_match and not name:
+                # this selector is for all containers on this namespace
+                return False
+            # either namespace_match or namespace isnt defined for this name
+            containers = [container for container in containers if not container.name.startswith(name)]
```
arikalon1, line 48: I think there's a bug here, because you override the containers list for each exclude.
So only the last exclude counts

src/robusta/core/triggers/container_oom_killed_trigger.py
```diff
@@ -0,0 +1,50 @@
...
+        # pod not excluded
+        return containers
```
arikalon1, line 50: I think you can replace it all with the following, I think it's much clearer:

    def get_relevant_oomkilled_containers(self, pod: Pod) -> List[ContainerStatus]:
        statuses = pod.status.containerStatuses + pod.status.initContainerStatuses
        return [status for status in statuses 
                if not self.__is_container_excluded(status, pod.metadata.namespace)]

    def __is_container_excluded(self, container: ContainerStatus, namespace: str) -> bool:
         
        for selector in self.exclude:
            if self.__is_excluded(selector.namespace, namespace, False) and \
                    self.__is_excluded(selector.name, container.name, True):
                return True
        return False
                 
    @staticmethod
    def __is_excluded(exclude: str, value: str, prefix: bool) -> bool:
        if not exclude:
            return False
        if prefix:
            return value.startswith(exclude)
        else:
            return value == exclude

src/robusta/core/triggers/oom_killed_trigger_base.py
```diff
@@ -0,0 +1,86 @@
...
+        super().__init__(
+            name_prefix=name_prefix,
+            namespace_prefix=namespace_prefix,
+            labels_selector=labels_selector,
+        )
+        self.rate_limit = rate_limit
+        self.exclude = exclude
+
+    @abstractmethod
+    def get_relevant_oomkilled_containers(self, pod: Pod):
```
arikalon1, line 47: please add return type

src/robusta/core/triggers/oom_killed_trigger_base.py
```diff
@@ -0,0 +1,86 @@
...
+            return False
+
+        exec_event = self.build_execution_event(event, {})
+
+        if not isinstance(exec_event, PodChangeEvent):
+            return False
+
+        pod = exec_event.get_pod()
+
+        containers = self.get_relevant_oomkilled_containers(pod)
```
arikalon1, line 65: These are statuses

src/robusta/core/triggers/oom_killed_trigger_base.py
```diff
@@ -0,0 +1,86 @@
...
+
+        if len(containers) == 0:
+            return False
+
+        oom_killed = pod_most_recent_oom_killed_container(pod,containers=containers, only_current_state=True)
```
arikalon1, line 70: When do you use only_current_state?

src/robusta/core/triggers/pod_oom_killed_trigger.py
```diff
@@ -1,63 +1,51 @@
...
         super().__init__(
             name_prefix=name_prefix,
             namespace_prefix=namespace_prefix,
             labels_selector=labels_selector,
+            rate_limit=rate_limit,
+            exclude=exclude,
         )
-        self.rate_limit = rate_limit
 
     def should_fire(self, event: TriggerEvent, playbook_id: str):
```
arikalon1, line 26: This is redundant

src/robusta/core/triggers/pod_oom_killed_trigger.py
```diff
@@ -1,63 +1,51 @@
...
-        )
-        namespace = pod.metadata.namespace
-        return RateLimiter.mark_and_test(
-            f"PodOOMKilledTrigger_{playbook_id}",
-            namespace + ":" + name,
-            self.rate_limit,
-        )
+        return super().should_fire(event, playbook_id)
+
+    def get_relevant_oomkilled_containers(self, pod: Pod):
```
arikalon1, line 29: Looks like it's identical to the function in the other trigger.

It should probably appear once in the base

arikalon1 (approved):
Nice work

Left some comments

src/robusta/core/triggers/container_oom_killed_trigger.py
Avi-Robusta: its supposed to override, for each exclude we want to remove all the excluded containers
all that will be left is the container that does not start with any of the names

src/robusta/core/triggers/oom_killed_trigger_base.py
Avi-Robusta: everywhere else specifically in the actions, 
We have seen a difference in state at the time the trigger sees the states and the time an action gets to see the states
for our example if we run three actions, by the time some of the actions are run the current state can already be "running" again and the last state will be "OOMKilled"
This was seen allot in testing

src/robusta/core/triggers/pod_oom_killed_trigger.py
Avi-Robusta: its not, also logically different

src/robusta/core/triggers/container_oom_killed_trigger.py
Avi-Robusta: issue with pydantic fixed

PR #470, opened 2022-08-07
Update privacy-and-security.rst

RohitYandigeri:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=s1cIiyp)
Updated the tech docs


aantn:
Hi @RohitYandigeri,
Thank you for the suggested changes. In this case, I would rather leave the existing language. We also don't do anything at the moment that I would call AI driven.

PR #471, opened 2022-08-07
Multi stack trace support

Avi-Robusta:


Avi-Robusta:
example file output
[debugger_stack_trace_568601.txt](https://github.com/robusta-dev/robusta/files/9276556/debugger_stack_trace_568601.txt)

playbooks/robusta_playbooks/pod_troubleshooting.py
```diff
@@ -277,20 +292,47 @@ def debugger_stack_trace(event: PodEvent, params: DebuggerParams):
...
+    try:
+        output_json = json.loads(output)
+        if len(output_json) == 0 or (len(output_json) == 1 and output_json[0]["status"].lower() != "success"):
+            # no stack traces returned or only one with error
+            error_message = 'Failed to get python stack trace'
+            if len(output_json) == 1:
+                error_message += f', debugger error {output_json[0]["error"]} at {output_json[0]["trace"]}'
+            logging.error(error_message)
+            blocks.append(MarkdownBlock(f"Error while getting python stack trace."))
+        elif len(output_json) == 1 and output_json[0]["status"] == "success":
```
RoiGlinik, line 311: @Avi-Robusta I think .lower() is missing here as well on the status?
maybe its better to save it in a variable for the first 2 blocks, but not a must

playbooks/robusta_playbooks/pod_troubleshooting.py
```diff
@@ -277,20 +292,47 @@ def debugger_stack_trace(event: PodEvent, params: DebuggerParams):
...
+                if thread_output.startswith("Current thread"):
+                    # this is the thread we are getting the stack trace from, not relevant for debugging
+                    continue
+                if thread_output and output_json[0]["status"].lower() == "success":
+                    blocks.append(MarkdownBlock(f"```\n{thread_output}\n```"))
+        else:
+            # print multiple stack traces to file
+            clean_output = []
+            for trace_object in output_json:
+                if trace_object["status"] != "success":
```
RoiGlinik, line 324: also here ? not sure if missing lower?

playbooks/robusta_playbooks/pod_troubleshooting.py
```diff
@@ -277,20 +292,47 @@ def debugger_stack_trace(event: PodEvent, params: DebuggerParams):
...
+            logging.error(error_message)
+            blocks.append(MarkdownBlock(f"Error while getting python stack trace."))
+        elif len(output_json) == 1 and output_json[0]["status"] == "success":
+            # print single stack trace directly to finding
+            single_stack_trace = output_json[0]["trace"]
+            for thread_output in single_stack_trace.split("\n\n"):
+                if thread_output.startswith("Current thread"):
+                    # this is the thread we are getting the stack trace from, not relevant for debugging
+                    continue
+                if thread_output and output_json[0]["status"].lower() == "success":
```
RoiGlinik, line 318: Also might be useful to use some const on that page for the "success" string

RoiGlinik (approved):
@Avi-Robusta  left some minor comments, but otherwise approved

playbooks/robusta_playbooks/pod_troubleshooting.py
```diff
@@ -268,6 +279,10 @@ def debugger_stack_trace(event: PodEvent, params: DebuggerParams):
     if not pid:
         logging.info(f"debugger_stack_trace - no relevant pids")
 
+    if params.traces_amount < 1 or params.sleep_duration_s < 0:
```
arikalon1, line 282: I think min duration should be 1
Regardless, if this is an error, then shouldn't we return here?

playbooks/robusta_playbooks/pod_troubleshooting.py
```diff
@@ -277,20 +292,47 @@ def debugger_stack_trace(event: PodEvent, params: DebuggerParams):
...
+                    # the full python stack trace of the error will appear here
+                    logging.error(f'Failed to get stack trace, debugger error {trace_object["error"]} at {trace_object["trace"]}')
+                    clean_output.append({"time": trace_object["time"], "status": "Error: Failed to get stack trace."})
+                else:
+                    clean_output.append(trace_object)
+            clean_file_output = json.dumps(clean_output, indent=4, sort_keys=True).replace('\\n', '\n')
+            blocks.append(FileBlock(f"debugger_stack_trace_{pid}.txt", clean_file_output.encode()))
+    except ValueError:  # includes simplejson.decoder.JSONDecodeError, could still be valid output
+        logging.error(f"failed to decode output")
+        blocks.append(MarkdownBlock(f"```\n{output}\n```"))
```
arikalon1, line 334: What is output here?
I don't think we should send, unexpected random data to the user

playbooks/robusta_playbooks/pod_troubleshooting.py
```diff
@@ -268,6 +279,10 @@ def debugger_stack_trace(event: PodEvent, params: DebuggerParams):
     if not pid:
```
arikalon1, line 279: What is the behavior if pid is none?
`debug-toolkit stack-trace {pid}` works?

playbooks/robusta_playbooks/pod_troubleshooting.py
```diff
@@ -277,20 +292,47 @@ def debugger_stack_trace(event: PodEvent, params: DebuggerParams):
...
-    for thread_output in output.split("\n\n"):
-        if thread_output.startswith("Current thread"):
-            # this is the thread we are getting the stack trace from, not relevant for debugging
-            continue
-        if thread_output:
-            blocks.append(MarkdownBlock(f"```\n{thread_output}\n```"))
-    finding.add_enrichment(blocks, annotations={SlackAnnotations.ATTACHMENT: True})
+    try:
+        output_json = json.loads(output)
+        if len(output_json) == 0 or (len(output_json) == 1 and output_json[0]["status"].lower() != "success"):
```
arikalon1, line 304: What is the data on output_json[0] ?
it has status, error, and trace fields?
A little hard to understand

playbooks/robusta_playbooks/pod_troubleshooting.py
arikalon1, line 318: I think we already checked if status is success on line 311

playbooks/robusta_playbooks/pod_troubleshooting.py
Avi-Robusta: default sleep duration is 1 but its possible to do repetitive stack traces with no sleep

playbooks/robusta_playbooks/pod_troubleshooting.py
Avi-Robusta: each stack trace object has

- time
- status
- error
- trace
fields, I can throw it in an enum
we only show the time and trace outputs in findings and error is printed to log if there is an error

playbooks/robusta_playbooks/pod_troubleshooting.py
```diff
@@ -1,9 +1,30 @@
 # TODO: move the python playbooks into their own subpackage and put each playbook in its own file
+import logging
+
 import humanize
 from robusta.api import *
 from typing import List
 
 
+class StackTraceObject:
```
arikalon1, line 9: Can we use Pydantic here? 

class StackTraceObject(BaseModel):
    time: float = None
    status: str = None
    error: str = None
    trace: str = None

Any reason not to?

PR #479, opened 2022-08-11
Feature/mattermost sink

anFatum:


src/robusta/integrations/mattermost/sender.py
```diff
@@ -0,0 +1,183 @@
...
+            "color": msg_color
+        }]
+        attachments.extend([{"image_url": attachment, "color": msg_color} for attachment in attachment_blocks])
+        msg = {
+            "username": "Robusta",
+            "icon_url": ROBUSTA_LOGO_URL,
+            "attachments": attachments
+        }
+        if self.channel:
+            msg['channel'] = self.channel
```
arikalon1, line 61: please use double quotes

src/robusta/integrations/mattermost/sender.py
```diff
@@ -0,0 +1,183 @@
...
+
+        return msg
+
+    def __to_mattermost(self, block: BaseBlock, sink_name: str) -> Optional[str]:
+        if isinstance(block, MarkdownBlock):
+            return Transformer.to_github_markdown(block.text)
```
arikalon1, line 67: Does Mattermost supports Slack markdown or github markdown, or both?

src/robusta/integrations/mattermost/sender.py
```diff
@@ -0,0 +1,183 @@
...
+        elif isinstance(block, FileBlock):
+            extension = re.findall(extension_regex, block.filename)
+            if not extension:
```
arikalon1, line 70: When will this happen?
Is this if someone created a file without an extension?
I would add an error log here, so it will be easier to understand what the issue is.

src/robusta/integrations/mattermost/sender.py
```diff
@@ -0,0 +1,183 @@
...
+                return ""
+            extension = extension[0][1:]
+            if extension not in SUPPORTED_IMAGES_FORMATS:
+                return ""
```
arikalon1, line 74: Does Mattermost supports svg?

docs/catalog/sinks/mattermost.rst
```diff
@@ -0,0 +1,76 @@
...
+4. Select the channel to receive webhook payloads, then click Add to create the webhook.
+
+.. note::
+
+    **Enable integrations to override usernames** and **Enable integrations to override profile picture icons**
+    must be set to true. Enable them from System Console > Integrations > Integration Management,
+    or ask your System Admin to do so. If not enabled, the icon of the creator of the webhook URL is used to post messages.
+
+
+Enabling override usernames and profile picture
```
arikalon1, line 40: I would write it before the section about creating the webhook
I would add the note about why we need to override it to it
so it would be:
1. allow user name override
2. create webhook
3. use webhook address to create sink config

src/robusta/integrations/mattermost/sender.py
```diff
@@ -0,0 +1,183 @@
...
+
+    def __to_mattermost(self, block: BaseBlock, sink_name: str) -> Optional[str]:
+        if isinstance(block, MarkdownBlock):
+            return Transformer.to_github_markdown(block.text)
+        elif isinstance(block, FileBlock):
+            extension = re.findall(extension_regex, block.filename)
+            if not extension:
+                return ""
+            extension = extension[0][1:]
+            if extension not in SUPPORTED_IMAGES_FORMATS:
```
arikalon1, line 73: I think we ignore all text files here, like .txt, .log, .yaml
Maybe we need to check if the suffix is an image, and if so to send it as an image.
If it's not, send it as a text file attachment.

WDYT?

arikalon1 (approved):
Great work!
I left some minot comments.
I think we're not handling well text files

src/robusta/integrations/mattermost/sender.py
anFatum: It does according to the docs, but I tried sending svgs several times and they weren't rendered properly, so I decided to leave them out.

docs/catalog/sinks/mattermost.rst
anFatum: Okay, I will reorder it!

src/robusta/integrations/mattermost/sender.py
anFatum: The thing is, Mattermost does not support sending files attachments.
The only way we can do it at this moment - hosting a Robusta bot, similar we do with slack. Should we switch to that solution?

src/robusta/integrations/mattermost/sender.py
anFatum: It does support both, but the links should be edited with GitHub style, so I add it there

src/robusta/integrations/mattermost/sender.py
arikalon1: got it, ok

src/robusta/integrations/mattermost/sender.py
arikalon1: The crash logs is the feature users likes the most.
So I think it's important to add it.
What do you mean by hosting a Robusta bot?

src/robusta/integrations/mattermost/sender.py
arikalon1: ok, great

src/robusta/integrations/mattermost/client.py
```diff
@@ -0,0 +1,109 @@
...
+        self.token_id = token_id
+        self.channel_id = channel_id
+        schema = schema or "https"
+        if schema not in _SUPPORTED_SCHEMAS:
+            raise AttributeError(f"{schema} is not supported schema, should be one of following: {_SUPPORTED_SCHEMAS}")
+        self.schema = schema
+        self._init_setup()
+
+    def _send_mattermost_request(self, url: str, method: HttpMethod, **kwargs):
+        headers = kwargs.pop("headers", {})
```
arikalon1, line 30: Sometime we call it without kwargs
Does that work ok? What will be in kwargs when we call only with url & method?

src/robusta/integrations/mattermost/client.py
```diff
@@ -0,0 +1,109 @@
...
+            "",
+            self.client_url,
+            _API_PREFIX,
+            endpoint,
+            *args
+        ])
+
+    def get_token_owner_id(self):
+        endpoint = "users/tokens"
+        url = self._get_full_mattermost_url(endpoint, self.token_id)
```
arikalon1, line 46: I think it might be easier to have patterns, and make it less generic:
For example:
USER_TOKENS_URL = f"{host}/{API_VERSION}/users/tokens/{token_id}"

src/robusta/integrations/mattermost/client.py
```diff
@@ -0,0 +1,109 @@
...
+        response = self._send_mattermost_request(url, HttpMethod.GET)
+        if response.status_code != HTTPStatus.OK:
```
arikalon1, line 48: If there's an error, I think response is None

src/robusta/integrations/mattermost/client.py
```diff
@@ -0,0 +1,109 @@
...
+        return "/".join([
+            f"{self.schema}:",
+            "",
+            self.client_url,
+            _API_PREFIX,
+            endpoint,
+            *args
+        ])
+
+    def get_token_owner_id(self):
```
arikalon1, line 44: Can you please add a return type?

src/robusta/integrations/mattermost/client.py
```diff
@@ -0,0 +1,109 @@
...
+        url = self._get_full_mattermost_url(endpoint, self.token_id)
+        response = self._send_mattermost_request(url, HttpMethod.GET)
+        if response.status_code != HTTPStatus.OK:
+            logging.warning("Cannot get owner token, probably bot has not enough permissions")
+            return
+        response_data = response.json()
+        return response_data.get("user_id")
+
+    def update_bot_settings(self, bot_id):
+        endpoint = "bots"
```
arikalon1, line 55: same here, I would just use a constant pattern

src/robusta/integrations/mattermost/client.py
```diff
@@ -0,0 +1,109 @@
...
+        url = self._get_full_mattermost_url(endpoint, bot_id)
+        response = self._send_mattermost_request(url, HttpMethod.PUT, json={
+            "username": "robusta",
+            "display_name": "Robusta"
+        })
+        if response.status_code != HTTPStatus.OK:
```
arikalon1, line 61: response might be None

src/robusta/integrations/mattermost/client.py
```diff
@@ -0,0 +1,109 @@
...
+            logging.warning("Cannot update bot settings, probably bot has not enough permissions")
+        self.update_bot_logo(bot_id)
+
+    def update_bot_logo(self, bot_id):
+        endpoint = f"users/{bot_id}/image"
+        img_data = process_request(ROBUSTA_LOGO_URL, HttpMethod.GET).content
```
arikalon1, line 67: This can also be None I think

src/robusta/integrations/mattermost/client.py
```diff
@@ -0,0 +1,109 @@
...
+        url = self._get_full_mattermost_url(endpoint)
+        response = self._send_mattermost_request(url, HttpMethod.POST, files=[("image", ("image", img_data))])
```
arikalon1, line 69: same

docs/catalog/sinks/mattermost.rst
```diff
@@ -0,0 +1,92 @@
...
+
+    .. code-block:: yaml
+
+        sinks_config:
+        - mattermost_sink:
+            name: personal_mattermost_sink
+            url: <YOUR MATTERMOST URL> (can be find out from url bar in browser, e.g. namespace.cloud.mattermost.com)
+            token: <YOUR BOT TOKEN> (the token we copied the first after bot creation)
+            token_id: <YOUR BOT TOKEN ID> (the token id visible in bot panel)
+            channel: <YOUR CHANNEL ID> (the channel id you want to send messages to)
```
arikalon1, line 85: Is there a way to use the channel name and not the channel id?
It will be much easier for users.
With Slack I think you can use: #my-channel-name

arikalon1 (approved):
Great work!

I left few comments.
It's mostly same comment, in few different places.
Can you also add a screenshot of how the message looks to the docs?

src/robusta/integrations/mattermost/client.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+class MattermostClient:
+    def __init__(
+            self, url: str, token: str, token_id: str, channel_id: str, schema: Optional[str]
+    ):
+        """
+        Set the Mattermost webhook url.
+        """
+        self.client_url = url.split(":")[-1].lstrip("/").rstrip("/")
```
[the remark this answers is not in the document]
anFatum: hmm I didn't think of that, I was trying to cut https:// or http:// starting if it is in the url, will change it so it aware of port as well

docs/catalog/sinks/mattermost.rst
anFatum: done!

docs/catalog/sinks/mattermost.rst
```diff
@@ -0,0 +1,92 @@
...
+Now we're ready to configure the Mattermost sink.
+
+.. admonition:: values.yaml
+
+    .. code-block:: yaml
+
+        sinks_config:
+        - mattermost_sink:
+            name: personal_mattermost_sink
+            url: <YOUR MATTERMOST URL> (can be find out from url bar in browser, e.g. namespace.cloud.mattermost.com)
```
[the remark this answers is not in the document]
anFatum: changed so user can provide full url if desired

src/robusta/integrations/common/requests.py
```diff
@@ -0,0 +1,57 @@
...
+    try:
+        http_method = map_type_to_method(method)
+        logging.debug(f"Try to proceed {method.name} request to {url}")
+        response = http_method(url, **kwargs)
+        response.raise_for_status()
+    except requests.RequestException as e:
+        logging.error(f"Error while {method.name} request to {url}")
+        logging.error(f"Response code: {response.status_code}")
+        logging.error(f"Request body: {response.request.body}")
+        logging.error(json.dumps(response.json()))
```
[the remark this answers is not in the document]
anFatum: I'm returning the response from error now, so it will never be None now

anFatum:
> Great work!
> 
> I left few comments. It's mostly same comment, in few different places. Can you also add a screenshot of how the message looks to the docs?

Yeah, I've updated the docs with new screenshot, so it includes the file attachments now

src/robusta/integrations/mattermost/client.py
anFatum: kwargs are used to pass extra arguments to the request (form data, params, body request, files, headers etc)
If we want a plain request (e.g. GET) we do not need to pass any extra arguments

docs/catalog/sinks/mattermost.rst
```diff
@@ -0,0 +1,72 @@
...
+            token: <YOUR BOT TOKEN> (the token we copied the first after bot creation)
+            token_id: <YOUR BOT TOKEN ID> (the token id visible in bot panel)
+            channel: <YOUR CHANNEL NAME> (the channel name you want to send messages to - either display name or channel name divided by hyphen (e.g. channel-name))
+            http_schema: <http/https> [OPTIONAL, HTTPS default] (if you want to send requests over HTTP or HTTPS, if url passed with schema, it overrides the one provided in params)
```
arikalon1, line 66: Can we just remove the schema parameter?
We always use it with the URL.
Why do we need to provide the option to override the url?
I think it would be the best to just have a url parameter, that includes the schema in it

arikalon1 (approved):

PR #484, opened 2022-08-17
fixed abnormal memory unit support

Avi-Robusta:
normal digits and scientific notation are automatically converted to units
anything less than 1m is automatically converted to 1m, for example 120u is 1m upon container creation
for ref on memory usage units you have this:
https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-memory

### DIFF
--- a/src/robusta/core/model/pods.py
+++ b/src/robusta/core/model/pods.py
@@ -1,3 +1,4 @@
+import logging
 from enum import Enum
 from typing import Optional, List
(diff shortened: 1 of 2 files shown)


arikalon1 (approved):
nice work, left some minor comments

arikalon1:
I think the most important part here, is to surround the parsing with try/except, so that if it fails, we won't kill the whole discovery, like it is now

PR #485, opened 2022-08-17
Docker cli

RoiGlinik:


docs/_static/robusta
```diff
@@ -0,0 +1,35 @@
+#!/bin/sh
+
+function cleanup() {
```
RoiGlinik, line 3: this code hides the cursor so the spinner will be less "jumpy"

docker-cli/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+RUN apt-get update \
+    && dpkg --add-architecture arm64 \
+    && apt-get install -y --no-install-recommends ssh curl python3-dev libffi-dev gcc \
+    && pip install ruamel.yaml.clib \
+    && curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl" \
+    && apt-get purge -y --auto-remove \
+    && rm -rf /var/lib/apt/lists/*
+
+RUN chmod +x kubectl
+RUN mv ./kubectl /usr/local/bin
```
arikalon1, line 12: Will this work on any os?

docker-cli/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM python:3.9-slim
+
+RUN apt-get update \
+    && dpkg --add-architecture arm64 \
+    && apt-get install -y --no-install-recommends ssh curl python3-dev libffi-dev gcc \
```
arikalon1, line 5: what's python3-dev?

docs/installation.rst
```diff
@@ -18,21 +18,52 @@ For this we need to install Robusta, and also connect at least one destination (
...
+                :class: warning
+
+                * Python 3.7 or higher is required
+                * If you are using a system such as macOS that includes both Python 2 and Python 3, run pip3 instead of pip.
+                * Errors about *tiller* mean you are running Helm 2, not Helm 3
+
+        .. tab-item:: Docker
+            :name: docker-cli-tab
+
+            Robusta CLI and Kubectl are included in the image. Any file generated in the image will be copied to your current directory.
```
arikalon1, line 45: I think this is too much information
I would just say it required docker to be present on the machine

docs/user-guide/robusta-cli.rst
```diff
@@ -4,13 +4,55 @@ CLI Commands
...
+            :class: warning
+
+            * Python 3.7 or higher is required
+            * If you are using a system such as macOS that includes both Python 2 and Python 3, run pip3 instead of pip.
+            * Errors about *tiller* mean you are running Helm 2, not Helm 3
+
+    .. tab-item:: Docker
+        :name: docker-cli-tab
+
+        Robusta CLI and Kubectl are included in the image. Any file generated in the image will be copied to your current directory.
```
arikalon1, line 27: here also, I think this is too informative

docs/user-guide/robusta-cli.rst
```diff
@@ -4,13 +4,55 @@ CLI Commands
...
+            chmod +x robusta
+
+        * Use the script, for example:
+
+        .. code-block:: bash
+            :name: docker-cli-example
+
+            ./robusta version
+
+        * Inorder to get the latest version of the docker image run:
```
arikalon1, line 45: should be: "In order"
And I think it should be:
"In order to upgrade the cli run:"

The docker is an implementation detail

skaffold.yaml
```diff
@@ -28,6 +28,7 @@ deploy:
 portForward:
 - resourceType: deployment
   resourceName: robusta-runner
+  namespace: default
```
arikalon1, line 31: why did you add that?

docs/installation.rst
```diff
@@ -18,21 +18,52 @@ For this we need to install Robusta, and also connect at least one destination (
...
+            For **Windows** please use `WSL <https://docs.microsoft.com/en-us/windows/wsl/install>`_.
```
arikalon1, line 46: What is WSL? Why do we need it?

arikalon1 (approved):
nice work!
Added few minor comments.

Do we know which OSs it will work on?
Should we test?

Feel free to fix and merge

skaffold.yaml
RoiGlinik: its left over from a fix. removed it

docker-cli/Dockerfile
RoiGlinik: it was needed to install the rumal.yaml.clib because there was no wheel for this arch. I just checked and they added a new one in august so it's redundant. We might be able to remove a some from the runner dockerfile as well

docs/installation.rst
RoiGlinik: the .sh script will not work on cmd or powershell. However, Docker installation on windows requests to install WSL which will provide a bash shell. There you can run the script.

RoiGlinik:
tested on a few of our macs machines and one windows machine

docker-cli/Dockerfile
RoiGlinik: I belive it will

PR #487, opened 2022-08-21
Performance v2

arikalon1:


src/robusta/patch/patch.py
```diff
@@ -0,0 +1,122 @@
+from typing import Union, List, Dict, get_type_hints
+from dataclasses import is_dataclass, InitVar
+from inspect import signature, getmodule
+from hikaru import HikaruDocumentBase, HikaruBase
+from ruamel.yaml import YAML
+
+try:
+    from typing import get_args, get_origin
+except ImportError:  # pragma: no cover
+    def get_args(tp):
```
aantn, line 10: This is weird. I'm sure there is a good reason for it though! Perhaps add a comment explaining why we're doing this?

src/robusta/patch/patch.py
```diff
@@ -0,0 +1,122 @@
...
+        return tp.__args__ if hasattr(tp, "__args__") else ()
+
+    def get_origin(tp):
+        return tp.__origin__ if hasattr(tp, "__origin__") else None
+
+NoneType = type(None)
+
+
+def create_monkey_patches():
+    print("Creating hikaru monkey patches")
```
aantn, line 20: Use `logging` not `print`

src/robusta/patch/patch.py
```diff
@@ -0,0 +1,122 @@
...
+    print("Creating yaml monkey patch")
+    YAML.official_plug_ins = official_plug_ins
+
+
+def official_plug_ins(self):
+    return []
+
+# hikaru meta.py monkey patch function
+@classmethod
+def get_empty_instance(cls):
```
aantn, line 32: I would explicitly add a comment marking the areas that are your changes vs the orignal

src/robusta/patch/patch.py
```diff
@@ -0,0 +1,122 @@
...
+    def get_args(tp):
+        return tp.__args__ if hasattr(tp, "__args__") else ()
+
+    def get_origin(tp):
+        return tp.__origin__ if hasattr(tp, "__origin__") else None
+
+NoneType = type(None)
+
+
+def create_monkey_patches():
```
aantn, line 19: I would add a comment here explaining the motivation for each of these patches.

aantn (commented):
Didn't look at the actual patches in depth, but looks good.
Left some minor comments mostly regarding comments.

src/robusta/patch/patch.py
arikalon1: I'm not sure tbh. I just copied that section from Hikaru meta.py
Do you think we should change something here?

src/robusta/patch/patch.py
arikalon1: Correct, fixing

src/robusta/patch/patch.py
arikalon1: adding it

src/robusta/patch/patch.py
arikalon1: Adding

src/robusta/patch/patch.py
aantn: No it's fine. Probably just to support older python versions

PR #493, opened 2022-08-24
Multi processing

arikalon1:


src/robusta/core/sinks/robusta/dal/supabase_dal.py
```diff
@@ -129,7 +129,7 @@ def to_service(self, service: ServiceInfo) -> Dict[Any, Any]:
             "account_id": self.account_id,
             "deleted": service.deleted,
             "service_key": service.get_service_key(),
-            "config": service.service_config.dict(),
+            "config": service.service_config.dict() if service.service_config else None,
```
aantn, line 132: Does this fix a bug unrelated to the rest of the changes?

src/robusta/integrations/prometheus/trigger.py
```diff
@@ -87,22 +106,23 @@ def __find_node_by_ip(cls, ip) -> Optional[Node]:
                 return node
         return None
 
-    @classmethod
-    def __load_node(cls, alert: PrometheusAlert, node_name: str) -> Optional[Node]:
+    @staticmethod
```
aantn, line 109: Any reason for changing this from a classmethod to staticmethod? It doesn't actually matter much, but you're calling another method on the same class (`__find_node_by_ip`). By keeping it as a class method and using `cls` the code will work even if someone changes the name of the class in the future.

aantn: Same question applies to other methods below.

aantn (approved):
Looks good. Left some comments.

src/robusta/core/sinks/robusta/dal/supabase_dal.py
arikalon1: yes

src/robusta/integrations/prometheus/trigger.py
arikalon1: I'll change it back

Issue #475, opened 2022-08-07
Add action to do http check


aantn:
@mitin20 requested this on Slack. (See [here](https://robustacommunity.slack.com/archives/C02R0LSAHNJ/p1659446720825649))

Is this do-able with the existing job-runner and curl? Should we have a native action for this?

Ideally we would have an option to only create the finding if the http check failed.

alikhanxgrid:
What kind of event should this action take? I have read the conversation and only saw talk about deployments. 
I suppose it would only take a DeploymentEvent as a parameter ? @aantn

aantn:
@alikhanxgrid yes, that sounds like the right solution!

Issue #222, opened 2022-02-18
Add action to show files in Volume


aantn:
**Motivation**
It can be hard to understand exactly what data is contained inside a Kubernetes volume. We can provide visibility with Robusta actions

**Suggested Implementation**
Add a Robusta action which:

1. Takes a Kubernetes volume as input
2. Runs a "reader" job/pod which mounts that volume  
3. The reader pod should run `ls -R` on the volume to show all files.
4. The Robusta action should gather the reader pod's output and send it to Slack/other sinks by creating a finding

**Bonus**
Add support for VolumeSnapshots too. For VolumeSnapshots you will first need to create a temporary Volume based on the snapshot, then do the above, and finally delete the temporary volume.

**Caveats**
The above sometims wont work if the Volume is in use - depending on the mount's [AccessMode](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes). This can be fixed various ways. For example, for ReadWriteOnce it can be bypassed by running the reader pod on the same node as the pod that is currrently using the volume. Alternatively, it could be fixed for *all* AccessModes by always creating a VolumeSnapshot and reading the snapshot not the original.

In any event, a first version doesn't need to support any of this.

AvineshTripathi:
I am looking forward this issue, doing a dive to get my hands dirty is there anything I should be aware about the issue or descriptions says all

aantn:
Everything is in the description. Feel free to ask any questions in the #beginners channel on Slack

AvineshTripathi:
Hey @aantn when you say volume is it a object that the function would receive or just the name of the volume

aantn:
@AvineshTripathi hope your exams went well!

The playbook action should take a PersistentVolumeEvent as input:
```
@action
def volume_analysis(event: PersistentVolumeEvent):
    volume = event.get_persistentvolume()
    print(f"got volume named {volume.metadata.name} as input with current status: {volume.status.reason}")
    # create a reader pod (using the builtin hikaru library) that does stuff with the volume
    reader_pod = Pod(
        ...
    )
```

To trigger it you can just provide the volume's name though. Something like the cli command:

```
robusta playbooks trigger volume_analysis name=some-volume namespace=some-ns
```

aantn:
@AvineshTripathi we have a new tutorial on writing playbooks actions that @Sheeproid wrote, so make sure you read that first:

https://docs.robusta.dev/master/tutorials/my-first-custom-action.html

[...]

PR #494, opened 2022-08-25
Issues #475 and #222

alikhanxgrid:
- Resolved issue #475 
- Resolved issue #222 (It's the first version. Will add support for volume snapshots)

Please review and check if anything needs to be changed.


playbooks/robusta_playbooks/volume_analysis.py
```diff
@@ -0,0 +1,76 @@
...
+    # Get persistent volume data the object contains data related to PV like metadata etc
+    volume = event.get_persistentvolume()
+
+    # Define a reader (busybox) pod which mounts the volume
+
+    reader_pod = RobustaPod(
+        apiVersion="v1",
+        kind="Pod",
+        metadata=ObjectMeta(
+            name="volume-inspector",
```
arikalon1, line 20: I think you can also add the namespace here: namespace=volume.spec.claimRef.namespace,

playbooks/robusta_playbooks/volume_analysis.py
```diff
@@ -0,0 +1,76 @@
...
+    """
+    This action shows you the files present on your persistent volume
+    """
+
+    # Get persistent volume data the object contains data related to PV like metadata etc
+    volume = event.get_persistentvolume()
+
+    # Define a reader (busybox) pod which mounts the volume
+
+    reader_pod = RobustaPod(
```
arikalon1, line 16: I would call this: reader_pod_spec

playbooks/robusta_playbooks/volume_analysis.py
```diff
@@ -0,0 +1,76 @@
...
+                            mountPath="/pvc",
+                            name="pvc-mount",
+                        )
+                    ],
+                )
+            ]
+        )
+    )
+    # create the pod defined
+    reader_pod_create = reader_pod.createNamespacedPod(
```
arikalon1, line 48: I would name this: reader_pod

playbooks/robusta_playbooks/volume_analysis.py
```diff
@@ -0,0 +1,76 @@
...
+        finding_type=FindingType.REPORT,
+        failure=False,
+    )
+        # exec a cmd on the reader pod to get file contents of PV
+        cmd = f"ls -R /pvc"
+        block_list: List[BaseBlock] = []
+        output = reader_pod_create.exec(cmd)
+        block_list.append(MarkdownBlock(
+            f"Files currently present on your volume {volume.metadata.name} are:*"))
+        block_list.append(MarkdownBlock(output))
```
arikalon1, line 68: I think I would add the results in a FileBlock
Markdown blocks as a size limit on some sinks, and the output here might be big

playbooks/robusta_playbooks/volume_analysis.py
```diff
@@ -0,0 +1,76 @@
...
+        finding.add_enrichment(block_list)
+        event.add_finding(finding)
+        print(f"output is {output}")
```
arikalon1, line 71: I would remove this

playbooks/robusta_playbooks/volume_analysis.py
```diff
@@ -0,0 +1,76 @@
+from robusta.api import action, PersistentVolumeEvent, RobustaPod, ObjectMeta, PodSpec, Volume, PersistentVolumeClaimVolumeSource, Container, VolumeMount, List, BaseBlock, MarkdownBlock,Finding,FindingSource,FindingType
+
+
+@action
```
arikalon1, line 4: This is a very useful action.

I tried to test it on our prometheus PVC, and had an error of "Multi-Attach error for volume", because the Prometheus PVC has AccessMode RWO (Read Write One)
From what I read, it means it can only be attached to one pod.

A solution for this might be, to run the `ls` command, on the pod that is already using the PVC, using `kubectl exec`

@aantn @alikhanxgrid what do you guys think?

arikalon1 (commented):
Great work @alikhanxgrid !
Thanks for your contribution 

Left few minor comments

playbooks/robusta_playbooks/volume_analysis.py
aantn: Sounds good to me. I had some other suggestions in the [original issue])https://github.com/robusta-dev/robusta/issues/222) (see "caveats" there) but this is a simpler solution.

aantn:
@alikhanxgrid really excited about this contribution! Thank you so much

playbooks/robusta_playbooks/http_check.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+class ServiceCheck(ActionParams):
+    """
+    :var url: In cluster target url.
+    """
+
+    url: str
+
+
```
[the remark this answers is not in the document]
alikhanxgrid: Yes seems like a good idea.

alikhanxgrid:
Closing PR. 
Will implement the suggestions provided and create new PR.

PR #505, opened 2022-08-30
add local runner guide + small compile error?

RoiGlinik:


RoiGlinik:
@aantn  Its the first draft.

Theres maybe a better way for the config step if you know lmk (also @arik)

anyway we can go with that and I can improve it later

docs/developer-guide/platform/local-runner.rst
```diff
@@ -0,0 +1,101 @@
...
+---------------
+
+To follow the guide
+
+* Run a local cluster using `Kind <https://kind.sigs.k8s.io/docs/user/quick-start/>`_  or `minikube <https://minikube.sigs.k8s.io/docs/start/>`_.
+
+Setting Up Robusta playbooks
+------------------------------
+
+Inside the Robutsa repo folder create the following file hierarchy:
```
aantn, line 14: Typo in `Robutsa`

docs/_ext/autorobusta.py
```diff
@@ -253,7 +253,7 @@ def __generate_rst(
         trigger_params = json.loads(self.options.get("trigger-params", "{}"))
 
         example_yaml = generator.generate_example_config(
-            action_definition.func, recommended_trigger, trigger_params
```
aantn, line 256: I don't think you should be removing this. It was added in [this PR](https://github.com/robusta-dev/robusta/pull/370/files) and removing it breaks one of the features added there.

Perhaps you need to run `poetry install` to get it working? Maybe for some reason the docs are importing an older version of the robusta library which doesn't have this?

docs/developer-guide/platform/local-runner.rst
```diff
@@ -0,0 +1,101 @@
...
+
+    robusta
+    ├── deployment
+        ├── defaults
+        │   ├── pyproject.toml
+        │   └── robusta_playbooks
+
+
+#. Under ``deployments`` create  ``defaults`` folder.
+#. Copy the ``robusta_plabyooks`` folder and the ``pyproject.toml`` file from the ``playbooks`` folder in the repo.
```
aantn, line 27: Typo in plabyooks

aantn (commented):
Left a few small comments. Two other comments overall:

1. I would add a link from the main dev-setup page 
2. It isn't clear *when* you would want to do this and how it differs from "Running Robusta locally" on the `dev-setup` page.

We need those two sections to make sense together in a coherent manner.

Also explain the motivation, that you can develop playbooks this way... not just what you should setup but *why*

docs/_ext/autorobusta.py
RoiGlinik: could be ill check this

PR #508, opened 2022-08-31
Feature/arm arch

anFatum:


aantn:
Hey, I happened to see it in my inbox and was curious, so I left two tiny comments. I did *not* review it.
Sorry about the noise :)

.github/workflows/release.yaml
```diff
@@ -43,6 +43,18 @@ jobs:
         sed -i 's/0.0.0/${{env.RELEASE_VER}}/g' src/robusta/_version.py pyproject.toml helm/robusta/Chart.yaml helm/robusta/values.yaml
         sed -i 's/0.0.1/${{env.RELEASE_VER}}/g' helm/robusta/Chart.yaml
 
+    # Set up the buildx to run build for multiple platforms
```
arikalon1, line 46: Do we really need these 2 github actions?
Because I saw that in kubewatch, it works without it.
Why do we need it here?

.github/workflows/test_robusta.yaml
```diff
@@ -19,6 +19,18 @@ jobs:
           run: |
             kubectl config get-contexts
 
+        # Set up the buildx to run build for multiple platforms
```
arikalon1, line 22: same question here

arikalon1 (approved):
very nice!

.github/workflows/release.yaml
arikalon1: never mind, I understand why we need it here...

PR #511, opened 2022-09-04
Event Hierarchy Diagram Updated

wahajXgrid:
I embedded a SVG diagram as a txt file.
Added a txt file and [draw.io](http://draw.io/) file in docs/images folder.

Please review the changes and let me know if you want anything changed thanks.


aantn:
Hey, thank you very much. I *think* you exported the image as a draw.io embed instead of a normal SVG.

Instead of selecting `Embed -> SVG` in draw.io, can you use `Export as -> SVG` and then select the download button. (See images below.)

Screen Shot 2022-09-05 at 18 25 23

Screen Shot 2022-09-05 at 18 25 35

This way the SVG will have a .svg file attachment and wont contain embedded Javascript.

Sorry about the extra step!

wahajXgrid:
Creating new PR

PR #513, opened 2022-09-04
Feature/Sink PagerDuty

taivo123:


src/robusta/core/sinks/pagerduty/pagerduty_sink.py
```diff
@@ -0,0 +1,119 @@
...
+            },
+            "routing_key": self.api_key,            
+            "event_action": PagerdutySink.__to_pagerduty_status_type(finding.title),
+            "dedup_key": finding.fingerprint
+        }
+        
+        headers = {"Content-Type": "application/json"}
+        response = requests.post(self.url, json=body, headers=headers)        
+
+    def __to_unformatted_text(cls, block: BaseBlock) -> str:
```
aantn, line 100: This should also be a @staticmethod and then you can remove the first parameter `cls`.

docs/catalog/sinks/PagerDuty.rst
```diff
@@ -0,0 +1,32 @@
...
+
+.. admonition:: values.yaml
+
+    .. code-block:: yaml
+
+        sinks_config:
+            - pagerduty_sink:
+                name: main_pagerduty_sink
+                api_key: <api key> # e.g. f653634653463678fadas43534506
+                url: <url> # e.g.  https://events.pagerduty.com/v2/enqueue/
```
aantn, line 19: Can the URL here ever change? From the PagerDuty [docs](https://developer.pagerduty.com/api-reference/368ae3d938c9e-send-an-event-to-pager-duty) I would assume its constant in which case you can define it as a constant in the source code. No need to expose it here as a parameter.

src/robusta/core/sinks/pagerduty/pagerduty_sink.py
```diff
@@ -0,0 +1,119 @@
...
+
+        if platform_enabled:
+            custom_details["🔎 Investigate"] = finding.investigate_uri
+
+            if finding.add_silence_url:
+                custom_details[
+                    "🔕 Silence"
+                ] = finding.get_prometheus_silence_url(self.cluster_name)
+
+        # custom fields
```
aantn, line 59: This looks good, I just wanted to verify that I understood it correctly. The fields in `custom_details` are all fields that you defined, which don't have an inherent meaning in PagerDuty itself, right?

If so, maybe add a comment explaining that. Looks good though.

src/robusta/core/sinks/pagerduty/pagerduty_sink.py
```diff
@@ -0,0 +1,119 @@
...
+
+                message_lines += text + "\n\n"
+        
+        custom_details["state_message"] = message_lines
+
+        body = {
+            "payload": {
+                "summary": finding.title,        
+                "severity": PagerdutySink.__to_pagerduty_severity_type(finding.severity),
+                "source": str(finding.source.name),
```
aantn, line 88: Would it make more sense to set `source` to something more specific? Right now, the values for `finding.source.name` will almost always be "kubernetes_api_server" or "prometheus" which are somewhat broad.

For example, we could map the following way:

* `source` is set to the cluster name
* `component` is set to the exact resource within the cluster that is associated with the alert - i.e.`finding.subject` [1]
* `group` is set to the resource's high-level component - i.e. `finding.service_key`

To give a concrete example, when an alert occurs on a pod named `some-deployment-abcd-efgh` then:
`component="some_namespace/pod/some-deployment-abcd-efgh"`
`group="some_namespace/deployment/some-deployment"`

The motivation here is to expose enough data to PagerDuty that you'll have the full flexibility to group however you want on the PagerDuty end. That said, I don't have much experience with PagerDuty, so let me know if this even makes sense or if I'm missing something obvious!

[1] To easily implement this, we should really add on a `FindingSubject.__str__` method to easily convert the finding subject to a string in a consistent manner - e.g. `{namespace}/{resource_type}/{name}`

src/robusta/core/sinks/pagerduty/pagerduty_sink.py
```diff
@@ -0,0 +1,119 @@
...
+                message_lines += text + "\n\n"
+        
+        custom_details["state_message"] = message_lines
+
+        body = {
+            "payload": {
+                "summary": finding.title,        
+                "severity": PagerdutySink.__to_pagerduty_severity_type(finding.severity),
+                "source": str(finding.source.name),                                
+                "class": str(finding.finding_type.name),
```
aantn, line 89: I think this should be `finding.aggregation_key`. I know the name is misleading, but the `aggregation_key` better represents the exact class of issue (e.g. the Prometheus alert's name) whereas `finding_type` is a broader category like "change" or "issue".

aantn (changes requested):
Very nice work. I've left some comments, but I just wanted to say that the code was a pleasure to read. Extremely readable and clear with just the right amount of comments.

docs/catalog/sinks/PagerDuty.rst
taivo123: I was not sure because my PagerDuty account recommended me different URLs,
e.g. https://events.pagerduty.com/generic/2010-04-15/create_event.json

src/robusta/core/sinks/pagerduty/pagerduty_sink.py
taivo123: Yes, I was inspired by another sink module (victorops or datadog etc) ;)

src/robusta/core/sinks/pagerduty/pagerduty_sink.py
taivo123: Source could be problematic when you are piping alerts from Global Prometheus (company level Prometheus) to Robusta and not from local (cluster level Prometheus) but I will do the change nevertheless, can always change back :)

docs/catalog/sinks/PagerDuty.rst
aantn: @taivo123 I think we can make it a constant. The other URL is for the [v1 API](https://developer.pagerduty.com/docs/ZG9jOjExMDI5NTc3-events-api-v1#making-a-request) and PagerDuty [recommends](https://support.pagerduty.com/docs/services-and-integrations#should-i-use-events-api-v1-or-v2) using the v2 API when possible.

src/robusta/core/sinks/pagerduty/pagerduty_sink.py
aantn: Thanks

src/robusta/core/sinks/pagerduty/pagerduty_sink.py
aantn: @taivo123 Oh, got it. It's possible to also send:

Global Prometheus -> Robusta Relay -> Relevant Robusta cluster -> PagerDuty

In which case the cluster_name will be correct.

If you need to send direct we can also look at other options or exposing a flag for this later.

Thanks for explaining.

src/robusta/core/sinks/pagerduty/pagerduty_sink.py
```diff
@@ -0,0 +1,124 @@
...
+                "class": finding.aggregation_key,
+                "custom_details": custom_details,
+            },
+            "routing_key": self.api_key,            
+            "event_action": PagerdutySink.__to_pagerduty_status_type(finding.title),
+            "dedup_key": finding.fingerprint
+        }
+        
+        headers = {"Content-Type": "application/json"}
+        response = requests.post(self.url, json=body, headers=headers)
```
aantn, line 102: I think we should log an error here. When testing I put in an incorrect API key and it silently fails without notifying the user.

You can do something like this:
```
 if response.status_code != 200:
   logging.error(f"Error sending message to PagerDuty: {response.status_code}, {response.reason}, {response.text}")
```

aantn: Whoops, my bad. You should check `response.ok` and not explicitly check for a status_code of 200. PagerDuty actually returns 202 errors, which are also OK.

PR #515, opened 2022-09-05
Modified docs_build.sh to support --port option

pavangudiwada:
Fixes **OSError: [Errno 98] Address already in use** error, caused when the default **sphinx-autobuild** port (8000) is reserved. 

image

### DIFF
--- a/docs_autobuild.sh
+++ b/docs_autobuild.sh
@@ -1,9 +1,13 @@
 #!/bin/bash
 
 cd docs && make clean && cd ..
+<<<<<<< HEAD
 
 if [ "$1" == "--port" ];then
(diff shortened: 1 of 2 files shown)


docs_autobuild.sh
```diff
@@ -1,4 +1,9 @@
 #!/bin/bash
 
 cd docs && make clean && cd ..
-poetry run sphinx-autobuild docs docs/_build/html || printf '\n\nError running docs_autobuild. \nMake sure you are using poetry >= 1.1.15. \nCheck out https://docs.robusta.dev/master/developer-guide/platform/docs-contributions.html\n'
+
+if [ "$1" == "--port" ];then
```
aantn, line 5: @pavangudiwada instead of the if/else can we always run `poetry run sphinx-autobuild $@ docs docs/_build/html` and such that `./docs_autobuild.sh` supports all `sphinx-autobuild` flags and passes them in automatically no matter what?

aantn (approved):

pavangudiwada (commented):
Made changes

PR #516, opened 2022-09-05
better explaining multiple cluster usage in the docs

Sheeproid:
Not filling clusterName by default, to emphasize re-use of the generated_config.yaml is preferred.
When install, the option --set clusterName should be used with the helm install command.

Updated docs to reflect this, and also explaining better about multi cluster install.

Screen Shot 2022-09-05 at 15 03 31
Screen Shot 2022-09-05 at 15 03 49

Screen Shot 2022-09-05 at 15 04 26


docs/installation.rst
```diff
@@ -78,20 +78,40 @@ Standard Installation
 
 3. Save ``generated_values.yaml``, somewhere safe. This is your Helm ``values.yaml`` file.
 
-4. Download the Helm chart and Install Robusta using Helm. On some clusters this can take a while [#f2]_, so don't panic if it appears stuck:
+.. admonition:: Multiple clusters
```
arikalon1, line 81: I would change this to:
Installing on multiple clusters

docs/installation.rst
```diff
@@ -78,20 +78,40 @@ Standard Installation
...
+    :class: important
+
+    Use the same ``generated_values.yaml`` for all your clusters (dev, prod, etc..). There's no need to run gen-config again.
+
+Standard Installation
+------------------------------
+
+1. Download the Helm chart:
```
arikalon1, line 89: I think download is the wrong wording.
You can download a local copy of the help chart, buy running `helm pull --untrar ...`

But the default way to install is just adding our helm repo, to the helm client, and work from there.

I would name this: Add Robusta's chart reppository
 (or something)

docs/installation.rst
```diff
@@ -78,20 +78,40 @@ Standard Installation
...
     helm repo add robusta https://robusta-charts.storage.googleapis.com && helm repo update
-    helm install robusta robusta/robusta -f ./generated_values.yaml
 
-5. Verify that Robusta is running two pods and there are no errors in the logs:
+2. Copy the cluster name from the current context and install Robusta using Helm. On some clusters this can take a while [#f2]_, so don't panic if it appears stuck:
+
+.. admonition:: Test clusters (e.g Kind, MiniCube, Colima)
+    :class: important
+
+    Test clusters tend to have fewer resources. To lower the resource requests of Robusta,
```
arikalon1, line 101: like

docs/installation.rst
```diff
@@ -78,20 +78,40 @@ Standard Installation
...
+    include ``--set isSmallCluster=true`` at the end of the install command.
+
+.. code-block:: bash
+   :name: cb-helm-install-only-robusta
+
+    helm install robusta robusta/robusta -f ./generated_values.yaml --set clusterName=<YOUR_CLUSTER_NAME> # --set isSmallCluster=true
+
+3. Verify that Robusta is running at least two pods and there are no errors in the logs:
```
arikalon1, line 109: Not sure about "running at pod"
Maybe:
Verify the two Robusta pods and running with no errors in the logs

docs/installation.rst
```diff
@@ -132,6 +151,24 @@ By default, Robusta sends notifications when Kubernetes pods crash.
 
    kubectl delete deployment crashpod
 
+Installing a second cluster
+---------------------------------
+
+When installing a second cluster on the same account, there's no need to run ``robusta gen-config`` again.
+Just follow the :ref:`Standard Installation` instruction using the ``generated_values.yaml`` file you've already created.
```
arikalon1, line 158: I would explicitly give the command here.

When installing a second cluster on the same account, there's no need to run ``robusta gen-config`` again.
Using your existing ``generated_values.yaml`` and the new `clusterName` run:

helm install robusta robusta/robusta -f ./generated_values.yaml --set clusterName=<YOUR_NEW_CLUSTER_NAME>

docs/installation.rst
```diff
@@ -132,6 +151,24 @@ By default, Robusta sends notifications when Kubernetes pods crash.
...
+
+.. admonition:: Where is my generated_values.yaml?
+
+    If you have lost your ``generated_values.yaml`` file, you can extract it from a cluster.
```
arikalon1, line 162: "you can extract it from any cluster running Robusta"

src/robusta/cli/main.py
```diff
@@ -89,31 +86,6 @@ class HelmValues(BaseModel, extra=Extra.allow):
     runner: Dict = None
     rsa: RSAKeyPair = None
 
-    def set_pod_configs_for_small_clusters(self):
-        self.kubewatch = PodConfigs.gen_config(FORWARDER_CONFIG_FOR_SMALL_CLUSTERS)
```
arikalon1, line 93: I think this was the only use of the PodConfigs class.
We can probably remove it

arikalon1 (approved):
Great work!
Left few minor comments

docs/installation.rst
Sheeproid: there are more pods for robusta if you install prometheus

docs/installation.rst
```diff
@@ -78,20 +78,40 @@ Standard Installation
...
-   :name: cb-helm-install-robusta
+   :name: cb-helm-repo-add-update-robusta
 
     helm repo add robusta https://robusta-charts.storage.googleapis.com && helm repo update
-    helm install robusta robusta/robusta -f ./generated_values.yaml
 
-5. Verify that Robusta is running two pods and there are no errors in the logs:
+2. Copy the cluster name from the current context and install Robusta using Helm. On some clusters this can take a while [#f2]_, so don't panic if it appears stuck:
+
+.. admonition:: Test clusters (e.g Kind, MiniCube, Colima)
```
arikalon1, line 98: Should be MiniKube

PR #518, opened 2022-09-06
Minikube warning and sink configuration

pavangudiwada:
1. Added a link to Sink configuration before Robusta config generation.
2. Added a warning about Robusta runner remaining in ContainerCreating state.

image

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -73,10 +73,10 @@ If you’d like to send Robusta messages to additional destinations (Discord, Te
 
     robusta gen-config
 
-.. admonition:: Robusta runner ImagePullBackOff
+.. admonition:: Robusta on Minikube
     :class: warning
 
-    On minikube `robusta-runner` could remain in `ContainerCreating` state due to this `issue <https://github.com/kubernetes/minikube/issues/14806>`_. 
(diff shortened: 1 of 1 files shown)


aantn:
@pavangudiwada I think the error for being stuck in `ContainerCreating` state happens at a later stage, right? Not in the gen-config stage.

pavangudiwada:
> @pavangudiwada I think the error for being stuck in `ContainerCreating` state happens at a later stage, right? Not in the gen-config stage.

Correct.  But a user will be asked details in the context of **local cluster**(minikube) in the `gen-config` stage. If they are not aware of this issue. They will go ahead with the process, and it ends up not working.  

In which case, they should regenerate/ modify the `generated_values.yaml` file etc. and go through the hassle of choosing a new cluster.

aantn:
@pavangudiwada  Oh, right. I forgot about that.

Can we change the wording then and just say "Don't install Robusta on Minikube" or something like that. Right now we're describing an error here which isn't relevant to the current stage. We're missing the main message which is "We recommend you don't install Robusta on Minikube..."

In any event, @Sheeproid is working on a change that will make it possible to put the message in the docs at a later stage. See https://github.com/robusta-dev/robusta/pull/516/files

aantn:
@pavangudiwada perfect.

aantn (approved):

aantn:
@Sheeproid FYI

PR #523, opened 2022-09-07
Two fixes: (1) Add FindingSubject.__str__ (2) Silence alert on resolve too

aantn:
Sorry about the PR on two unrelated topics. I accidentally committed both of them together and because they're both tiny decided to leave them together.

(1) Is important for the PagerDuty sink where we convert the FindingSubject to string. We should have a standard representation for it.

(2) Fixes a bug we spoke about.


aantn:
Note that Black autoformatted the whole file `src/robusta/core/reporting/base.py` but I only changed (added) the __str__ function there.

arikalon1 (approved):
LGTM

PR #527, opened 2022-09-08
Added playbook http_action and persistent_volume_action

alikhanxgrid:
#475 
#222


playbooks/robusta_playbooks/persistent_volume_actions.py
```diff
@@ -0,0 +1,170 @@
...
+                        volume_mount_name = volume.name
+                        break
+
+                # Use name of the mounted volume to find the correct volume mount
+                container_found_flag = False
+                container_volume_mount = None
+                for container in pod.spec.containers:
+                    if (container_found_flag):
+                        break
+                    if(container.volumeMounts):
```
arikalon1, line 51: I think you can remove that if, because volumeMount is a list
So, if it's empty nothing happens in the for loop

playbooks/robusta_playbooks/persistent_volume_actions.py
```diff
@@ -0,0 +1,170 @@
...
+                finding.add_enrichment(
+                    [
+
+                        FileBlock(
+                            f"Data.txt: ", result.encode()),
+
+                    ]
+                )
+                # delete the reader pod
+                reader_pod.delete()
```
arikalon1, line 84: I would put the delete pod, in a `finally` clause.
So if there's an exception, we'll still delete the reader pod, and not leave it in the cluster

playbooks/robusta_playbooks/persistent_volume_actions.py
```diff
@@ -0,0 +1,170 @@
...
+
+        else:
+            finding.add_enrichment(
+                [
+                    MarkdownBlock(
+                        f"ERROR: Persistent volume named {pv.metadata.name} have no persistent volume claim."
```
arikalon1, line 90: I would remove the word ERROR

playbooks/robusta_playbooks/persistent_volume_actions.py
```diff
@@ -0,0 +1,170 @@
...
+                    ),
+                ]
+            )
+
+    except RuntimeError as e:
```
arikalon1, line 95: Usually, when there's an exception, we don't send it to the user.
Most user will not know what to do with it.

In the framework, we have code that except any exception thrown from an action, and adds some "user friendly" text.
So I think you can remove the except sections here

arikalon1 (commented):
Well done. Great work
Left few minor comments.

playbooks/robusta_playbooks/http_actions.py
```diff
@@ -0,0 +1,152 @@
+import json
```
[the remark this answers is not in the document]
alikhanxgrid: Resolved.

playbooks/robusta_playbooks/http_actions.py
```diff
@@ -0,0 +1,152 @@
...
+    except Exception as e:
+        finding.title = f"{action_params.url} is un-reachable"
+        finding.add_enrichment(
+            [
+                MarkdownBlock(
+                    f"*Error*\n```\n{e}\n```"
+                ),
+            ]
+        )
+    if(action_params.get_response):
```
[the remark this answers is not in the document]
alikhanxgrid: Yes didn't see that thanks. 
resolved. Let me know if the current change is good or get_response needs to go completely.

playbooks/robusta_playbooks/http_actions.py
```diff
@@ -0,0 +1,152 @@
...
```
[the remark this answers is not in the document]
alikhanxgrid: Resolved

playbooks/robusta_playbooks/persistent_volume_actions.py
alikhanxgrid: Makes sense. 
Should I just remove the exceptions completely or log the exceptions here and not send them to sinks instead which can help easier debugging of action in future.

arikalon1: Please take a look at this line:
https://github.com/robusta-dev/robusta/blob/5a404cbde06ff5635d0e3fc20e4b295dfb5310b1/src/robusta/core/playbooks/playbooks_event_handler_impl.py#L229

We catch and print any exception thrown from the action, and add an enrichment saying: "Oops... Error processing {action.action_name}"

So the exception is already logged, no need to catch and print it

playbooks/robusta_playbooks/persistent_volume_actions.py
alikhanxgrid: Resolved

playbooks/robusta_playbooks/persistent_volume_actions.py
alikhanxgrid: Resolved

playbooks/robusta_playbooks/persistent_volume_actions.py
alikhanxgrid: Resolved. 
Check if the current change is ok. Feel free to tell me anything.

playbooks/robusta_playbooks/persistent_volume_actions.py
```diff
@@ -0,0 +1,170 @@
...
+
+        if pv_claimref != None:
+            # Do this if there is a PVC attached to PV
+            pvc_obj = PersistentVolumeClaim.readNamespacedPersistentVolumeClaim(
+                name=pv_claimref.name, namespace=pv_claimref.namespace).obj
+            pod_obj = get_pod_related_to_pvc(pvc_obj, pv)
+            if (pod_obj != None):
+                # Do this if a Pod is using PVC
+
+                pod = RobustaPod.find_pod(
```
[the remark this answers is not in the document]
alikhanxgrid: Resolved.

playbooks/robusta_playbooks/persistent_volume_actions.py
```diff
@@ -0,0 +1,170 @@
...
+        failure=False,
+    )
+
+    # Get persistent volume data the object contains data related to PV like metadata etc
+    pv = event.get_persistentvolume()
+    pv_claimref = pv.spec.claimRef
+
+    try:
+
+        if pv_claimref != None:
```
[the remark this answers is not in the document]
alikhanxgrid: Resolved.

playbooks/robusta_playbooks/persistent_volume_actions.py
```diff
@@ -0,0 +1,170 @@
...
+    finding = Finding(
+        title=f"Persistent Volume content",
+        source=FindingSource.MANUAL,
+        aggregation_key=function_name,
+        finding_type=FindingType.REPORT,
+        failure=False,
+    )
+
+    # Get persistent volume data the object contains data related to PV like metadata etc
+    pv = event.get_persistentvolume()
```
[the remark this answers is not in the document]
alikhanxgrid: Resolved. 
Added an if condition before this statement.

arikalon1 (approved):
Great work @alikhanxgrid 
Thank you for your contribution

PR #529, opened 2022-09-09
Fix edge case where `robusta logs` failed

aantn:
The `robusta logs` command is broken on the following edge case and this PR fixes it.

# The Edge Case
I use spot instances on one of my clusters. This led to a situation where a node on which Robusta was running was terminated and I ended up with three Robusta pods (obviously only one of which is running):

Screen Shot 2022-09-09 at 17 00 50

Screen Shot 2022-09-10 at 1 13 57

### DIFF
--- a/src/robusta/cli/main.py
+++ b/src/robusta/cli/main.py
@@ -1,5 +1,6 @@
 import base64
 import json
+import logging
 import random
 import subprocess
 import time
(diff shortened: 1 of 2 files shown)


aantn:
Note that this change is smaller than it looks. A big part of the merge is just black autoformatting.

arikalon1 (approved):
Nice work

PR #530, opened 2022-09-10
fixed memory spec parse

AndrewChubatiuk:
```2022-09-10 05:46:26.297 ERROR    error parsing memory 268435456
  File "/usr/local/lib/python3.9/site-packages/robusta/core/model/pods.py", line 98, in get_number_of_bytes_from_kubernetes_mem_spec
    raise Exception("number of bytes could not be extracted from memory spec: " + mem_spec)
Exception: number of bytes could not be extracted from memory spec: 268435456```

### DIFF
--- a/src/robusta/core/model/pods.py
+++ b/src/robusta/core/model/pods.py
@@ -96,7 +96,7 @@ class PodResources(BaseModel):
                 return int(mem_spec[:-1]) * k8s_memory_factors[mem_spec[-1]]
 
             try:
-                return float(mem_spec)
+                return int(mem_spec)
             except ValueError:
                 raise Exception("number of bytes could not be extracted from memory spec: " + mem_spec)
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/robusta/core/model/pods.py [hunk 1]
```diff
@@ -95,6 +95,9 @@ def get_number_of_bytes_from_kubernetes_mem_spec(mem_spec: str) -> int:
             if len(mem_spec) > 1 and mem_spec[-1] in k8s_memory_factors:
                 return int(mem_spec[:-1]) * k8s_memory_factors[mem_spec[-1]]
 
+            if mem_spec.isdigit():
+                return int(mem_spec)
+
```
arikalon1, line 100: If we're already fixing it, I would also add support for memory spec with the format of "129e6". (This format can be parsed to float)
Looks like this format is also allowed:
https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-memory

arikalon1 (commented):
Thank you for your contribution @AndrewChubatiuk 
The fix looks good.

Can you please sign the CLA?

arikalon1 (approved):

arikalon1:
Thank you for your contribution @AndrewChubatiuk

PR #539, opened 2022-09-13
Overcommited enrichments

anFatum:
That's a PR without actual links to YouTube channel, and there may be some message edits
I want to be sure that my approach is okay before the content is submitted


src/robusta/integrations/resource_analysis/memory_analyzer.py
```diff
@@ -106,3 +96,35 @@ def get_number_of_bytes_from_kubernetes_mem_spec(mem_spec: str) -> int:
...
+    """
+    factor, suffix = (None, None)
+    for factor, suffix in UNITS_MAPPING:
+        if total_bytes >= factor:
+            break
+    if not (factor or suffix):
+        return total_bytes
+    amount = round(total_bytes / factor, 2)
+
+    if isinstance(suffix, tuple):
```
arikalon1, line 124: I would add a comment to explain what is this special case

src/robusta/integrations/resource_analysis/prometheus_analyzer.py
```diff
@@ -0,0 +1,49 @@
...
+        )
+        return results
+
+    def _get_query_value(self, results: Optional[list], offset: int = 0) -> Optional[float]:
+        if not results:
+            return None
+        return float(results[offset].get("value", [None, None])[1])
+
+    def _timed_query(self, promql_query: str, duration: timedelta) -> Optional[list]:
+        if not self.prometheus_tzinfo:
```
arikalon1, line 33: Can we choose some default timezone?
Why is it important to have timezone?

If it's a mandatory param, maybe it's better to add it to the method parameters

arikalon1 (approved):
Great work!
Left few minor comments.

I think we don't send the video links to the UI
There were some other sinks we skipped I think.
Is there a specific reason?

src/robusta/integrations/resource_analysis/prometheus_analyzer.py
```diff
@@ -0,0 +1,49 @@
+import logging
+from datetime import datetime, timedelta, tzinfo
+from typing import Optional
+from prometheus_api_client import PrometheusConnect
+
+from ..prometheus.utils import PrometheusDiscovery
+from ...core.model.env_vars import PROMETHEUS_REQUEST_TIMEOUT_SECONDS
+
+
+class PrometeusAnalyzer:
```
arikalon1, line 10: Prometheus has a typo here

playbooks/robusta_playbooks/overcommit_enrichments.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+@action
+def cpu_overcommited_enricher(alert: PrometheusKubernetesAlert, params: PrometheusParams):
+    """
+    Enrich the finding with a detailed explanation for the cause of the CPU overcommitment.
+    Includes recommendations for the identified cause.
+    """
+    cpu_analyzer = CpuAnalyzer(params.prometheus_url)
+    cpu_requests = cpu_analyzer.get_total_cpu_requests(timedelta(minutes=10))
```
arikalon1, line 13: 10 minutes sounds good to me, but I'm not 100% sure we won't have to change it.
Would it be better to add it as a parameter to the playbook, with default value of 10?

src/robusta/core/reporting/base.py
```diff
@@ -37,14 +39,27 @@ def from_severity(severity: str) -> "FindingSeverity":
...
+            return "🟠"
+        elif self == FindingSeverity.HIGH:
+            return "🔴"
+
 
+class VideoLink(BaseModel):
+    url: str
+    name: str = "See more"
+
+    def to_json(self):
```
arikalon1, line 60: Pydantic has builtin to_json
When using BaseModel we get to/from json/yaml for free

PR #545, opened 2022-09-17
failed jobs actions

arikalon1:
Actions to gather extra information on failed jobs.
Triggered by default by 2 alerts: KubeJobFailed, KubeJobCompletion

failed-job-enrichments


playbooks/robusta_playbooks/job_actions.py
```diff
@@ -125,4 +125,108 @@ def job_events_enricher(event: JobEvent, params: EventEnricherParams):
...
+    job = event.get_job()
+    if not job:
+        logging.error(
+            f"cannot run job_pod_enricher on alert with no job object: {event}"
+        )
+        return
+
+    pod = get_job_latest_pod(job)
+
+    if pod:
```
Avi-Robusta, line 153: i would do 
```
if not pod:
    logging.error("Failed to locate pod for job...")
    return
```

playbooks/robusta_playbooks/job_actions.py
```diff
@@ -125,4 +125,108 @@ def job_events_enricher(event: JobEvent, params: EventEnricherParams):
...
+                pod.metadata.name,
+                pod.metadata.namespace,
+                included_types=params.included_types,
+                max_events=params.max_events,
+            )
+            if events_table_block:
+                event.add_enrichment([events_table_block], {SlackAnnotations.ATTACHMENT: True})
+
+        if params.logs:
+            log_data = pod.get_logs()
```
Avi-Robusta, line 167: if there are multiple containers in the job will this give us all of the logs or just or the first container?
(like if the initContainer causes the event or if there is multiple containers)
Have you tested this on jobs with multiple containers? (I dont know if its common since I havn't ever needed to used jobs)

playbooks/robusta_playbooks/job_actions.py
```diff
@@ -125,4 +125,108 @@ def job_events_enricher(event: JobEvent, params: EventEnricherParams):
...
+    failed = job_status.failed if job_status.failed else 0
+    job_rows: List[List[str, str]] = [
+        ["status", __job_status_str(job_status)],
+        ["completions", f"{succeeded}/{job_spec.completions}"],
+        ["failures", f"{failed}"],
+        ["backoffLimit", f"{job_spec.backoffLimit}"],
+        ["duration", f"{job_status.startTime} - {end_time}"],
+        ["containers", "------------------"]
+    ]
+    for container in job_spec.template.spec.containers:
```
Avi-Robusta, line 200: do you think we should include job_spec.template.spec.initContainers also?

playbooks/robusta_playbooks/job_actions.py
```diff
@@ -125,4 +125,108 @@ def job_events_enricher(event: JobEvent, params: EventEnricherParams):
...
+            if log_data:
+                event.add_enrichment(
+                    [FileBlock(f"{pod.metadata.name}.log", log_data.encode())],
+                )
+
+
+@action
+def job_info_enricher(event: JobEvent):
+    """
+    Given a Kubernetes job, add information about the job, from the job spec and status.
```
Avi-Robusta, line 177: Dont know if this is important for this PR or not, just asking to see what you think, maybe add it to notion if you think its out of scope:
is there a way to determine if a rule in `spec.podFailurePolicy.rules` caused the job to fail and which one?
since the job could fail due to bad rules configured like this example where we fail on returning 0
 ```
 podFailurePolicy:
    rules:
    - action: FailJob
      onExitCodes:
        containerName: mycontainer
        operator: In             
        values: [0]

```

File: playbooks/robusta_playbooks/job_actions.py [hunk 5]
```diff
@@ -125,4 +125,108 @@ def job_events_enricher(event: JobEvent, params: EventEnricherParams):
...
+        return
+
+    job_status: JobStatus = job.status
+    job_spec: JobSpec = job.spec
+
+    end_time = job_status.completionTime if job_status.completionTime else "None"
+    succeeded = job_status.succeeded if job_status.succeeded else 0
+    failed = job_status.failed if job_status.failed else 0
+    job_rows: List[List[str, str]] = [
+        ["status", __job_status_str(job_status)],
```
Avi-Robusta, line 193: does JobCondition.message give any information that could be useful for users?

Avi-Robusta:
looks good

playbooks/robusta_playbooks/job_actions.py
arikalon1: it gives us the logs of the first container.
This is the way we do it everywhere with the logs_enricher.
It's not perfect, but I think it's ok

playbooks/robusta_playbooks/job_actions.py
arikalon1: I wasn't aware of that field.
It's alpha on k8s 1.25
I don't think we need to add it yet

playbooks/robusta_playbooks/job_actions.py
arikalon1: Fixing

playbooks/robusta_playbooks/job_actions.py
arikalon1: good point, fixing

playbooks/robusta_playbooks/job_actions.py
arikalon1: good point, I'll add it

PR #547, opened 2022-09-18
Prometheus manual trigger

Avi-Robusta:
Added trigger to run prometheus queries via robusta.

This is an example dataset for ram usage of a pod
Screen Shot 2022-09-18 at 11 03 29

for the manual trigger 

curl -X POST http://localhost:5000/api/trigger -H 'Content-Type: application/json' -d '{"action_name": "prometheus_enricher", "action_params": {"promql_query": "sum(container_memory_working_set_bytes{job=\"kubelet\", metrics_path=\"/metrics/cadvisor\", pod=~\"tetris-79c4c6958b-hwmlk\", container!=\"\", image!=\"\"})", "end_time": "2022-09-17 09:40:59", "starts_at": "2022-09-14 09:40:59", "metadata_json_str": "{\"units\":\"bytes\",\"query_type\":\"resource-memory\",\"object_queried\":{\"type\":\"Pod\",\"name\":\"tetris-79c4c6958b-hwmlk\"}}"}}'


playbooks/robusta_playbooks/prometheus_enrichments.py
```diff
@@ -0,0 +1,44 @@
+from robusta.api import *
+
+
+def parse_timestamp_string(date_string: str) -> datetime:
```
arikalon1, line 4: If we return None, the return type should probably be Optional[datetime]

playbooks/robusta_playbooks/prometheus_enrichments.py
```diff
@@ -0,0 +1,44 @@
...
+
+@action
+def prometheus_enricher(event: ExecutionBaseEvent, params: PrometheusQueryParams):
+    """
+        Enriches the finding with a prometheus query
+
+        for example prometheus queries see here:
+        https://prometheus.io/docs/prometheus/latest/querying/examples/
+    """
+    if not params.promql_query or (not params.graph_duration_minutes and not params.starts_at):
```
arikalon1, line 22: I think promql_query should be mandatory in the params.

I'm not sure what is the allowed configuration options.
Is it either `graph_duration_minutes` or both `starts_at` and `ends_at`?
Is there a way to express that in the Pydantic class?

```
class PrometheusDuration(BaseModel):
    duration_minutes: int


class PrometheusDateRange(BaseModel):
    starts_at: str
    ends_at: str


class PrometheusQueryParams(PrometheusParams):
    """
    :var promql_query: the prometheusql query you want to run
    :var graph_duration_minutes: the amount of minutes back you want results for - this is not needed if starts_at is defined
    :var starts_at: the start date/time for the query - this is not needed if graph_duration_minutes is defined
    :var end_time: the end time for the query, if none is set than it defaults to now

    both starts at and ends_at use format "%Y-%m-%d %H:%M:%S"

    """
    promql_query: str
    duration: Union[PrometheusDateRange, PrometheusDuration]
```

src/robusta/core/playbooks/prometheus_enrichment_utils.py
```diff
@@ -32,33 +33,47 @@ def get_node_internal_ip(node: Node) -> str:
...
     prom = PrometheusConnect(url=prometheus_base_url, disable_ssl=True)
-    end_time = datetime.now(tz=starts_at.tzinfo)
+    if not starts_at:
+        raise Exception("No start time specified for query.")
+    if not end_time:
+        end_time = datetime.now(tz=starts_at.tzinfo)
     alert_duration = end_time - starts_at
     graph_duration = max(alert_duration, timedelta(minutes=graph_duration_minutes))
-    start_time = end_time - graph_duration
+    start_time = end_time - max(alert_duration, graph_duration)
```
arikalon1, line 52: graph_duration is bigger than alert_duration, see one line above

src/robusta/core/playbooks/prometheus_enrichment_utils.py
```diff
@@ -32,33 +33,47 @@ def get_node_internal_ip(node: Node) -> str:
...
     resolution = 250  # 250 is used in Prometheus web client in /graph and looks good
     increment = max(graph_duration.total_seconds() / resolution, 1.0)
-    result = prom.custom_query_range(
+    logging.warning(f" resolution : {graph_duration.total_seconds()/ increment}, seconds {graph_duration.total_seconds()} inc {increment}")
```
arikalon1, line 55: Why do we always print that warning?

arikalon1 (commented):
Nice work!
Left few comments

src/robusta/core/playbooks/prometheus_enrichment_utils.py
Avi-Robusta: that was a log that needed to be removed

src/robusta/core/sinks/robusta/dal/model_conversion.py
```diff
@@ -82,6 +82,8 @@ def to_evidence_json(
                 structured_data.append({"type": "header", "data": block.text})
             elif isinstance(block, ListBlock):
                 structured_data.append({"type": "list", "data": block.items})
+            elif isinstance(block, JsonBlock):
```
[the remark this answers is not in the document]
Avi-Robusta: Why weren't we sending Json blocks to the UI?
Maybe it makes more sense to add a variable in base block to not send an enrichment to the UI instead of duplicating the block types?

src/robusta/core/model/prometheus.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+class PrometheusScalarValue(BaseModel):
+    """
+        While usually this is a list of size 2 in the form of [float_timestamp, str_returned_value]
+        the list size and return value types are not guaranteed
+    """
+    timestamp: float
+    value: str
+
```
Avi-Robusta, line 24: make more clear

playbooks/robusta_playbooks/prometheus_enrichments.py
```diff
@@ -0,0 +1,45 @@
...
+
+        for example prometheus queries see here:
+        https://prometheus.io/docs/prometheus/latest/querying/examples/
+    """
+    if not params.promql_query:
+        logging.error(f"Invalid or missing parameter, prometheus_enricher requires a promql query.")
+        return
+    starts_at, ends_at = parse_duration(params.duration)
+    if not starts_at or not ends_at:
+        logging.error(f"Invalid duration params, verify the times are of format '%Y-%m-%d %H:%M:%S %Z'")
```
arikalon1, line 39: I think we should through an error here.
If it's only returns, I think it won't work well with the caller (manual trigger here probably)
It's better to throw an error saying this is an invalid request

src/robusta/core/external_apis/prometheus/models.py
```diff
@@ -0,0 +1,111 @@
...
+        value = str(raw_scalar_list[1])
+        super().__init__(timestamp=timestamp, value=value)
+
+
+class PrometheusVector(BaseModel):
+
+    metric: PrometheusMetric
+    value: PrometheusScalarValue
+
+    def __init__(self, **kwargs):
```
arikalon1, line 41: Can you explain why we need this init function?
Wouldn't it work without it?

src/robusta/core/external_apis/prometheus/models.py
```diff
@@ -0,0 +1,111 @@
+import logging
+
+from pydantic import BaseModel
+from typing import List, Dict, Optional
+
+
+"""
+    Known commonly returned keys in dictionary 'metric':
+    __name__: the name of the outer function in prometheus 
+   Other keys dictionary 'metric' sometimes contains:
```
arikalon1, line 10: I think these are the metrics labels

src/robusta/core/reporting/blocks.py
```diff
@@ -366,3 +367,22 @@ def __init__(self, choices: Dict[str, CallbackChoice]):
...
+
+class PrometheusBlock(BaseBlock):
+    """
+    Formatted prometheus query results with metadata
+    """
+
+    data: PrometheusQueryResult
+    metadata: Dict[str, str]
+
+    def __init__(self, data: PrometheusQueryResult, query: str = None):
```
arikalon1, line 380: I would make the query mandatory, and remove the if below
I think it is mandatory anyway

arikalon1 (approved):
Great job.
I left few minor comments

The only thing I'm missing is sanitation of the promql_query we're receiving

playbooks/robusta_playbooks/prometheus_enrichments.py
```diff
@@ -0,0 +1,45 @@
...
+
+@action
+def prometheus_enricher(event: ExecutionBaseEvent, params: PrometheusQueryParams):
+    """
+        Enriches the finding with a prometheus query
+
+        for example prometheus queries see here:
+        https://prometheus.io/docs/prometheus/latest/querying/examples/
+    """
+    if not params.promql_query:
```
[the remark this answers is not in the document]
Avi-Robusta: While the field is mandatory you can still pass it an empty string "" which we don't want to waste time running

PR #552, opened 2022-09-19
removing config map changes by default

Avi-Robusta:
-- breaking changes - Removed config-map playbooks from built in playbook
need to alert users of change
this was removed due to spammy config map changes were sent to UI


arikalon1 (approved):
nice

PR #555, opened 2022-09-20
Self host cli and docs

RoiGlinik:
This is adds the new CLI command to generate the self host values + backend config.
Also migrated the guides to the docs.


docs/self-host/bare.rst
```diff
@@ -0,0 +1,117 @@
...
+
+Login to the UI
+^^^^^^^^^^^^^^^^
+
+Go to ``https://platform.MY_DOMAIN`` and sign up. Then try to log in.
+
+Support
+^^^^^^^^^^^^^^^^
+
+For further help contact us on `Slack <https://join.slack.com/t/robustacommunity/shared_invite/zt-10rkepc5s-FnXKvGjrBmiTkKdrgDr~wg>`_ or `support@robusta.dev <mailto:support@robusta.dev>`_
```
arikalon1, line 117: The join slack link is temporary.
I wouldn't use it here.
You can point the reader to join via the robusta home page

docs/self-host/gke.rst
```diff
@@ -0,0 +1,149 @@
...
+
+Login to the UI
+^^^^^^^^^^^^^^^^
+
+Go to ``https://platform.MY_DOMAIN`` and sign up. Then try to log in.
+
+Support
+^^^^^^^^^^^^^^^
+
+For further help contact us on `Slack <https://join.slack.com/t/robustacommunity/shared_invite/zt-10rkepc5s-FnXKvGjrBmiTkKdrgDr~wg>`_ or `support@robusta.dev <mailto:support@robusta.dev>`_
```
arikalon1, line 149: same, slack join is temporary

docs/self-host/google-sso.rst
```diff
@@ -0,0 +1,60 @@
...
+   :name: cb-google-sso-yaml-values
+
+    ENABLE_GOOGLE_SIGNUP: true
+    GOOGLE_CLIENT_ID: <your-client-id>
+    GOOGLE_CLIENT_SECRET: <your-client-secret>
+
+Support
+^^^^^^^^^^^^^^^
+
+For further help contact us on `Slack <https://join.slack.com/t/robustacommunity/shared_invite/zt-10rkepc5s-FnXKvGjrBmiTkKdrgDr~wg>`_ or `support@robusta.dev <mailto:support@robusta.dev>`_
```
arikalon1, line 60: same

docs/self-host/slack.rst
```diff
@@ -0,0 +1,80 @@
...
+#. Click create new app (on top right corner).
+#. Choose “From app manifest”  (down on the left menu).
+#. Choose your desired workspace. (step 1)
+#. Choose JSON format (step 2), and paste your copy of the app template into the form.
+#. Select create (step 3)
+#. Go to your app page -> Basic Information
+#. Find ``Install your app`` and install it to the relevant workspace.
+#. Go to ``Display Information`` and add our :download:`App icon </_static/app-logo.png>`
+
+Adjust your self_host_values.yaml file
```
arikalon1, line 72: I think it's a little too verbose
I would just write:
```
Use the created app credentials
---------------------------------

#. Go to your app page -> Basic Information
#. In the App Credentials section find: ``Client ID``, ``Client Secret`` and ``Signing Secret``.
#. Open the self_host_values.yaml file
#. replace ``slackClientId``, ``slackClientSecret`` and ``slackSigningSecret`` respectively.

```

src/robusta/cli/self_host.py
```diff
@@ -0,0 +1,180 @@
...
+            backendProfile.dict(exclude={"custom_profile"}), output_file, indent=1
+        )
+        typer.secho(
+            f"Saved configuration to {backendconfig_path} - save this file for future use!",
+            fg="red",
+        )
+
+
+class RobustaUI(BaseModel):
+    RELAY_HOST: str = ""
```
arikalon1, line 42: Why do we need the default value "" here?

src/robusta/cli/self_host.py
```diff
@@ -0,0 +1,180 @@
...
+    SUPABASE_URL: str = ""
+    SUPABASE_KEY: str = ""
+    service = {"nodePort": 30311}  # platform.domain
```
arikalon1, line 45: Should we allow overriding it in the CLI?

src/robusta/cli/self_host.py
```diff
@@ -0,0 +1,180 @@
...
+    storePassword: str = ""
+    storeUser: str = "apiuser-robustarelay@robusta.dev"
+    storeUrl: str = ""
+    storeApiKey: str = ""  # anon key
+    slackClientId: str = "your-client-id"
+    slackClientSecret: str = "your-client-secret"
+    slackSigningSecret: str = "your-signing-secret"
+    syncActionAllowedOrigins: str = ""
+    provider: str = ""
+    apiNodePort: int = 30313  # api.domain
```
arikalon1, line 66: should we allow oveeride ?

src/robusta/cli/self_host.py
```diff
@@ -0,0 +1,180 @@
...
+    STATIC_IP_NAME: str = "robusta-platform-ip"
+    RELAY_PASSWORD: str = gen_secret(12)
+    RELAY_USER: str = "apiuser-robustarelay@robusta.dev"
+    DOMAIN: str = ""
+    PROVIDER: str = ""
+    # SUPABASE
+    JWT_SECRET: str = gen_secret(32)
+    JWT_EXPIRY: int = 3600
+    ANON_KEY: str = JWT.encode(
+        {"role": "anon", "iss": "supabase", "iat": 1661029200, "exp": 1818795600},
```
arikalon1, line 90: I would comment what iss/iat/exp means
And I would put the values in constants with meaningful names. 
For example: `1_DAY_EXPIRATION = 1818795600`

src/robusta/cli/self_host.py
```diff
@@ -0,0 +1,180 @@
...
+            "exp": 1818795600,
+        },
+        JWT_SECRET,
+    )
+    SUPABASE_URL: str = "http://kong:8000"  # Internal URL
+    PUBLIC_REST_URL: str = ""  ## Studio Public REST endpoint - replace this if you intend to use Studio outside of localhost
+
+    # POSTGRES
+    POSTGRES_PORT: int = 5432
+    POSTGRES_STORAGE: str = "50Gi"
```
arikalon1, line 107: Not sure if we decided 50Gi
Maybe using 100Gi is a better default

Do we also want to expose the `storageClass`?

src/robusta/cli/self_host.py
```diff
@@ -0,0 +1,180 @@
...
+
+
+app = typer.Typer()
+
+
+@app.command()
+def gen_config(
+    provider: str = typer.Option(
+        "",
+        help="Cloud host provider.",
```
arikalon1, line 144: I would add the allowed values to the docs. i.e. gke/on-prem

I think this should be a mandatory parameter with no default.
So the user must specify if it's on-prem or gke

src/robusta/cli/self_host.py
```diff
@@ -0,0 +1,180 @@
...
+    ),
+    domain: str = typer.Option("", help="domain used to route the self host services."),
+):
+    """Create self host configuration file"""
+
+    if provider not in {"", "gke"}:
```
arikalon1, line 150: I would use `on-prem` instead of ""

In addition, you're allowing empty `domain` which is illegal I think.
I would make it mandatory, with no default.
So you can't run without a `domain`

arikalon1 (approved):
Great work!
Left some minor comments.
Feel free to fix and merge

docs/self-host/gke.rst
RoiGlinik: ill use the bit.ly link instead

src/robusta/cli/self_host.py
RoiGlinik: removed here

src/robusta/cli/self_host.py
RoiGlinik: will add

src/robusta/cli/self_host.py
RoiGlinik: updated to 100 and added the storage class as well

src/robusta/cli/self_host.py
RoiGlinik: sounds good

PR #558, opened 2022-09-28
Added new query params to investigate link

anFatum:


src/robusta/core/reporting/base.py
```diff
@@ -187,11 +189,32 @@ def attribute_map(self) -> Dict[str, str]:
...
+            return "graphs"
+        if self.service.resource_type.lower() == "job":
+            return "jobs"
+        return "services"
+
+    def get_investigate_uri(self, account_id: str, cluster_name: Optional[str] = None):
+        uri_path = self._map_service_to_uri()
+        params = {
+            "account": account_id,
+            "clusters": f"[\"{cluster_name}\"]" if cluster_name else None,
```
arikalon1, line 203: if it's always a single cluster, maybe it's better to call it `cluster` and `namespace` ?

src/robusta/core/reporting/base.py
```diff
@@ -187,11 +189,32 @@ def attribute_map(self) -> Dict[str, str]:
...
+            "namespaces": f"[\"{self.subject.namespace}\"]" if self.subject.namespace else None,
+            "kind": self.service.resource_type if self.service else None,
+            "name": self.service.name if self.service else None,
```
arikalon1, line 206: I think this might be confusing
`name` is ok I think
but `names` is a little confusing.
can we call it `key` or something?

arikalon1 (approved):
Great work.
Left few minor comments

src/robusta/core/reporting/base.py
anFatum: That’s how parameters are passed on the frontend in the query string, that’s why it is in plural

src/robusta/core/reporting/base.py
anFatum: The same thing - these are the parameters in the query string

src/robusta/core/reporting/base.py
arikalon1: got it

src/robusta/core/reporting/base.py
arikalon1: got it

PR #559, opened 2022-10-02
Simplified Dockerfile with a little reduced size

pavangudiwada:
- Added `--no-dev` for all poetry install commands, helps save space and time by not installing Dev dependencies (sphinx)
- Moved some commands around, this doesn't have an impact on the size of the image but saves time spent on updating the Repositories multiple times.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,10 +2,8 @@
 FROM python:3.9-slim
 RUN apt-get update \
     && dpkg --add-architecture arm64 \
-    && apt-get install -y --no-install-recommends git ssh socat wget curl libcairo2 python3-dev libffi-dev socat gcc \
+    && apt-get install -y --no-install-recommends git ssh socat wget curl libcairo2 python3-dev libffi-dev socat \
     && pip3 install --no-cache-dir --upgrade pip \
-    && pip3 install --no-cache-dir ruamel.yaml.clib \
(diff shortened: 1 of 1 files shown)


aantn:
@RoiGlinik re-assigning to you

Dockerfile
```diff
@@ -2,39 +2,26 @@
 FROM python:3.9-slim
 RUN apt-get update \
     && dpkg --add-architecture arm64 \
-    && apt-get install -y --no-install-recommends git ssh socat wget curl libcairo2 python3-dev libffi-dev socat \
-    && apt-get purge -y --auto-remove \
+    && apt-get install -y --no-install-recommends git ssh socat wget curl libcairo2 python3-dev libffi-dev socat gcc \
+    && pip3 install --no-cache-dir --upgrade pip \
+    && pip3 install --no-cache-dir ruamel.yaml.clib \
```
RoiGlinik, line 7: this is can be deleted as well. and so is the GCC in install and purge line

RoiGlinik (commented):
Just need to delete some extra old package and its good

RoiGlinik:
ruamel.yaml.clib lib released a wheel a few weeks ago so there is no need for gcc and the rumal.aml.lib line can be deleted

Dockerfile
pavangudiwada: Sure

PR #567, opened 2022-10-09
start testing robusta installation on python 3.10

aantn:


RoiGlinik:
@aantn  , oops I missed that .

Issue #551, opened 2022-09-18
Restart job on OOM kill and give it higher memory limits (and request)


Sheeproid:
**Is your feature request related to a problem?**
When a job fails because of OOM error, sometimes the solution is to restart it with higher memory limits/requests.

**Describe the solution you'd like**
Create an action in robusta that restarts a job and increasing it's memory requests/limits.

Some considerations:

- To prevent an infinite loop of job's pod crashing and getting more memory, some solution is needed. For example, make a parameter for the actions for the "max" CPU/Memory allowed for the job restarting actions.
- There should be some way to specify the jobs the filter needs to work on by name. Make using a regex is best. However, maybe it can be done via the tigger itself.
- A finding should be created notifying about the job's requests/limits increase

**Additional context**
- Job failure trigger to be used in conjunction with this actions - [here](https://docs.robusta.dev/master/catalog/triggers/smart.html#job-failure)

wahajXgrid:
I just wanted to know, Will the Max parameter takes max limit memory or max request memory?

arikalon1:
I think it should be `request`. Max mem/cpu `request`
For cpu, we recommend using only `request`, and for memory we recommend `request`=`limit`
So I think `request` will be the right parameter

PR #569, opened 2022-10-11
Job restart on oomkilled 

wahajXgrid:
### Issue

[Restart job on OOM kill and give it higher memory limits/requests ](https://github.com/robusta-dev/robusta/issues/551)

### Description

Kubernetes job can be failed due to oomkilled and sometimes the solution is to increase its container resources. This action will trigger when oomkilled occurs and restart the job with the given memory increment.
To prevent an infinite loop of the job's pod crashing and getting more memory this action takes the "Max" resource parameter from the user.

### Checklist
- [x] Manually tested
- [x] Made sure the title of the PR is a good description that can go into the release notes (if applicable)
- [x] Every function, interface, and class has a comment describing what it does and the input/output parameters

### DIFF
--- a/docs/developer-guide/platform/dev-setup.rst
+++ b/docs/developer-guide/platform/dev-setup.rst
@@ -13,7 +13,7 @@ Installing Robusta in-cluster from source
 1. ``git clone`` the source code.
 2. Install `skaffold <https://skaffold.dev/>`_ and `helm <https://helm.sh/>`_
 3. Run ``robusta gen-config`` and copy the result to ``deployment/generated_values.yaml``
-4. Run ``skaffold run --tail``
+4. Run ``skaffold run --tail``, or ``skaffold run --tail -p apple-m1-dev`` if you're using Apple M1
 
 Common errors
 ---------------


playbooks/robusta_playbooks/job_restart_on_oomkilled.py
```diff
@@ -0,0 +1,196 @@
+from robusta.api import *
```
Sheeproid, line 1: @Avi-Robusta already worked on getting oom killed container, maybe you can reuse his code.
here is a snippet.
```
find_most_recent_oom_killed_container(pod, container_statuses=container_statuses, only_current_state=True)
PodContainer.get_requests(container)
PodContainer.get_limits(container)
```

Sheeproid (commented):
Thanks for contributing. See comments :)

playbooks/robusta_playbooks/job_restart_on_oomkilled.py
```diff
@@ -0,0 +1,196 @@
...
+            event.get_job()
+            .spec.template.spec.containers[index]
+            .resources.requests["memory"]
+        )
+        finding.title = f" MAX REACHED "
+
+        finding.add_enrichment(
+            [
+                MarkdownBlock(
+                    f"*You have reached the memory limit*\n```\n{job_temp}\n```"
```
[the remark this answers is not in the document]
wahajXgrid: job_temp variable contains only request memory, which I have defined at line 75.

playbooks/robusta_playbooks/job_restart_on_oomkilled.py
```diff
@@ -0,0 +1,196 @@
...
+        if status_flag:
+            job_spec = restart_job(job_event, params.increase_to)
+
+            job_temp = job_spec.spec.template.spec.containers[index].resources.requests[
+                "memory"
+            ]
+            finding.add_enrichment(
+                [
+                    MarkdownBlock(
+                        f"*Job Restarted With Memory Increament*\n```\n{job_temp}\n```"
```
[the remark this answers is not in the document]
wahajXgrid: Resolved.

playbooks/robusta_playbooks/job_restart_on_oomkilled.py
```diff
@@ -0,0 +1,196 @@
...
+        split_lim = split_req
+
+    return ResourceRequirements(
+        limits={"memory": (str(split_lim) + lim_unit)},
+        requests={"memory": (str(split_req) + req_unit)},
+    )
+
+
+# Function to get job's Pod
+def get_job_pod(namespace, job):
```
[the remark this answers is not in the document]
wahajXgrid: Actually, the reason why I am searching for the oomkilled container in a pod is to increase the memory of the right container,
but in the case of get_job_pod, I am finding the pod related to this job (there can be other pods as well) and it also returns the right pod because the trigger calls only when the job's pod failed.

"get_job_pod" has been changed to “get_job_latest_pod” from job_utils.

playbooks/robusta_playbooks/job_restart_on_oomkilled.py
```diff
@@ -0,0 +1,196 @@
...
+            status_flag = True
+            break
+
+    # Extracting request['memory'] from the containers and comparing with max_resource
+    max_res, mem = split_num_and_str(
+        job_event.spec.template.spec.containers[index].resources.requests["memory"]
+    )
+    if float(max_res) < params.max_resource:
+
+        if status_flag:
```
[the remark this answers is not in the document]
wahajXgrid: Resolved.

playbooks/robusta_playbooks/job_restart_on_oomkilled.py
```diff
@@ -0,0 +1,196 @@
...
+
+
+class IncreaseResources(ActionParams):
+
+    """
+    :var increase_to: (optional).Users will specify how much they want to increase in each restart.
+    :var max_resource: This variable prevent an infinite loop of job's pod crashing and getting more memory.The action won't increase the memory again when the "Max" limit reached.
+    """
+
+    increase_to: Optional[float] = 1
```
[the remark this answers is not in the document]
wahajXgrid: resolved.

playbooks/robusta_playbooks/job_restart_on_oomkilled.py
```diff
@@ -0,0 +1,196 @@
+from robusta.api import *
+
+
+class IncreaseResources(ActionParams):
```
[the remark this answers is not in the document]
wahajXgrid: I am re-using the same unit, just to keep it dynamic.
assuming that the user knows the unit of the memory and how much it to increase accordingly.

playbooks/robusta_playbooks/job_restart_on_oomkilled.py
```diff
@@ -0,0 +1,196 @@
...
+    container_list = get_container_list(
+        job_event.spec.template.spec.containers, increase_to=increase_to
+    )
+    job_spec = RobustaJob(
+        metadata=ObjectMeta(
+            name=job_event.metadata.name,
+            namespace=job_event.metadata.namespace,
+            labels=job_event.metadata.labels,
+        ),
+        spec=JobSpec(
```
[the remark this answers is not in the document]
wahajXgrid: I didn't find any, I was thinking to create an issue for that.

playbooks/robusta_playbooks/job_restart_on_oomkilled.py
```diff
@@ -50,49 +52,57 @@ def job_restart_on_oomkilled(event: JobEvent, params: IncreaseResources):
...
+            job_spec = restart_job(job_event, params.increase_by, params.max_resource)
 
             job_temp = job_spec.spec.template.spec.containers[index].resources.requests[
                 "memory"
             ]
             finding.add_enrichment(
                 [
                     MarkdownBlock(
-                        f"*Job Restarted With Memory Increament*\n```\n{job_temp}\n```"
+                        f"*Job Restarted With Memory Increment*\n```\n{job_temp}\n```"
```
Sheeproid, line 70: can we say which container OOM crashed and specify the increased? e.g "Job named X crashed because of OOM, increasing container <CONTAINER NAME> memory from X<UNIT> to Y<UNIT>"

playbooks/robusta_playbooks/job_restart_on_oomkilled.py
```diff
@@ -27,17 +27,19 @@ def job_restart_on_oomkilled(event: JobEvent, params: IncreaseResources):
...
-
 
     """
     Retrieves job's pod information
     """
-    pod = get_job_pod(job_event.metadata.namespace, job_event.metadata.name)
+    try:
+        pod = get_job_latest_pod(job_event)
+    except:
+        logging.error(
```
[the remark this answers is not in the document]
wahajXgrid: resolved

wahajXgrid:
Screenshot from 2022-11-08 14-33-18

Screenshot from 2022-11-08 14-30-57

playbooks/robusta_playbooks/job_restart_on_oomkilled.py
```diff
@@ -9,18 +30,18 @@ class IncreaseResources(ActionParams):
     :var max_resource: This variable prevent an infinite loop of job's pod crashing and getting more memory.The action won't increase the memory again when the "Max" limit reached.
     """
 
-    increase_by: Optional[float] = 1
+    increase_by: Optional[str] = 1
```
Sheeproid, line 33: is "1" still a relevant value? Should it be null if optional?
Also, can you give an example in the increase_by documentation? "Users will specify how much they want to increase in each restart, e.g 500Mib"

playbooks/robusta_playbooks/job_restart_on_oomkilled.py
```diff
@@ -1,5 +1,26 @@
...
+    MarkdownBlock,
+    ResourceRequirements,
+    RobustaJob,
+    ObjectMeta,
+    JobSpec,
+    PodSpec,
+    PodTemplateSpec,
+)
 from ...src.robusta.core.playbooks.job_utils import get_job_latest_pod
+from bitmath import KiB, GiB, MiB
```
Sheeproid, line 21: I think bitmath module does not come out of the box. It should be added to the pyproject.toml dependency file with a specific version (e.g bitmath="^1.3.3")

playbooks/robusta_playbooks/job_restart_on_oomkilled.py
```diff
@@ -36,79 +57,308 @@ def job_restart_on_oomkilled(event: JobEvent, params: IncreaseResources):
...
+        # splitting num and str
+        existing_limit_memory, existing_lim_unit = split_num_and_str(limits)
+        existing_req_memory, existing_req_unit = split_num_and_str(reqests)
+
+        split_increased_memory, split_increased_memory_unit = split_num_and_str(
+            increase_by
+        )
+
+        # Checking if provided unit is same as job's memory unit
+        if existing_req_unit == split_increased_memory_unit:
```
Sheeproid, line 198: I see you used bitmath, however the code repeats itself several times.
I saw bitmath have a "Rich Comparison" feature, allowing you to compare between Mib and Gib, for example.

I think the code can be much shorter if you try:
1. Getting the limit/request string values from the container.
2. Use bitmath.parse_string to convert the string to bitmath.
4. Use bitmath.parse_string to convert max_resource & increase_by value to bitmath.
5. Compare between the two.
6. If applicable add the request/limit to the increase_by using bitmath
7. Create ResourceRequirements using str()

Sheeproid (commented):
nice usage of bitmath, I think we can make it save us code

playbooks/robusta_playbooks/job_restart_on_oomkilled.py
wahajXgrid: added.

playbooks/robusta_playbooks/job_restart_on_oomkilled.py
wahajXgrid: changed to 0.

playbooks/robusta_playbooks/job_restart_on_oomkilled.py
```diff
@@ -82,21 +79,26 @@ def job_restart_on_oomkilled(event: JobEvent, params: IncreaseResources):
...
                     increase_resource(
                         container,
-                        params.max_resource,
+                        max_resource,
                         params.increase_by,
                         keep_the_same,
                     )
                 )
             else:
+                print("call")
```
Sheeproid, line 101: please remove print

playbooks/robusta_playbooks/job_restart_on_oomkilled.py
```diff
@@ -0,0 +1,247 @@
...
+
+
+class IncreaseResources(ActionParams):
+
+    """
+    :var increase_by: (optional).Users will specify how much they want to increase in each restart, (e.g: 500MiB).
+    :var max_resource: This variable prevent an infinite loop of job's pod crashing and getting more memory.The action won't increase the memory again when the "Max" limit reached, (eg: 1GiB).
+    """
+
+    increase_by: Optional[str] = 0
```
Sheeproid, line 30: can we set it to 500Mi by default? Otherwise if it is 0 it won't do anything, right? it may also get into an infinite loop

playbooks/robusta_playbooks/job_restart_on_oomkilled.py
```diff
@@ -0,0 +1,247 @@
...
+        startupProbe=container.startupProbe,
+        envFrom=container.envFrom,
+        imagePullPolicy=container.imagePullPolicy,
+        resources=memory_increment(
+            container.resources,
+            increase_by,
+            max_resource,
+            keep_the_same,
+        )
+        if (container.resources.limits and container.resources.requests)
```
Sheeproid, line 176: if the container does not have limits and requests, will this cause an infinite restarts for the job?

playbooks/robusta_playbooks/job_restart_on_oomkilled.py
```diff
@@ -0,0 +1,247 @@
...
+        formatted_limit = limits.format("{value:.0f}{unit}").rstrip("B")
+
+        return ResourceRequirements(
+            limits={"memory": (str(formatted_limit))},
+            requests={"memory": (str(formatted_request))},
+        )
+
+
+# Function to split number and string from memory[string]
+def split_num_and_str(num_str: str):
```
Sheeproid, line 204: this function is unused and can be removed

playbooks/robusta_playbooks/job_restart_on_oomkilled.py
```diff
@@ -0,0 +1,247 @@
...
+        else:
+            running_containers.append(status.name)
+
+    """
+    Updating pod's containers resources if required
+    """
+    for index, container in enumerate(pod.spec.containers):
+        if container.name in oomkilled_containers:
+
+            req_memory = bitmath.parse_string_unsafe(
```
Sheeproid, line 83: if an oomkilled_container does not have requests or limit, I think we should fail the actions immediately and not restart the job

playbooks/robusta_playbooks/job_restart_on_oomkilled.py
wahajXgrid: The purpose of this condition is to check if the container has resource limits and requests defined. The script is trying to increase the requested memory of the container, and it needs to make sure that the container has memory limits and requests defined before trying to increase it. If the container does not have memory limits and requests defined, the script will not be able to increase the requested memory and it will return None.

playbooks/robusta_playbooks/job_restart_on_oomkilled.py
wahajXgrid: done.

playbooks/robusta_playbooks/job_restart_on_oomkilled.py
wahajXgrid: added.

playbooks/robusta_playbooks/job_restart_on_oomkilled.py
wahajXgrid: resolved.

playbooks/robusta_playbooks/job_restart_on_oomkilled.py
wahajXgrid: removed.

PR #572, opened 2022-10-14
first structure

AlexPrm93:


aantn:
@AlexPrm93 `_build` is supposed to be .gitignored, but it's contents are being added for some reason. Can you fix that?

PR #580, opened 2022-10-21
Add sinks tutorials

anFatum:


aantn (approved):
Looks good to me.

@pavangudiwada make sure we add this to the new docs as well. Maybe also worth moving the videos to YouTube and having them on our account there.

PR #585, opened 2022-11-07
Hard delete jobs

Avi-Robusta:
this version of supabase_py has a bug in delete.
```
# in package supabase_py/lib/query_builder.py for the 
def _execute_monkey_patch(self) -> Dict[str, Any]:
    """Temporary method to enable syncronous client code."""
    method: str = self.http_method.lower()
    additional_kwargs: Dict[str, Any] = {}
    if method == "get":
        func = requests.get
    elif method == "post":
        func = requests.post
        # Additionally requires the json body (e.g on insert, self.json==row).
        additional_kwargs = {"json": self.json}
    elif method == "put":
        func = requests.put
    elif method == "patch":
        func = requests.patch
    elif method == "delete":
        func = requests.delete
    else:
        raise NotImplementedError(f"Method '{method}' not recognised.")
    url: str = str(self.session.base_url).rstrip("/")
    query: str = str(self.session.params)
    response = func(f"{url}?{query}", headers=self.session.headers, **additional_kwargs)
    return {
        "data": response.json(),
        "status_code": response.status_code,
    }

QueryRequestBuilder.execute = _execute_monkey_patch
```
response.json() throws an exception on status code 204(no content) since there is no data to be decoded
delete success returns 204
We have a similar solution in relay

### DIFF
--- a/src/robusta/core/sinks/robusta/dal/supabase_dal.py
+++ b/src/robusta/core/sinks/robusta/dal/supabase_dal.py
@@ -305,7 +305,8 @@ class SupabaseDal:
                 .eq("cluster_id", self.cluster)
                 .eq("service_key", job.get_service_key())))
         status_code = res.get("status_code")
-        if status_code != 204:
+        valid_deleted_statuses = [204, 200, 202]
+        if status_code not in valid_deleted_statuses:
             logging.error(
                 f"Failed to delete job {job} error: {res.get('data')}"
             )
--- a/src/robusta/core/sinks/robusta/dal/supabase_dal.py
+++ b/src/robusta/core/sinks/robusta/dal/supabase_dal.py
@@ -4,6 +4,8 @@ import threading
 import time
 from typing import List, Dict, Any
 from supabase_py.lib.auth_client import SupabaseAuthClient
+from postgrest_py.request_builder import QueryRequestBuilder
+import requests
 
 from .model_conversion import ModelConversion
 from ....model.cluster_status import ClusterStatus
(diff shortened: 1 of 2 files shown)


src/robusta/core/sinks/robusta/dal/supabase_dal.py
```diff
@@ -293,6 +295,45 @@ def publish_jobs(self, jobs: List[JobInfo]):
...
+        if not job:
+            return
+
+        res = (
+            self.__delete_patch(self.client.table(JOBS_TABLE).delete()
+                .eq("account_id", self.account_id)
+                .eq("cluster_id", self.cluster)
+                .eq("service_key", job.get_service_key())))
+        status_code = res.get("status_code")
+        if status_code != 204:
```
arikalon1, line 308: I would not limit it to 204. 
I think 200 or 202 is also a valid response for a DELETE operation.
I understand it's what currently returned, but supporting also 200 might be safer.
See here: https://restfulapi.net/http-methods/#:~:text=DELETE%20API%20Response,was%20already%20removed.

src/robusta/core/sinks/robusta/robusta_sink.py
```diff
@@ -289,6 +289,14 @@ def __publish_new_nodes(
 
         self.dal.publish_nodes(updated_nodes)
 
+    def __safe_delete_job(self, job_key):
+        try:
+            self.__jobs_cache[job_key].deleted = True
```
arikalon1, line 294: I think we don't need this any more.
We changed the deleted field to True, because we were doing soft delete. 
So we were updating the DB with deleted=true
But now we're using hard delete

arikalon1 (approved):
nice work!
Left few minor comments.

src/robusta/core/sinks/robusta/robusta_sink.py
Avi-Robusta: todo: add comment why this is important

PR #592, opened 2022-11-10
supporting k memory bytes

Avi-Robusta:
We have seen this error allot in user clusters due to k not being supported
```
[31m2022-11-10 14:23:33.324 ERROR    error parsing memory 131072k
  File "/usr/local/lib/python3.9/site-packages/robusta/core/model/pods.py", line 112, in get_number_of_bytes_from_kubernetes_mem_spec
    return int(float(mem_spec))
```


arikalon1 (approved):
yeah, I saw it as well
LGTM

PR #593, opened 2022-11-11
Updated upgrade instructions

pavangudiwada:
Updated **upgrade** instructions with `--set clusterName=<YOUR_CLUSTER_NAME>` since "clusterName" is removed from generated_values.yaml.


arikalon1 (approved):
Looks good

PR #596, opened 2022-11-14
fix yaml manifests so robusta can be deploy to secure clusters

teknologista:
Right now robusta fails to deploy to secured clusters that use a PodSecurity Policy (CIS 1.5 and 1.6) blocking pods from running as root.
This PR adds the relevant securityContext to allow the workloads to be deployed on such clusters.

2 examples of those "security as a default" clusters are Rancher RKE2 (aka Rancher Government) and Elastisys Compliant Kubernetes:

https://docs.rke2.io
https://elastisys.io/compliantkubernetes/


arikalon1 (approved):
Thanks for your contribution @teknologista 
I think it's a lot better with the hardened configuration

PR #597, opened 2022-11-14
update sinks on global config change

Avi-Robusta:


arikalon1 (approved):
Looks good.
The websocket receiver also uses the cluster name and account id.
If we're already fixing it, might be also good to create a new receiver if one of those change.
Checkout ConfigLoader.__reload_receiver(..)

Avi-Robusta:
> Looks good. The websocket receiver also uses the cluster name and account id. If we're already fixing it, might be also good to create a new receiver if one of those change. Checkout ConfigLoader.__reload_receiver(..)

Ill do that in a new PR

PR #600, opened 2022-11-17
Jobs bugs

arikalon1:


src/robusta/core/discovery/discovery.py
```diff
@@ -132,11 +132,20 @@ def discovery_process() -> DiscoveryResults:
...
+                    job_labels = job.spec.selector.match_labels
+                elif job.metadata.labels:
+                    job_name = job.metadata.labels.get("job-name", None)
+                    if job_name:
+                        job_labels = {"job-name": job_name}
+
+                if job_labels:  # add job pods only if we found a valid selector
+                    job_pods = [
+                        pod.metadata.name for pod in pod_items
+                        if job_labels.items() <= (pod.metadata.labels or {}).items()
```
Avi-Robusta, line 147: was going to ask about jobs with the same names in different namespaces but I saw that there is a label `controller-uid` that the pod + job both share with a unique identifier by default, so no extra namespace filtering should be needed

src/robusta/core/discovery/discovery.py
```diff
@@ -132,11 +132,20 @@ def discovery_process() -> DiscoveryResults:
...
+                job_pods = []
+                job_labels = {}
+                if job.spec.selector:
+                    job_labels = job.spec.selector.match_labels
+                elif job.metadata.labels:
+                    job_name = job.metadata.labels.get("job-name", None)
+                    if job_name:
+                        job_labels = {"job-name": job_name}
+
+                if job_labels:  # add job pods only if we found a valid selector
```
Avi-Robusta, line 144: we should only add pods in the same namespace since we can have the case that `job_labels = {"job-name": job_name}` and you can have the same job running in different namespaces

arikalon1: it's always a namespaced query

src/robusta/core/playbooks/job_utils.py
```diff
@@ -5,12 +5,31 @@
...
+def get_job_selector(job: Job) -> Optional[str]:
+    """
+    First try to get by controller_uid, then by selectors, and if both not available by name
+    """
+    job_labels = job.metadata.labels
+    controller_uid = job_labels.get(CONTROLLER_UID, None)
+    if controller_uid:
+        return f"{CONTROLLER_UID}={controller_uid}"
+    elif job.spec.selector and job.spec.selector.matchLabels:
+        return ",".join(f"{k}={v}" for (k,v) in job.spec.selector.matchLabels.items())
```
Avi-Robusta, line 17: you can use build_selector_query(job.spec.selector)

Avi-Robusta (approved):

src/robusta/core/discovery/discovery.py
arikalon1: fixing

src/robusta/core/playbooks/job_utils.py
arikalon1: fixing

PR #611, opened 2022-11-28
Revert "added events to track pageviews of installation (for posthog) (#610)"

Sheeproid:
This reverts commit b95483bf73ca0d9b8e3cf96886938b7fb8284dbf.

it has a bug because it searches for the _static/analytics.js page in a relative location to the specific page


pavangudiwada (approved):
Looks good

PR #613, opened 2022-11-29
add events to track installation events

Sheeproid:
using root path for including /_static/anayltics.js


Sheeproid:
still does not work because of path issues

PR #614, opened 2022-11-29
added events to track pageviews of installation (for posthog)

Sheeproid:
also removed redundant call to analytics.page()


pavangudiwada (approved):
Works as expected

PR #620, opened 2022-12-06
Feature/prometheus button

anFatum:
Instead of showing an oops message when we can't find Prometheus, show a message that Prometheus couldn't be found.

I'm not sure though if the message is okay?


arikalon1:
@anFatum We can delete this PR, right?

PR #621, opened 2022-12-06
Feature/delete svgs

anFatum:
I skip the SVGs everywhere we have PNGs available. Also, I left the option to keep sending SVGs if that is by chance required by the user

### DIFF
--- a/src/robusta/core/sinks/discord/discord_sink.py
+++ b/src/robusta/core/sinks/discord/discord_sink.py
@@ -12,10 +12,11 @@ class DiscordSink(SinkBase):
         self.sender = DiscordSender(
             self.url,
             self.account_id,
-            self.cluster_name
+            self.cluster_name,
+            self.params
         )
 
     def write_finding(self, finding: Finding, platform_enabled: bool):
         self.sender.send_finding_to_discord(
-            finding, self.sink_name, platform_enabled
+            finding,  platform_enabled
         )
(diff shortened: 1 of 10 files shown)


RoiGlinik:
@anFatum I have some thoughts about this implementation.


It seems we want to move to a default of PNG instead of the current SVG

currently the function add_pngs_for_all_svgs seems to do the duplication of the images. 

1. I suggest we add to the alert_graph_enricher svg/png argument with default output to PNG instead of SVG

If anyone wants SVG, they can change that in the gen_values.yaml file with an argument to the enricher.

Currently the UI does not use SVG capabilities anyway so PNG will do. 

2. We might wanna get a higher res for the image instead of 1280/720 (optional)

3. disable the add_pngs_for_all_svgs   calls.  (it's not relevant)

4. we could potentially make the add_pngs_for_all_svgs behavior an enricher as well (just an idea but not necessary)


Then changes won't be necessary in every sink.

The current implementation seems a bit too rigid for blocking SVG files completely.

RoiGlinik:
@anFatum I am not sure if some users want this use case, They can probably do something like that
  - actions:
      - node_disk_analyzer: {}
      - alert_graph_enricher:
          item_type: Node
          resource_type: Disk
          type: png
      - alert_graph_enricher:
          item_type: Node
          resource_type: Disk
          type: svg
    triggers:
      - on_prometheus_alert:
          alert_name: NodeFilesystemSpaceFillingUp
    sinks:
      - robusta_ui_sink

PR #622, opened 2022-12-07
Update prom charts style

RoiGlinik:


aantn:
@RoiGlinik don't forget to attach before/after screenshots that we can use in the release notes.

RoiGlinik:
image

image

arikalon1 (approved):
looks good

PR #623, opened 2022-12-07
Add host param to opsgenie_sink

LeaveMyYard:
Added a new parameter to opsgenie sink: `host`
If it is not provided it will be `None`, meaning the default from opsgenie_sdk will be used.
Otherwise - it will be used in ospgenie configuration.

See https://docs.opsgenie.com/docs/opsgenie-python-sdk-configurations


arikalon1 (approved):
nice work

PR #624, opened 2022-12-08
Encoding RSAKeyPair by default with new public and private values

pavangudiwada:
* Modified the auth-config helm chart with new public and private values.
* Added code to encode RSA keys by default.

### DIFF
--- a/src/robusta/cli/auth.py
+++ b/src/robusta/cli/auth.py
@@ -22,10 +22,10 @@ app = typer.Typer()
 
 
 class RSAKeyPair(BaseModel):
-    prv: str = "none"
-    pub: str = "none"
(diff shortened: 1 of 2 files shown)


src/robusta/cli/auth.py
```diff
@@ -22,8 +22,10 @@
 
 
 class RSAKeyPair(BaseModel):
-    prv: str
-    pub: str
+    prv: str = "none"
```
arikalon1, line 25: I think the default value should be None and not "none"
For example:
prv: str = None

It's like python null, as opposed to string with the value "none"

arikalon1 (approved):
Very nice
Left one minor comment

PR #628, opened 2022-12-09
Feature/prometheus alert

anFatum:
Duplicate of https://github.com/robusta-dev/robusta/pull/620 - changed author email


src/robusta/core/external_apis/prometheus/prometheus_cli.py
```diff
@@ -1,5 +1,6 @@
 from .models import PrometheusQueryResult
 from datetime import datetime
+from robusta.integrations.prometheus.utils import check_prometheus_connection
```
arikalon1, line 3: please use relative import

src/robusta/core/playbooks/playbooks_event_handler_impl.py
```diff
@@ -226,6 +228,17 @@ def __run_playbook_actions(
 
                 try:
                     registered_action.func(execution_event, params)
+                except PrometheusNotFound as e:
+                    logging.error(str(e))
+                    execution_event.add_enrichment(
+                        [
+                            MarkdownBlock(
+                                text=f"Robusta couldn't connect to the Prometheus client, check if the service is "
```
arikalon1, line 236: The link here is about configuring custom alert manager, but I think here we're trying to connect to Prometheus.
right?
If this is the case, I would change to:

Could not connect to the Prometheus. 
Please add to `globalConfig` in `generated_values.yaml` the cluster `prometheus_url`.
For example:
```
globalConfig:
  prometheus_url: http://prometheus-server.monitoring.svc.cluster.local:9090
```

arikalon1 (approved):
looks good
Add minor comments

PR #629, opened 2022-12-14
test new docker cli action

RoiGlinik:


RoiGlinik:
Pushes an image of the docker CLI with the release

PR #636, opened 2022-12-23
Fix: minor changes to fix CLI command

k4kratik:
`playbooks` was missing from the guide


arikalon1 (approved):
Thank you for your contribution @k4kratik

PR #639, opened 2022-12-26
Pyright static type checker

LeaveMyYard:
Added [PyRight](https://github.com/microsoft/pyright) typechecker. Fixed all errors.
Added [PyRight](https://github.com/microsoft/pyright) to pre-commit hooks.


.pre-commit-config.yaml
```diff
@@ -29,9 +29,7 @@ repos:
...
         args: [--settings-path=pyproject.toml]
 
-  # - repo: https://github.com/pre-commit/mirrors-mypy
-  #   rev: v0.790
-  #   hooks:
-  #     - id: mypy
-  #       language: system
-  #       args: [--config-file=mypy.ini]
+  - repo: https://github.com/RobertCraigie/pyright-python
+    rev: v1.1.284
```
Avi-Robusta, line 33: the newest version is 1.1.287, maybe try to run a test and see if it works out of the box for us

LeaveMyYard: I will check for the newest version, but Microsoft is releasing minor releases few times a month so it will be hard to always update it

docs/_ext/autorobusta.py
```diff
@@ -119,7 +120,8 @@ def __document_field_content(cls, field: ModelField, show_code: bool, show_optio
         if show_code:
             content.extend(cls.__document_field_example(field))
 
-        if typing.get_origin(field.type_) == typing.Union:
+        # TODO: Will not work for 3.7
```
Avi-Robusta, line 123: Will this impact users trying to build robusta on python 3.7?

playbooks/robusta_playbooks/java_pod_troubleshooting.py
```diff
@@ -138,6 +140,11 @@ def run_jdk_command_on_pid(
 
 
 def run_java_toolkit_command(jdk_cmd: str, pod: RobustaPod, override_jtk_image: str):
+    assert pod.metadata is not None
+    assert pod.metadata.name is not None
+    assert pod.spec is not None
+    assert pod.spec.nodeName is not None
```
Avi-Robusta, line 146: I see we do the same asserts allot
would a functions like this work? :
```
def pod_assert(pod):
    assert pod.metadata is not None
    assert pod.metadata.name is not None
    assert pod.spec is not None
    assert pod.spec.nodeName is not None
    
def node_assert(node):
```

playbooks/robusta_playbooks/job_actions.py
```diff
@@ -164,10 +172,13 @@ def job_pod_enricher(event: JobEvent, params: JobPodEnricherParams):
     pod = get_job_latest_pod(job)
 
     if not pod:
+        assert job.metadata is not None
```
Avi-Robusta, line 175: what about 
`        
        assert job.metadata.name is not None
        assert job.metadata.namespace is not None
`?

docs/_ext/autorobusta.py
LeaveMyYard: The thing is - it was in the codebase, it is not my change.
PyRight raised an error since I was using python 3.7 as the lowest supported version, so it found that this might not work. I think there is at least one more place in the code that will not work with 3.7, that I told @arikalon1 about. Maybe we should discuss on how to better solve it if we should.

And, just for reminder, I left those comments.

PR #643, opened 2022-12-28
Added comments to redirect users to using RobustaCLI to generate Helm…

pavangudiwada:
Added comments and links to the Installation page

### DIFF
--- a/helm/README.md
+++ b/helm/README.md
@@ -1,7 +1,7 @@
 # Robusta helm chart
 Robusta's helm chart is served from our gcp bucket.
 
-NOTE: **Don't use the Helm values file directly**. Check the [Installation instructions](https://docs.robusta.dev/master/getting-started/installation.html) 
+NOTE: **Don't use the Helm values file directly**. It has missing values that need to be filled in. Check the [Installation instructions](https://docs.robusta.dev/master/getting-started/installation.html)
 
 # Upload a new chart version
 If you want to upload a new chart version, follow these steps:
--- a/helm/README.md
+++ b/helm/README.md
@@ -1,7 +1,7 @@
 # Robusta helm chart
 Robusta's helm chart is served from our gcp bucket.
 
-NOTE: **Don't use the Helm values file directly**. Check the [Installation instructions](https://docs.robusta.dev/master/getting-started/installation.html) to generate one, using the RobustaCLI. 
+NOTE: **Don't use the Helm values file directly**. Check the [Installation instructions](https://docs.robusta.dev/master/getting-started/installation.html) 
 
 # Upload a new chart version
 If you want to upload a new chart version, follow these steps:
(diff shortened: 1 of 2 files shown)


helm/README.md
```diff
@@ -9,4 +11,4 @@ If you want to upload a new chart version, follow these steps:
 4. From the directory, `helm`, run: `./upload_chart.sh`
 
 # Installing robusta with the helm chart
-See https://docs.robusta.dev/master/getting-started/installation.html
+See https://docs.robusta.dev/master/getting-started/installation.html.
```
aantn, line 14: Get rid of the period. It can mess up the link.

helm/README.md
```diff
@@ -1,5 +1,7 @@
 # Robusta helm chart
-Robusta's helm chart is served from our gcp bucket
+Robusta's helm chart is served from our gcp bucket.
+
+NOTE: **Don't use the Helm values file directly**. Check the [Installation instructions](https://docs.robusta.dev/master/getting-started/installation.html) to generate one, using the RobustaCLI.
```
aantn, line 4: `robusta cli` not `RobustaCLI`

helm/robusta/values.yaml
```diff
@@ -1,6 +1,9 @@
 # most of this file is documented at https://docs.robusta.dev/master/user-guide/configuration.html
 # if you have any questions, feel free to ask via github issues or Slack (see link on robusta.dev)
 
+# NOTE: Do not use this file to install Robusta. Check the Installation isntructions to generate your generated_values.yaml(Helm values file) using RobustaCLI.
```
aantn, line 4: >  NOTE: Do not use this file to install Robusta.
Why not?

For the second half of the sentence, I would just say "see the installation instructions instead." No need to mention the generated_values or Robusta cli here IMO. They can read about it in the installation page.

Also, typo in `isntructions`.

pavangudiwada: Fixed

helm/README.md
pavangudiwada: Fixed

helm/README.md
pavangudiwada: Removed.

helm/robusta/values.yaml
```diff
@@ -1,6 +1,8 @@
 # most of this file is documented at https://docs.robusta.dev/master/user-guide/configuration.html
 # if you have any questions, feel free to ask via github issues or Slack (see link on robusta.dev)
 
+# NOTE: Do not use this file to install Robusta. See the installation instructions instead https://docs.robusta.dev/master/installation.html
```
aantn, line 4: I would still explain why you shouldn't use this file directly. E.g. 

`Do not use this file as-is when installing Robusta. It has missing values that need to be filled in. See the installation...`

aantn (approved):

helm/robusta/values.yaml
pavangudiwada: Thank you, fixed it.

PR #656, opened 2023-01-03
Improve `get_resolution_from_duration` calculation

LeaveMyYard:
Make resolution for prometheus requests to depend on duration.
Make it easy to configure that function.


src/robusta/core/playbooks/prometheus_enrichment_utils.py
```diff
@@ -54,6 +55,22 @@ def run_prometheus_query(
     )
 
 
+_DEFAULT_RESOLUTION = 500
+_RESOLUTION_DATA: Dict[timedelta, Union[int, Callable[[timedelta], int]]] = {
```
arikalon1, line 59: I think 250 points should be the minimum. (for 1 hour)
For a day, it seems a little low. 200/24 = ~once every 7 min. I think it's too low.
Same for week, 300 ~ twice an hour. 

I think for between 1h and 1 day I would choose once a minute (max of 1440)
And between 1d and 1w once every 5m (max of 2016)
Above that I would just choose 3000

src/robusta/core/playbooks/prometheus_enrichment_utils.py
```diff
@@ -54,6 +55,22 @@ def run_prometheus_query(
...
+    timedelta(hours=1): 150,
+    timedelta(days=1): 200,
+    timedelta(weeks=1): 300,
+    # timedelta(days=30): lambda duration: int(duration.total_seconds() / 60),
```
arikalon1, line 63: I would remove if not used

arikalon1 (approved):
Looks good.
Left minor comments

PR #657, opened 2023-01-03
Adding job support for action Related_Pods

Avi-Robusta:


playbooks/robusta_playbooks/k8s_resource_enrichments.py
```diff
@@ -38,7 +38,9 @@ def related_pods(event: KubernetesResourceEvent):
     if resource.kind not in supported_resources:
         raise ActionException(ErrorCodes.RESOURCE_NOT_SUPPORTED, f"Related pods is not supported for resource {resource.kind}")
 
-    if resource.kind == "Pod":
+    if resource.kind == "Job":
+        pods = get_job_all_pods(resource)
```
arikalon1, line 42: Did you check the case you couldn't find any pods?
This function will return None, and I think it breaks on line 51

Need to figure out what's the correct response format the UI can handle
Is it an empty table? An ActionException? Something else?

arikalon1 (approved):
looks good
left one comment

playbooks/robusta_playbooks/k8s_resource_enrichments.py
Avi-Robusta: checked it, UI is good with an empty table

PR #662, opened 2023-01-09
add support for am versions <= 0.21.

RoiGlinik:
In older versions, isEqual was not used yet.
negative operators did not exist as well, only equal ones will work.


arikalon1 (approved):
LGTM

PR #664, opened 2023-01-10
Fix missing label on container memory graph

LeaveMyYard:
Prometheus metrics are not guaranteed (in `src/robusta/core/external_apis/prometheus/models.py`):
`The Metric object is defined in prometheus as a dictionary so no specific labels are guaranteed`

So, to fix an issue, added `chart_labels_override` parameter to `create_chart_from_prometheus_query` function to override specific labels on the chart.

All overrides data for future use is stored at `src/robusta/core/playbooks/prometheus_enrichment_utils.py:242`


src/robusta/core/playbooks/prometheus_enrichment_utils.py
```diff
@@ -220,6 +228,14 @@ def create_resource_enrichment(
         if title_override
         else f"{resource_type.name} {values_format_text} for this {item_type.name.lower()}"
     )
+
+    # NOTE: Some queries do not prodice automatic labels, so we need to override them
```
arikalon1, line 232: Small typo in `prodice`

src/robusta/core/playbooks/prometheus_enrichment_utils.py
```diff
@@ -220,6 +228,14 @@ def create_resource_enrichment(
...
+    # Numerical index is the number of the series in the chart to override (excluding lines)
+    chart_labels_overrides = {
+        (ResourceChartResourceType.Memory, ResourceChartItemType.Container): {
+            0: "Memory Usage",
```
arikalon1, line 236: I understand the comment on the index, but it's a little confusing.
What's the scenario where a query returns multiple series?
If we wanted to run the container memory query without filter on the container. 
Will we get series per container?  I'm trying to think if this solution works if the number of the returned series is unknown.
Can there be a different override label for different series on the result, for a single query?

arikalon1 (commented):
Left one comment
Not sure I got it right
LMK what you think

src/robusta/core/playbooks/prometheus_enrichment_utils.py
```diff
@@ -220,6 +231,14 @@ def create_resource_enrichment(
         if title_override
         else f"{resource_type.name} {values_format_text} for this {item_type.name.lower()}"
     )
+
+    # NOTE: Some queries do not produce automatic labels, so we need to provide them
+    # Parameter in lambda is the number of the series in the chart to override (excluding lines)
+    # It could be used if there are multiple series in the chart
+    chart_label_factories: Dict[ResourceKey, ChartLabelFactory] = {
+        (ResourceChartResourceType.Memory, ResourceChartItemType.Container): lambda i: "Memory Usage",
```
arikalon1, line 239: Is it right to add the `container` label here?
For this query you have the container label in the query ($container)
I think the chart title is "Memory Usage"
WDYT?

arikalon1 (approved):
Nice
Left a small comment

PR #668, opened 2023-01-11
Namespaces discovery

LeaveMyYard:
Add discovery of namespaces, save them to new table to a supabase for frontend to use.

Requires this migration to be added:
https://github.com/robusta-dev/robusta-storage/pull/11


LeaveMyYard:
TODO: Is status field even needed? Can namespace be in any other state, rather then `Active` and `Terminating` (that will be before deletion)?

arikalon1 (approved):
Nice work
1 small comment

PR #673, opened 2023-01-15
Trim Datadog event body length to max allowed characters

MacroPower:
Datadog events are limited to 4000 characters. See:

https://docs.datadoghq.com/api/latest/events/

While the API itself automatically truncates these, the Datadog Python client library will error since it strictly enforces the defined schema.


arikalon1 (approved):
Looks good
Thank you for your contribution @MacroPower

PR #677, opened 2023-01-16
removed pods that finished running from discovery

Avi-Robusta:
There have been cases of orphan pods appearing in our service discovery that wernt completed by GC

### DIFF
--- a/src/robusta/core/discovery/discovery.py
+++ b/src/robusta/core/discovery/discovery.py
@@ -246,7 +246,7 @@ def is_pod_finished(pod: V1Pod) -> bool:
     try:
         # all containers in the pod have terminated, this pod should be removed by GC
         return pod.status.phase.lower() in ['succeeded', 'failed']
-    except Exception: # phase is an optional field
+    except AttributeError:  # phase is an optional field
         return False
 
 
--- a/src/robusta/core/discovery/discovery.py
+++ b/src/robusta/core/discovery/discovery.py
@@ -141,10 +141,8 @@ class Discovery:
                         extract_total_pods(pod),
                         extract_ready_pods(pod),
                     )
-                    for pod in pod_items
-                    if not pod.metadata.owner_references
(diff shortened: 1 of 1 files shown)


src/robusta/core/discovery/discovery.py
```diff
@@ -205,6 +205,14 @@ def is_pod_ready(pod: V1Pod) -> bool:
     return False
 
 
+def is_pod_finished(pod: V1Pod) -> bool:
+    try:
+        # all containers in the pod have terminated, this pod should be removed by GC
+        return pod.status.phase.lower() in ['succeeded', 'failed']
+    except Exception: # phase is an optional field
```
LeaveMyYard, line 212: Better change to `except AttributeError`, as exactly `AttributeError` is expected.

LeaveMyYard (approved):

src/robusta/core/discovery/discovery.py
```diff
@@ -205,6 +205,14 @@ def is_pod_ready(pod: V1Pod) -> bool:
     return False
 
 
+def is_pod_finished(pod: V1Pod) -> bool:
+    try:
+        # all containers in the pod have terminated, this pod should be removed by GC
+        return pod.status.phase.lower() in ['succeeded', 'failed']
```
arikalon1, line 211: Are we deliberately including "Unknown" and "Pending" phases?

Avi-Robusta: Yes, (unknown , pending or running pods with no owner references)
It wont be an orphan pod if its Unknown or pending. (Since the GC should remove them)

Also since it is a deliberate pod being run by a user he should be receiving alerts due to the pod pending or Unknown phase so it will be important for us to display that for him on the apps screen and not hide it.

arikalon1: sounds good, thanks for clarifying

PR #678, opened 2023-01-16
Logs enricher improvements

RoiGlinik:
support for new use cases

request from UI with specific container.
multiple containers and init containers.
Read and verify container subject from prometheus alerts.
use same .log ext.


playbooks/robusta_playbooks/alerts_integration.py
```diff
@@ -261,20 +264,48 @@ def logs_enricher(event: PodEvent, params: LogEnricherParams):
...
         RegexReplacementStyle[params.regex_replacement_style] if params.regex_replacement_style else None
     )
-    log_data = pod.get_logs(
-        regex_replacer_patterns=params.regex_replacer_patterns,
-        regex_replacement_style=regex_replacement_style,
-        previous=params.previous,
-    )
-    if not log_data:
-        return
+    for container in containers:
```
arikalon1, line 292: I think this breaks the UI logs
The UI logs uses the `logs_enricher` action
The UI doesn't send container atm, the PodEvent is created without a container, and often nothing is crashing.
On such a scenario, `containers` will be empty, and we'll have no logs.

I think the default behavior, should be like before, which is to get the logs of the first container, or get logs of all containers

If we return more than one file block, we need to verify the UI keeps working as before

playbooks/robusta_playbooks/alerts_integration.py
```diff
@@ -261,20 +264,48 @@ def logs_enricher(event: PodEvent, params: LogEnricherParams):
...
+            for container_status in all_statuses
+            if container_status.name == event.get_subject().container
+        ]
+
+    # If no container is specified, find containers in an unhealthy state. good for pending pods
+    if len(containers) == 0:
+        containers = [
+            container_status.name
+            for container_status in all_statuses
+            if container_status.state.waiting is not None and container_status.restartCount >= 1
```
arikalon1, line 284: I think `logs_enricher` can be called not only for unhealthy pods.
Some users can call it on high cpu or memory alert.
I don't think we should filter out unhealthy containers. 

If `previous` is `true`, we should filter out containers that were not restarted.
WDYT?

arikalon1 (commented):
nice work
left 2 comments

RoiGlinik:
@arikalon1 You are right, i used the old behavior for now when there is no container to support all the old cases.

playbooks/robusta_playbooks/alerts_integration.py
```diff
@@ -261,20 +264,38 @@ def logs_enricher(event: PodEvent, params: LogEnricherParams):
             )
         return
 
+    container: Optional[str] = None
```
arikalon1, line 267: I think it would be more readable to have:
```
container: str = pod.spec.containers[0].name
```
And to remove the last `elif` clause

arikalon1 (approved):
Looks good
1 minor comment

PR #679, opened 2023-01-16
Add automountServiceAccountToken

LeaveMyYard:
Add automountServiceAccountToken to `values.yaml`. Propagate the value to helm templates with Deployments.


arikalon1 (approved):
LGTM

PR #683, opened 2023-01-16
Fix pre-merge checks

aantn:
Right now the `run_tests` actions doesn't run when you open a PR. It only runs when a commit is pushed after opening the PR, or on an approved branch (I think) before opening the PR.


arikalon1 (approved):
LGTM

PR #686, opened 2023-01-17
fix issues with docker cli release

RoiGlinik:
- used to run in parallel to release and before the python cli was uploaded and was 1 release behind.
- missing latest tag for script upgrade.
- skip prereleases because of latest tag.


Avi-Robusta (approved):

RoiGlinik:
I forced Avi to merge. I will make sure the image is ok on next release

PR #701, opened 2023-01-23
added failed time for jobs

Avi-Robusta:
also fixed bugs with non active statuses

### DIFF
--- a/src/robusta/core/model/jobs.py
+++ b/src/robusta/core/model/jobs.py
@@ -1,3 +1,4 @@
+from datetime import datetime
 from typing import Dict, List, Optional
 
 from kubernetes.client import V1Container, V1Job, V1JobSpec, V1JobStatus, V1PodSpec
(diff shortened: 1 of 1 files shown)


src/robusta/core/model/jobs.py
```diff
@@ -39,22 +40,38 @@ class JobStatus(BaseModel):
...
     conditions: List[JobCondition]
 
     @staticmethod
     def from_api_server(job: V1Job) -> "JobStatus":
         job_status: V1JobStatus = job.status
         job_conditions: List[JobCondition] = [
-            JobCondition(type=condition.type, message=condition.message) for condition in (job_status.conditions or [])
+            JobCondition(type=condition.type, message=condition.message)
+            for condition in (job_status.conditions or [])
+            if condition.status.lower() == "true"
```
Avi-Robusta, line 52: we were passing all the conditions even if they wernt active

arikalon1: nice

arikalon1 (approved):
nice work

PR #703, opened 2023-01-25
Add ability to provide Authorization header to Prometheus and alert manager to use with OpenShift

### LINKED ISSUE #616: Add support for authorization header when calling Prometheus
Some Prometheus installations requires authentication.

For example, when using https://github.com/openshift/oauth-proxy to authenticate requests to Prometheus.

We'd like to provide a Bearer token, that can be sent to Prometheus in an Authorization header

LeaveMyYard:
In the scope of this PR added various new settings:
- PROMETHEUS_SSL_ENABLED env variable to configure to use ssl with Prometheus (default False)
- prometheus_auth parameter to configure Authentication header when connecting to Prometheus (default not set)
- alertmanager_auth to configure Authentication header when connecting to Alert manager (default not set)

Additionally, changed the logic of check_prometheus_connection function to fix an issue with OpenShift responding to 
```GET /``` with status code 503 (was expected 200). Now, we will make some query to prometheus.

Closes #616

### CODE REVIEW

File: src/robusta/core/model/env_vars.py [hunk 1]
```diff
@@ -29,6 +29,10 @@
 
 PROMETHEUS_REQUEST_TIMEOUT_SECONDS = float(os.environ.get("PROMETHEUS_REQUEST_TIMEOUT_SECONDS", 90.0))
 PROMETHEUS_ENABLED = os.environ.get("PROMETHEUS_ENABLED", "false").lower() == "true"
+PROMETHEUS_AUTH_HEADER = os.environ.get("PROMETHEUS_AUTH_HEADER")
```
arikalon1, line 32: I think we should have those as a part of `globalConfig`

arikalon1 (approved):
Very nice!

Can you please add something in the docs that explains when and how to use the new parameters? (I think around [here](https://docs.robusta.dev/master/user-guide/alert-manager.html#setting-up-a-custom-prometheus-alertmanager-and-grafana) can be a good place)
Specifically 
PROMETHEUS_SSL_ENABLED
prometheus_auth ( I'm not sure it's clear how to generate the basic auth token, I would also add an example with Bearer token)

In the github issue, I would also leave a link to the docs for how to configure it

bear-redhat:
Thank you for the PR!
It would be great if we can specify the CA certificate. We need to configure the CA certificate by specifying a file, since in some [cases](https://docs.openshift.com/container-platform/4.12/security/certificates/service-serving-certificate.html#add-service-certificate-configmap_service-serving-certificate), the certificate is signed by an internal CA.

LeaveMyYard:
> Thank you for the PR! It would be great if we can specify the CA certificate. We need to configure the CA certificate by specifying a file, since in some [cases](https://docs.openshift.com/container-platform/4.12/security/certificates/service-serving-certificate.html#add-service-certificate-configmap_service-serving-certificate), the certificate is signed by an internal CA.

Hey!
Sure, I will be adding certificates also, as requested in #616

LeaveMyYard:
@bear-redhat additional CA certificate can be added after https://github.com/robusta-dev/robusta/pull/561/ was merged:
1. You need to base64 encode your custom certificates.
2. In your `generated_values.yaml` file under runner add
```yaml
runner:
  certificate: "your-encoded-data" # base64 encoded
```

PR #704, opened 2023-01-29
removed noisy logs

Avi-Robusta:
set them to debug inorder to be able to see them via logging level change if needed


arikalon1 (approved):
thank you!

PR #705, opened 2023-01-29
added a job failure to built in playbooks

Avi-Robusta:
added on job failure trigger to built in playbooks,
It's more immediately fired than prometheus so it can give logs for pods that will be terminated after job failure.

### DIFF
--- a/helm/robusta/values.yaml
+++ b/helm/robusta/values.yaml
@@ -94,6 +94,18 @@ builtinPlaybooks:
   - image_pull_backoff_reporter:
       rate_limit: 3600
 
+- triggers:
+  - on_job_failure: {}
+  actions:
+  - create_finding:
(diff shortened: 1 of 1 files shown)


helm/robusta/values.yaml
```diff
@@ -94,6 +94,18 @@ builtinPlaybooks:
   - image_pull_backoff_reporter:
       rate_limit: 3600
 
+- triggers:
+  - on_job_failure: {}
+  actions:
+  - create_finding:
```
Avi-Robusta, line 100: None of these events create a finding, they only add enrichments so without this we get a "generic changed" finding without any additional information like service_key, subject name, kind, namespace ...

arikalon1 (approved):
Looks good
Add small comment

PR #709, opened 2023-01-30
Pyright static type checker

LeaveMyYard:


playbooks/robusta_playbooks/deployment_status_report.py
```diff
@@ -76,6 +76,9 @@ def report_rendering_task(event: ExecutionBaseEvent, action_params: ReportParams
 
 
 def has_matching_diff(event: DeploymentChangeEvent, fields_to_monitor: List[str]) -> bool:
+    if event.obj is None or event.old_obj is None:
```
LeaveMyYard, line 79: `event.obj` or `event.old_obj` might be None, added additional check for that

PR #710, opened 2023-01-31
Add customPlaybooks to autogenerated yaml

aantn:
Instead of generating example yaml like:

```yaml
triggers:
  - foo
actions:
  - bar
```

Now we generate yaml like:

```yaml
customPlaybooks:
- triggers: 
   - foo
   actions:
   - bar
```

This change is due to real world confusion I saw :)


arikalon1 (approved):
LGTM

PR #717, opened 2023-02-05
add ttl option to ui sink and cluster status

RoiGlinik:
default for 6 months.

There is an option to use Interval type from Postgres but not sure its necessary.

### DIFF
--- a/src/robusta/core/model/cluster_status.py
+++ b/src/robusta/core/model/cluster_status.py
@@ -7,3 +7,4 @@ class ClusterStatus(BaseModel):
     version: str
     last_alert_at: Optional[str]  # ts
     light_actions: List[str]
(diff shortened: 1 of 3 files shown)


src/robusta/core/model/cluster_status.py
```diff
@@ -7,3 +7,4 @@ class ClusterStatus(BaseModel):
     version: str
     last_alert_at: Optional[str]  # ts
     light_actions: List[str]
+    ttl_hours: int = 4380  # Time before unactive cluster data is deleted. 6 Months default.
```
arikalon1, line 10: Don't think we want the default value here

arikalon1 (approved):
LGTM
Please just add a short description to the sink docs

PR #720, opened 2023-02-06
Update webhook.rst

Aref-Riant:
shouldn't it be sinks?


arikalon1 (approved):
Yes, you're right
Thanks @Aref-Riant

PR #723, opened 2023-02-06
Fix cpu graphs by adding label + correct units

LeaveMyYard:
- Added new ChartValuesFormat for CPU Usage
- Added chart label factories for CPU


arikalon1 (approved):
LGTM

PR #725, opened 2023-02-09
Feature: Playbook base64 encoded values

pavangudiwada:
Users can use base64 encoded keys with **external playbooks repos** and **git change audit** action.

### DIFF
--- a/playbooks/robusta_playbooks/git_change_audit.py
+++ b/playbooks/robusta_playbooks/git_change_audit.py
@@ -12,7 +12,10 @@ from robusta.api import (
     KubernetesAnyChangeEvent,
     action,
     is_matching_diff,
+    is_base64_encoded
 )
+import base64
+import logging
 
 
 class GitAuditParams(ActionParams):
@@ -24,6 +27,14 @@ class GitAuditParams(ActionParams):
 
     :example git_url: "git@github.com:arikalon1/robusta-audit.git"
     """
+
+    def post_initialization(self):
+        secret_key = self.git_key.get_secret_value()
+
+        if is_base64_encoded(secret_key): 
+            self.git_key = SecretStr(base64.b64decode(secret_key).decode("utf-8"))
+        else:
+            self.git_key = SecretStr(secret_key)
 
     cluster_name: str
     git_url: str
@@ -44,6 +55,7 @@ skipped_kinds: List[str] = [
     "ClusterRole",
     "ClusterRoleBinding",
     "ServiceAccount",
+    "ConfigMap"
 ]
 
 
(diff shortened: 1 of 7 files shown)


playbooks/robusta_playbooks/git_change_audit.py
```diff
@@ -64,38 +76,42 @@ def git_change_audit(event: KubernetesAnyChangeEvent, action_params: GitAuditPar
...
+            if obj_diff(event.obj.spec, old_spec, action_params.ignored_changes):  # we have a change in the spec
+                git_repo.commit_push(
+                    hikaru.get_yaml(event.obj.spec),
+                    path,
+                    name,
+                    f"Update {event.obj.kind} named {event.obj.metadata.name} on namespace {namespace}",
+                    action_params.cluster_name,
+                )
+    except Exception as e:
+        logging.exception(e)
```
arikalon1, line 116: Don't think we need that
The exc_info logs the stack trace

src/robusta/utils/base64_utils.py
```diff
@@ -0,0 +1,14 @@
+import base64
+import logging
+
+
+def is_base64_encoded(value: str) -> bool:
+    try:
+        if base64.b64encode(base64.b64decode(value)) == bytes(value, 'utf-8'):  # key is b64
```
arikalon1, line 7: I think you can just do:
`return base64.b64encode(base64.b64decode(value)) == bytes(value, 'utf-8')`  # value is b64

src/robusta/utils/base64_utils.py
```diff
@@ -0,0 +1,14 @@
...
+            return True
+        else:
+            return False
+    except Exception as e:
+        logging.error("Error during base64 check")
```
arikalon1, line 12: I don't think we need to print anything here
It's not a real error
When we get a value which isn't b64, we expect that to happen. It's just a way to test if value is b64 encoded
Printing this as errors will just spam the logs

arikalon1 (approved):
Nice work!

Left some minor comments
Please add the relevant information to the docs, on how to use the b64 encoded values, and also how to create it, from the multi line keys.

src/robusta/utils/base64_utils.py
pavangudiwada: Understood, fixed.

src/robusta/utils/base64_utils.py
pavangudiwada: Fixed

playbooks/robusta_playbooks/git_change_audit.py
pavangudiwada: Got it.

PR #726, opened 2023-02-09
create generic resource event enricher

RoiGlinik:


RoiGlinik:
@arikalon1  I did some major changed. regarding your question.
1. No, it could be, requires some changes in some places. I blocked some of them for now. 
2. Yes it works.
I added support for fetching related pods as well for some of the resources. add object column and merged it in the table for the UI. an example:
image

PR #729, opened 2023-02-12
slack add support for additional cert

RoiGlinik:
When adding custom cert, use the same store in slack ssl context


arikalon1 (approved):
LGTM

PR #730, opened 2023-02-14
Fix config load order bug with LA

Avi-Robusta:


Avi-Robusta:
it was solved by the order change

this `runner_config.light_actions if runner_config.light_actions else [] ` is incase someone messes with the config and removes it all together(or adds a typo there) to make sure there wont be issues with the sinks being configured

arikalon1 (approved):
LGTM

PR #737, opened 2023-02-17
run discovery in batches

arikalon1:
Changed discovery to work in batches for bigger clusters
Not throwing error in case of no rsa keys
Removed forwarder events of ReplicationController (not supported on runner, and not in use)

### DIFF
--- a/src/robusta/core/discovery/discovery.py
+++ b/src/robusta/core/discovery/discovery.py
@@ -23,6 +23,8 @@ from kubernetes.client import (
 )
 from pydantic import BaseModel
 
+from robusta.core.model.cluster_status import ClusterStats
+from robusta.core.model.env_vars import DISCOVERY_MAX_BATCHES, DISCOVERY_BATCH_SIZE
 from robusta.core.discovery import utils
 from robusta.core.model.jobs import JobInfo
 from robusta.core.model.namespaces import NamespaceInfo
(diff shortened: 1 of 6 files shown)


src/robusta/core/discovery/discovery.py
```diff
@@ -70,87 +72,132 @@ def discovery_process() -> DiscoveryResults:
...
-                ]
-            )
         except Exception:
             logging.error(
                 "Failed to run periodic service discovery",
                 exc_info=True,
             )
 
-        # discover nodes
+        # discover nodes - no need for batching. Number of nodes is not big enough
```
Avi-Robusta, line 200: Bug-
nodes uses `pod_items` for getting the pods list so it will need batching too.
the pod_items here will only be the last batch run in the pods discovery so the rest of the pods will be missing in discovery

src/robusta/core/discovery/discovery.py
```diff
@@ -70,87 +72,132 @@ def discovery_process() -> DiscoveryResults:
...
+                if not continue_ref:
+                    break
+
+            # discover pods
+            continue_ref = None
+            for _ in range(DISCOVERY_MAX_BATCHES):
+                pods: V1PodList = client.CoreV1Api().list_pod_for_all_namespaces(
+                    limit=DISCOVERY_BATCH_SIZE, _continue=continue_ref
+                )
+                pod_items = pods.items
```
Avi-Robusta, line 175: add node_requests in this loop since pod_items gets overwritten and we only need a little data in the pod for the node

src/robusta/core/discovery/discovery.py
```diff
@@ -169,28 +216,37 @@ def discovery_process() -> DiscoveryResults:
...
+                        job_labels = job.spec.selector.match_labels
+                    elif job.metadata.labels:
+                        job_name = job.metadata.labels.get("job-name", None)
+                        if job_name:
+                            job_labels = {"job-name": job_name}
+
+                    if job_labels:  # add job pods only if we found a valid selector
+                        job_pods = [
+                            pod.metadata.name
+                            for pod in pod_items
```
Avi-Robusta, line 237: also will not be all pods, pod_items is only last batch of pods

src/robusta/core/discovery/discovery.py
Avi-Robusta: also might want to try to match pod to job here too

src/robusta/core/discovery/discovery.py
Avi-Robusta: might need to move job discovery before pod inorder to detect if a pod is related to a job

src/robusta/core/model/cluster_status.py
```diff
@@ -1,10 +1,21 @@
...
+
+
 class ClusterStatus(BaseModel):
     account_id: str
     cluster_id: str
     version: str
     last_alert_at: Optional[str]  # ts
     light_actions: List[str]
     ttl_hours: int
+    stats: ClusterStats
```
Avi-Robusta, line 21: reminder to add pr to git storage

arikalon1: done

src/robusta/core/discovery/discovery.py
arikalon1: fixed

src/robusta/core/discovery/discovery.py
arikalon1: fixed

src/robusta/core/discovery/discovery.py
arikalon1: fixed

Avi-Robusta (approved):
Looks good, handles errors really well. (I ran some tests) 

I was a bit concerned about storing all pod metadata but I calculated worst case scenario and its like ~20MB so its nothing.

One last thing the cluster stats are in the logs now (because of cluster status) so you will see this every five minutes.

cluster status {'account_id': '****', 'cluster_id': 'TestCluster', 'version': '0.0.0', 'light_actions': 29, 'ttl_hours': 4380, 'stats': {'deployments': 14, 'statefulsets': 2, 'daemonsets': 18, 'replicasets': 32, 'pods': 67, 'nodes': 4, 'jobs': 12}, 'updated_at': 'now()'}

I personally like it since it gives useful cluster information but you might feel like its noisy so i wanted to point it out

PR #745, opened 2023-02-22
Job restart on oomkilled

wahajXgrid:
### Issue

[Restart job on OOM kill and give it higher memory limits/requests ](https://github.com/robusta-dev/robusta/issues/551)

### Description

Kubernetes job can be failed due to oomkilled and sometimes the solution is to increase its container resources. This action will trigger when oomkilled occurs and restart the job with the given memory increment.
To prevent an infinite loop of the job's pod crashing and getting more memory this action takes the "Max" resource parameter from the user.

### Checklist
- [x] Manually tested
- [x] Made sure the title of the PR is a good description that can go into the release notes (if applicable)
- [x] Every function, interface, and class has a comment describing what it does and the input/output parameters


Sheeproid (approved):
Nice 👍

PR #749, opened 2023-02-26
Cluster provider: Detecting, Silencing + Enriching

Avi-Robusta:
TODO:

- [x]  Verify cluster detection on azure cluster
- [x]  Verify cluster detection on EKS cluster


Avi-Robusta:
Example of how alert_explanation_enricher looks
Screen Shot 2023-02-26 at 13 25 39

Avi-Robusta:
Added warning events for KubeContainerWaiting to show why the container is waiting in the event
(Didn't add logs since if container is waiting there shouldn't be logs)

Screen Shot 2023-02-27 at 11 25 51

src/robusta/core/model/cluster_status.py
```diff
@@ -9,6 +9,7 @@ class ClusterStats(BaseModel):
     pods: int
     nodes: int
     jobs: int
+    provider: str
```
Avi-Robusta, line 12: needs to technically be optional

src/robusta/utils/cluster_provider_discovery.py
```diff
@@ -0,0 +1,120 @@
...
+
+# the value is a node label unique to the provider
+NODE_LABELS: Dict[ClusterProviderType, str] = {
+    ClusterProviderType.Minikube: "minikube.k8s.io/name",
+    ClusterProviderType.DigitalOcean: "doks.digitalocean.com/version",
+    ClusterProviderType.Kops: "kops.k8s.io/instancegroup",
+    ClusterProviderType.Kapsule: "k8s.scaleway.com/kapsule"
+}
+
+__CLUSTER_PROVIDER__ = None
```
LeaveMyYard, line 35: Here I propose not to use a global variable, that is changed in other functions of this module, but rather to have a class, manage everything in it and then to have a global variable with an instance of this class - so the logic is encapsulated in that class

LeaveMyYard: If to call that class something like "ClusterProviderDetector" it will make sense that we have a static instance of that in module
This way we could avoid using `global` actually

src/robusta/utils/cluster_provider_discovery.py
```diff
@@ -0,0 +1,120 @@
...
+        __CLUSTER_PROVIDER__ = _find_cluster_provider().value
+        logging.info(f"{__CLUSTER_PROVIDER__} cluster detected")
+
+
+def _get_node_label(node, label) -> Optional[str]:
+    try:
+        return node.metadata.labels[label]
+    except KeyError:
+        pass
+    return None
```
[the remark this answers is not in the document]
LeaveMyYard: Maybe it is even better to not have another function for that - as this oneliner is also small enough

src/robusta/utils/cluster_provider_discovery.py
```diff
@@ -0,0 +1,120 @@
...
+        elif _is_detect_cluster_from_kubelet_version(nodes, 'gke'):
+            return ClusterProviderType.GKE
+        elif _is_detect_cluster_from_kubelet_version(nodes, 'eks'):
+            return ClusterProviderType.EKS
+
+        cluster_provider_from_labels = _detect_provider_from_node_labels(nodes)
+        if cluster_provider_from_labels != ClusterProviderType.Unknown:
+            return cluster_provider_from_labels
+
+    except Exception:
```
LeaveMyYard, line 118: It is actually bad to have `except Exception`. It is like saying "There might be some bug in the code", and we should always try to elliminate all bugs. Is it possible to try to catch possible errors in places they should raise? What are the exact places an error might occure? Maybe to decrease the try block to only those places?

I don't say we can't do with this, just saying we should at least try to elliminate it.
If it is not possible - move the try block to where the function is being called, as a lot of code in try block is not really good.

src/robusta/utils/cluster_provider_discovery.py
```diff
@@ -0,0 +1,120 @@
...
+        pass
+    return False
+
+
+def _is_detect_cluster_from_kubelet_version(nodes, kubelet_substring) -> bool:
+    try:
+        node = nodes[0]
+        kubelet_version = node.status.nodeInfo.kubeletVersion
+        if kubelet_substring in kubelet_version:
+            return True
```
LeaveMyYard, line 94: Could be written in one, as we will return False in any case, if this is False:

```python
return kublet_substring in kubelet_version
```

src/robusta/utils/cluster_provider_discovery.py
```diff
@@ -0,0 +1,120 @@
...
+    except (AttributeError, TypeError):
+        # missing kubeletVersion
+        pass
+    return False
```
LeaveMyYard, line 98: And if to change the first return, now we can remove `pass` and make the `return False` just in the try block, it will be a little more readable

LeaveMyYard: We could do the same on the above function, just a little bit more readable

LeaveMyYard (changes requested):

src/robusta/utils/cluster_provider_discovery.py
Avi-Robusta: Specifically here since it is being run at a highly sensitive part of robustas launch the safest thing is to swallow any exception. Since this is not operation critical code and robusta should be able to continue to fully function if there is any exceptions that do occur here it is future-error-proofing this code.

LeaveMyYard: Agree, then let's just move this try block to `discover_cluster_provider`, where we call this function

playbooks/robusta_playbooks/alerts_integration.py
```diff
@@ -127,6 +128,32 @@ def node_restart_silencer(alert: PrometheusKubernetesAlert, params: NodeRestartP
...
+    alert_explanation: str
+    recommended_resolution: Optional[str]
+
+
+@action
+def alert_explanation_enricher(alert: PrometheusKubernetesAlert, params: AlertExplanationParams):
+    """
+    Enrich the finding an explanation and recommendation of how to resolve the issue
+    """
+    blocks = [MarkdownBlock(
```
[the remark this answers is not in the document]
Avi-Robusta: Had a weird bug with black, it wasnt working for me
this was the fix i found
https://stackoverflow.com/questions/71673404/importerror-cannot-import-name-unicodefun-from-click

src/robusta/utils/cluster_provider_discovery.py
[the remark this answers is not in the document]
Avi-Robusta: Labels are an optional field and not guarenteed, so it will still be multiline to check it correnctly

src/robusta/utils/cluster_provider_discovery.py
```diff
@@ -18,103 +18,95 @@ class ClusterProviderType(Enum):
...
+
+    @staticmethod
+    def _detect_provider_from_hostname(nodes: List[Node]) -> Optional[ClusterProviderType]:
+        nodes_host_names = [ClusterProviderDiscovery._get_node_label(node, "kubernetes.io/hostname") for node in nodes]
+        for host_name in nodes_host_names:
+            if not host_name:
+                continue
+            for cluster_type in HOSTNAME_MATCH:
+                cluster_hostname_regex = HOSTNAME_MATCH[cluster_type]
+                if bool(re.match(cluster_hostname_regex, host_name)):
```
LeaveMyYard, line 66: bool is redundant

src/robusta/utils/cluster_provider_discovery.py
```diff
@@ -18,103 +18,95 @@ class ClusterProviderType(Enum):
...
 
-def _detect_provider_from_hostname(nodes) -> Optional[ClusterProviderType]:
-    nodes_host_names = [_get_node_label(node, "kubernetes.io/hostname") for node in nodes]
-    for host_name in nodes_host_names:
-        if not host_name:
-            continue
-        for cluster_type in HOSTNAME_MATCH:
-            cluster_hostname_regex = HOSTNAME_MATCH[cluster_type]
-            if len(re.findall(cluster_hostname_regex, host_name)) >= 1:
+class ClusterProviderDiscovery:
```
LeaveMyYard, line 37: Better avoid of using whole class of @staticmethod in favor of having a normal class with just one instance initiated in module global scope:

```python
   cluster_provider_discovery = ClusterProviderDiscovery()
   __all__ = ["cluster_provider_discovery"]
```

src/robusta/utils/cluster_provider_discovery.py
```diff
@@ -18,103 +18,95 @@ class ClusterProviderType(Enum):
...
+    def init_provider_discovery():
+        try:
+            ClusterProviderDiscovery.cluster_provider = ClusterProviderDiscovery._find_cluster_provider()
+            logging.info(f"{ClusterProviderDiscovery.cluster_provider.value} cluster discovered.")
+        except Exception:
+            logging.error("Error detecting cluster type", exc_info=True)
+
+    @staticmethod
+    def get_cluster_provider():
+        return ClusterProviderDiscovery.cluster_provider.value
```
LeaveMyYard, line 50: Add type annotation for ClusterProviderType and remove .value
If you want to use it as a string, do 
```python
class ClusterProviderType(str, Enum):  # Order is important, (Enum, str) will not work
    GKE = "GKE"
    AKS = "AKS"
    EKS = "EKS"
    Kind = "Kind"
    Minikube = "Minikube"
    RancherDesktop = "RancherDesktop"
    Kapsule = "Kapsule"
    Kops = "Kops"
    DigitalOcean = "DigitalOcean"
    Unknown = "Unknown"
```

LeaveMyYard (changes requested):

src/robusta/integrations/prometheus/trigger.py
```diff
@@ -75,6 +77,12 @@ def should_fire(self, event: TriggerEvent, playbook_id: str):
         if not prefix_match(self.instance_name_prefix, labels.get("instance")):
             return False
 
+        provider = cluster_provider.get_cluster_provider()
+        if provider and self.k8s_providers and len(self.k8s_providers) > 0:
```
LeaveMyYard, line 81:
```suggestion
        if provider and self.k8s_providers:
```

LeaveMyYard (approved):

PR #752, opened 2023-02-27
Related pods add containers

RoiGlinik:


playbooks/robusta_playbooks/k8s_resource_enrichments.py
```diff
@@ -43,32 +46,103 @@ def to_pod_row(pod: Pod, cluster_name: str) -> List:
...
+    elif kind == "Node":
         pods = Pod.listPodForAllNamespaces(field_selector=f"spec.nodeName={resource.metadata.name}").obj.items
     else:
         selector = build_selector_query(resource.spec.selector)
         pods = PodList.listNamespacedPod(namespace=resource.metadata.namespace, label_selector=selector).obj.items
 
+    return pods
+
+def to_pod_obj(pod: Pod, cluster: str) -> Dict[str,Any]:
+    resource_requests = pod_requests(pod)
```
Avi-Robusta, line 70: if we are already passing 70% of the pods yaml why don't we just pass the whole yaml of the pod and parse it on the the frontend?
it's less code in runner and more versatile for frontend?

playbooks/robusta_playbooks/k8s_resource_enrichments.py
```diff
@@ -43,32 +46,103 @@ def to_pod_row(pod: Pod, cluster_name: str) -> List:
...
+            "created": getattr(state, "startedAt", None)
+        })
+    
+    return containers
+ 
+
+
+
+@action
+def resource_related_pods(event: KubernetesResourceEvent):
```
Avi-Robusta, line 122: is the only difference that you added containers?
It might be word it to add the whole yaml so we dont have to make a new related pods for every new field we want.

playbooks/robusta_playbooks/k8s_resource_enrichments.py
```diff
@@ -43,32 +46,103 @@ def to_pod_row(pod: Pod, cluster_name: str) -> List:
...
         selector = build_selector_query(resource.spec.selector)
         pods = PodList.listNamespacedPod(namespace=resource.metadata.namespace, label_selector=selector).obj.items
 
+    return pods
+
+def to_pod_obj(pod: Pod, cluster: str) -> Dict[str,Any]:
+    resource_requests = pod_requests(pod)
+    resource_limits = pod_limits(pod)
+    addresses = ",".join([address.ip for address in pod.status.podIPs])
+    return {
```
Avi-Robusta, line 73: we generally don't just create Dicts to return if it has a explicit format, I would create a class like 
class RelatedPods(BaseModel) to return for type/parameter validation

(to_pod_row also seems to not be following that rule )

RoiGlinik:
@Avi-Robusta I will look into these changes today I think. It might break some of your apps page

playbooks/robusta_playbooks/k8s_resource_enrichments.py
RoiGlinik: I will create a specific class for the format

src/robusta/core/sinks/robusta/dal/model_conversion.py
```diff
@@ -146,6 +147,8 @@ def to_evidence_json(
                     )
 
                 structured_data.append({"type": "callbacks", "data": callbacks})
+            elif isinstance(block, JsonBlock):
```
Avi-Robusta, line 150: how does this look in the UI and slack?

Avi-Robusta: Also this might send JsonBlock to all the sinks now, there might have been a reason why this wasnt allowed until now, maybe check where the finding is sent without this part?

playbooks/robusta_playbooks/k8s_resource_enrichments.py
```diff
@@ -43,32 +72,105 @@ def to_pod_row(pod: Pod, cluster_name: str) -> List:
...
+    """
+    Return the list of pods and containers related to that k8s resource.
+    For example, return all pods of a given k8s Deployment
+
+    Supports Deployments, ReplicaSets, DaemonSets, StatefulSets and Pods
+    """
+    pods = get_related_pods(event.get_resource())
+    cluster = event.get_context().cluster_name
+    event.add_enrichment([
+       JsonBlock(json.dumps([to_pod_obj(pod, cluster).dict() for pod in pods]))
```
arikalon1, line 161: I don't understand why you're using json blocks, that can only be used by one sink

src/robusta/core/sinks/robusta/dal/model_conversion.py
arikalon1: yes, because it why would you send json to slack

playbooks/robusta_playbooks/k8s_resource_enrichments.py
arikalon1: In addition, this will work only on newer runner versions.
Why would we want that?

RoiGlinik:
@Avi-Robusta  Could you do a quick last scan before I merge

playbooks/robusta_playbooks/k8s_resource_enrichments.py
```diff
@@ -20,6 +25,38 @@
     pod_restarts,
 )
 
+class RelatedPodParams(ActionParams):
+    """
+    :var block_type: output format type of the action.
+    """
+    output_format: str = "table"  # table, json
```
Avi-Robusta, line 32: The comment "block_type" and variable name "output_format" don't match, i would also add the types to the comment so it will automatically show up in the docs

personally since its only two types I would do something like output_as_table: bool =True but what you wrote is okay too

Avi-Robusta (approved):
looks good, one small change in a comment you left

Before you merge:
- [ ] check how table looks on slack now

Avi-Robusta:
two run_tests tests are failing here

- [ ] check to see if its your branch or master with the issue

PR #760, opened 2023-03-03
Issue 755

arikalon1:
Enable loading playbook repo url from env var
Bug fixes:
Take alert manager auth from SecretStr
Remove ending slash from alertmanager_url, if exists
Trace exception on alertmanager calls

### DIFF
--- a/playbooks/robusta_playbooks/silence.py
+++ b/playbooks/robusta_playbooks/silence.py
@@ -1,11 +1,12 @@
 import json
+import logging
 from datetime import datetime
 from enum import Enum
 from typing import Dict, List, Optional
 from uuid import UUID
 
 import requests
-from pydantic import BaseModel, SecretStr
+from pydantic import BaseModel, SecretStr, validator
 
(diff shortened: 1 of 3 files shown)


playbooks/robusta_playbooks/silence.py
```diff
@@ -69,6 +70,12 @@ class BaseSilenceParams(ActionParams):
     alertmanager_auth: Optional[SecretStr] = None
     grafana_api_key: str = None  # type: ignore
 
+    @validator("alertmanager_url", allow_reuse=True)
```
LeaveMyYard, line 73: I think `allow_reuse` is redundant (https://docs.pydantic.dev/usage/validators/#reuse-validators)

LeaveMyYard (approved):

playbooks/robusta_playbooks/silence.py
arikalon1: I added it, because reloading playbooks runs the decorator multiple times

PR #774, opened 2023-03-09
Enhancement: Added - 'The account name you entered is already in use.…

ganeshrvel:
Enhancement: Added - 'The account name you entered is already in use' error text.

Prompt the user with a suggested account name


arikalon1 (approved):
LGTM

PR #777, opened 2023-03-12
Imagepull backoff improvements

Avi-Robusta:
updated to its own smart trigger and added a fire_delay param to reduce noise on false backoffs on pod startup.

There is no way to count how many backoffs occurred so a time delay was the best option. 
If the trigger doesn't fire after the time delay than the imagepull was resolved


Avi-Robusta:
the time delay might be useful for crashloop backoffs too since sometimes pods crash due to missing resources that are being initiated. (like PV, pvc's)

src/robusta/core/triggers/pod_crash_loop_trigger.py
```diff
@@ -48,6 +55,19 @@ def should_fire(self, event: TriggerEvent, playbook_id: str):
...
+        run_time_seconds = 0
+
+        # startTime does not exist every time pod update is fired, like when the pod is just created
+        if pod.status.startTime:
+            run_time_seconds = (
+                datetime.utcnow().timestamp() - parse_kubernetes_datetime_to_ms(pod.status.startTime) / 1000
+            )
+
+        # sometimes pods can fire falsely on pod startup
+        # due to not loading a needed component
```
arikalon1, line 67: I'm not sure what you mean in that comment.
Regardless, do we really need both checks here?
Both checking the pod is running for more than 2 minutes, and that it had more than 2 restarts?

src/robusta/core/triggers/pod_image_pull_backoff.py
```diff
@@ -0,0 +1,79 @@
...
+from robusta.integrations.kubernetes.api_client_utils import parse_kubernetes_datetime_to_ms
+from robusta.integrations.kubernetes.autogenerated.triggers import PodChangeEvent, PodUpdateTrigger
+from robusta.integrations.kubernetes.base_triggers import K8sTriggerEvent
+from robusta.utils.rate_limiter import RateLimiter
+
+
+class PodImagePullBackoffTrigger(PodUpdateTrigger):
+    """
+    :var rate_limit: Limit firing to once every `rate_limit` seconds
+    :var fire_delay: Limit firing for a pod that has run atleast `fire_delay` seconds
```
arikalon1, line 13: same comment suggestion here

arikalon1 (approved):
nice
left few minor comments

playbooks/robusta_playbooks/image_pull_backoff_enricher.py
```diff
@@ -48,20 +47,18 @@ def image_pull_backoff_reporter(event: PodEvent, action_params: RateLimitParams)
...
     namespace = pod.metadata.namespace
 
-    # Perform a rate limit for this pod according to the rate_limit parameter
-    if not RateLimiter.mark_and_test(
-        "image_pull_backoff_reporter",
-        namespace + ":" + replicaset_name,
-        action_params.rate_limit,
-    ):
-        return
+    if action_params.rate_limit:
```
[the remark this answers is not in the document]
Avi-Robusta: good catch

src/robusta/core/triggers/pod_crash_loop_trigger.py
Avi-Robusta: I think the multiple restarts would still reduce noise like we do now, 
if a pod ran for an hour and then randomly crashed but isn't looping, than since we only report on the second crash we wont report it.
I wouldnt remove it but if you would like to report every crash I would set the default restart_count to 0 and not remove it incase users want to use it to reduce noise.
WDYT?

Issue #728, opened 2023-02-10
Bug in HPA alert enrichment


luigiaparicio:
**Describe the bug**
I have a Deployment and HPA configured to autoscaling Pods by CPU utilization.

When HPA scale the Deployment to the maximum number of replicas, Robusta notifies "HPA reached max replicas" and inside the message nforms a "Current avg cpu utilization" percent that is lower than the HPA target CPU configured in HPA.
In my case target is configured at 80%, and Robusta alerts say 18% current 

**To Reproduce**
1. Generate load inside the pod, to increase the CPU usage
2. HPA scales up to max replicas because current is higher than target. 
3. Wait for Robusta to detect this behavior and alert
4. Look in the alert message and see that current is lower than HPA target cpu threshold

**Expected behavior**
Current percent informed should be higher than HPA target cpu threshold

**Screenshots**
<img width="596" alt="Screenshot 2023-02-10 at 10 12 11" src="https://user-images.githubusercontent.com/6013828/218124144-46c93e71-d111-49d9-b120-aa2e3aef8ca9.png">

**Additional context**
This tests were done in an OpenShift cluster version 4.10 and notifications received in Slack

PR #786, opened 2023-03-19
average calculation precentage fixed

Avi-Robusta:
currentCPUUtilizationPercentage is already the average

https://github.com/robusta-dev/robusta/issues/728

### DIFF
--- a/playbooks/robusta_playbooks/autoscaler.py
+++ b/playbooks/robusta_playbooks/autoscaler.py
@@ -69,10 +69,7 @@ def alert_on_hpa_reached_limit(event: HorizontalPodAutoscalerChangeEvent, action
     if hpa.status.desiredReplicas != hpa.spec.maxReplicas:
         return  # didn't reached max replicas limit
 
-    avg_cpu = int(
-        hpa.status.currentCPUUtilizationPercentage
-        / (hpa.status.currentReplicas if hpa.status.currentReplicas > 0 else 1)
-    )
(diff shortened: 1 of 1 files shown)


playbooks/robusta_playbooks/autoscaler.py
```diff
@@ -69,10 +69,7 @@ def alert_on_hpa_reached_limit(event: HorizontalPodAutoscalerChangeEvent, action
     if hpa.status.desiredReplicas != hpa.spec.maxReplicas:
         return  # didn't reached max replicas limit
 
-    avg_cpu = int(
-        hpa.status.currentCPUUtilizationPercentage
-        / (hpa.status.currentReplicas if hpa.status.currentReplicas > 0 else 1)
-    )
+    avg_cpu = int(hpa.status.currentCPUUtilizationPercentage)
```
arikalon1, line 72: @Avi-Robusta 
This value isn't exactly avg cpu usage. 
From the docs:
```
current average CPU utilization over all pods, represented as a percentage of requested CPU, e.g. 70 means that an average pod is using now 70% of its requested CPU.
```

Do you think it will help to change the text to something like:
`On average, pods scaled under this HPA are using 75% of the requested cpu`

@aantn thoughts?

aantn: @arikalon1 yes, I like that text. Super clear to me. @Avi-Robusta can we update it?

Avi-Robusta: on it

PR #789, opened 2023-03-21
Fix GitHub test action

RoiGlinik:


RoiGlinik:
Use a fixed poetry version for the moment

PR #792, opened 2023-03-22
log enricher regex + PrometheusRuleFailures playbook 

Avi-Robusta:
adds regex filter to log_enricher + playbook for PrometheusRuleFailures to bring only relevant logs
It is important to note that the regex filter is run before our regex replacement 
so a filter can return all lines that contain the secret XYZ for example but the regex replacement will print it as ***

### DIFF
--- a/playbooks/robusta_playbooks/prometheus_enrichments.py
+++ b/playbooks/robusta_playbooks/prometheus_enrichments.py
@@ -4,13 +4,12 @@ from typing import List, Optional, Tuple, Union
 
 from prometheus_api_client import PrometheusApiClientException
 from kubernetes import client
-from kubernetes.client import V1ServiceList
 from kubernetes.client.models.v1_service import V1Service
 from robusta.api import (
     ExecutionBaseEvent,
-    PrometheusAlert,
     PrometheusDateRange,
     PrometheusDuration,
+    PrometheusKubernetesAlert,
     PrometheusQueryParams,
     PrometheusQueryResult,
     action,
@@ -94,7 +93,7 @@ def get_duplicate_kubelet_msg(rule_group) -> str:
 
 
 @action
-def prometheus_rules_enricher(alert: PrometheusAlert):
+def prometheus_rules_enricher(alert: PrometheusKubernetesAlert):
     """
     Enriches the finding with a for information due to rule failure
     """
--- a/playbooks/robusta_playbooks/prometheus_enrichments.py
+++ b/playbooks/robusta_playbooks/prometheus_enrichments.py
@@ -1,10 +1,13 @@
 import logging
+import re
 from datetime import datetime, timedelta
-from typing import Optional, Tuple, Union
+from typing import List, Optional, Tuple, Union
 
 from prometheus_api_client import PrometheusApiClientException
 from robusta.api import (
+    ActionParams,
     ExecutionBaseEvent,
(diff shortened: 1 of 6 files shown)


playbooks/robusta_playbooks/prometheus_enrichments.py
```diff
@@ -1,16 +1,19 @@
...
     ExecutionBaseEvent,
     PrometheusDateRange,
     PrometheusDuration,
+    PrometheusKubernetesAlert,
     PrometheusQueryParams,
     action,
     run_prometheus_query,
 )
-from robusta.core.reporting.blocks import PrometheusBlock
+from robusta.core.reporting.blocks import MarkdownBlock, PrometheusBlock
```
arikalon1, line 16: don't import from robusta.core in playbooks. use robusta.api

playbooks/robusta_playbooks/prometheus_enrichments.py
```diff
@@ -61,3 +64,32 @@ def prometheus_enricher(event: ExecutionBaseEvent, params: PrometheusQueryParams
     event.add_enrichment(
         [PrometheusBlock(data=prometheus_result, query=params.promql_query)],
     )
+
+
+def get_duplicate_kubelet_msg(rule_group) -> str:
```
arikalon1, line 69: can you add type annotation to rule_group?

playbooks/robusta_playbooks/prometheus_enrichments.py
```diff
@@ -61,3 +64,32 @@ def prometheus_enricher(event: ExecutionBaseEvent, params: PrometheusQueryParams
...
+    """
+    checks if there are multiple kubelets and
+    """
+    NO_MESSAGE = ""
```
arikalon1, line 73: Is it better to have a return value of Optional[str] and return None instead of NO_MESSAGE?

playbooks/robusta_playbooks/prometheus_enrichments.py
```diff
@@ -61,3 +64,32 @@ def prometheus_enricher(event: ExecutionBaseEvent, params: PrometheusQueryParams
...
+    if not rule_group or "kubelet" not in rule_group:
+        return NO_MESSAGE
+    labels_selectors = "k8s-app=kubelet, app.kubernetes.io/managed-by=prometheus-operator"
+    v1 = client.CoreV1Api()
+    kubelet_services: List[V1Service] = v1.list_service_for_all_namespaces(label_selector=labels_selectors).items
```
arikalon1, line 78: Is there a way to build a generic solution, that find this error for any duplicated service, rather than a hard coded solution only for kubelet?

arikalon1 (approved):
looks good

Left some comments

playbooks/robusta_playbooks/prometheus_enrichments.py
Avi-Robusta: Duplicate services might not always cause this issue. (Specifically kubelet should always be scraped due to there being a ServiceMonitor for it)

Screen Shot 2023-04-13 at 21 37 06

It would be better to use the prometheus targets api to verify the duplicate services are being scraped. I will update this code when I finish the target down enricher to make it more generic. (I already have code to get the prometheus targets there) I'll note to add it in that PR.

PR #803, opened 2023-03-30
Aggregated api enricher

Avi-Robusta:
add default enricher for KubeAggregatedAPIDown
Either display yaml of apiservice or if the service does not exist display relevant message

image
image


playbooks/robusta_playbooks/api_service.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+class ApiService(BaseModel):
+    name: str
+    service_name: Optional[str]
+    service_namespace: Optional[str]
+    conditions: List[ApiServiceConditions]
+    full_json: dict
+
+    def get_json(self):
```
RoiGlinik, line 25: I think you can delete this function. Seems to not be used

playbooks/robusta_playbooks/api_service.py
```diff
@@ -0,0 +1,139 @@
...
+    def get_available_condition(self) -> Optional[ApiServiceConditions]:
+        if not self.conditions:
+            return None
+        available_conditions = [condition for condition in self.conditions if condition.type == "Available"]
+        if not available_conditions:
+            return None
+        return available_conditions[0]
+
+    @staticmethod
+    def parse_apiservice(apiservice: dict, name: str):
```
RoiGlinik, line 46: I think it will be more concise to move the code to the constructor , if its possible

playbooks/robusta_playbooks/api_service.py
```diff
@@ -0,0 +1,139 @@
...
+        api_services_json: V1APIServiceList = (
+            ApiregistrationV1Api()
+            .list_api_service(field_selector=field_selector, _preload_content=False)
+            .data.decode("utf-8")
+        )
+        api_services_json_list = json.loads(api_services_json).get("items", [])
+        if len(api_services_json_list) == 0:
+            logging.info(f"Api service {name} not found.")
+            return None
+        if len(api_services_json_list) > 1:
```
RoiGlinik, line 72: I am not sure if its a real use case or more of a sanity check. 
If its possible ( 2 services that expose the same api service? ) it might be necessary to handle that case.
But I am not sure if this case is even possible

RoiGlinik (approved):
Left some minor comments . 👍

playbooks/robusta_playbooks/api_service.py
Avi-Robusta: Shouldn't be possible but its a sanity check since list_api_service has issues and is an old version.

PR #804, opened 2023-03-30
feat: use slack upload v2 api

daviderli614:
Deployed robusta-runner in kubernetes(helm-chart version 0.10.14), having a warn logs:
```
/usr/local/lib/python3.9/site-packages/slack_sdk/web/internal_utils.py:450: UserWarning: client.files_upload() may cause some issues like timeouts for relatively large files. Our latest recommendation is to use client.files_upload_v2(), which is mostly compatible and much stabler, instead.
  warnings.warn(message)
```
Use files_upload_v2 instead of files_upload will be better.

### DIFF
--- a/src/robusta/integrations/slack/sender.py
+++ b/src/robusta/integrations/slack/sender.py
@@ -191,7 +191,7 @@ class SlackSender:
         with tempfile.NamedTemporaryFile() as f:
             f.write(block.contents)
             f.flush()
-            result = self.slack_client.files_upload(title=block.filename, file=f.name, filename=block.filename)
+            result = self.slack_client.files_upload_v2(title=block.filename, file=f.name, filename=block.filename)
             return result["file"]["permalink"]
 
     def prepare_slack_text(self, message: str, files: List[FileBlock] = []):


arikalon1 (commented):
Thank you for your contribution @daviderli614 

I tried to run it, and got an error from the Slack API:
```
slack_sdk.errors.SlackApiError: The request to the Slack API failed. (url: https://www.slack.com/api/files.info)
The server responded with: {'ok': False, 'error': 'missing_scope', 'needed': 'files:read', 'provided': 'files:write,chat:write,chat:write.public'}
=
```

Seems like the new upload api requires `files:read` scope, which is missing from our Slack app.
Were you able to upload a file successfully?

daviderli614:
Hi @arikalon1 
The files_upload_v2 api seems to require the `files:read` permission, which may require the robusta bot to enable this permission, or set the parameter `request_file_info=False`, which one do you think is better?

Here are some references I've seen:
https://github.com/slackapi/python-slack-sdk/issues/1292       
https://github.com/slackapi/python-slack-sdk/issues/1286

arikalon1:
Hi @daviderli614,
Seems like the Slack sdk created a non-backward compatible change.
I read the issues you pointed above, and it's about the same issue.
Sadly, we can't force all our Slack app users to re-install the app.
So it looks like we have no choice but to keep using the old `file_upload` implementation

daviderli614:
Thank

PR #808, opened 2023-04-01
pod_status_silencer - silence alerts based on pod status. (For exampl…

arikalon1:
…e, Evicted pods)

delete_job action

### DIFF
--- a/playbooks/robusta_playbooks/alerts_integration.py
+++ b/playbooks/robusta_playbooks/alerts_integration.py
@@ -7,6 +7,7 @@ from typing import Any, Dict, List, Optional
 
 import requests
 from hikaru.model import Node
+
 from robusta.api import (
     ActionParams,
     AlertResourceGraphEnricherParams,
@@ -82,6 +83,39 @@ def silence_alert(alert: PrometheusKubernetesAlert, params: SilenceAlertParams):
     if params.log_silence:
         logging.info(f"silencing alert {alert}")
     alert.stop_processing = True
+
+
+class StatusSilencerParams(ActionParams):
+    """
+    :var include: If available, will stop processing unless the pod status is in the include list
+    :var exclude: If available, will stop processing if the pod status is in the exclude list
+
+    :example include: ["Pending"]
+    :example exclude: ["Evicted"]
+    """
+
+    include: Optional[List[str]]
+    exclude: Optional[List[str]]
+
+
+@action
+def pod_status_silencer(event: PodEvent, params: StatusSilencerParams):
+    """
+    Stop execution based on pod statuses.
+    """
+    pod = event.get_pod()
+    if not pod:
+        logging.info("Cannot run pod_status_silencer with no pod. skipping")
+        return
+
+    if params.include:  # Stop unless pod status in include list
+        if pod.status.phase not in params.include:
+            event.stop_processing = True
+            return
+
+    if params.exclude:
+        if pod.status.phase in params.exclude:
+            event.stop_processing = True
 
 
 @action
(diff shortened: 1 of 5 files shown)


playbooks/robusta_playbooks/alerts_integration.py
```diff
@@ -84,6 +85,39 @@ def silence_alert(alert: PrometheusKubernetesAlert, params: SilenceAlertParams):
     alert.stop_processing = True
 
 
+class StatusSilencerParams(ActionParams):
+    """
+    :var include: If available, will stop processing unless the pod status is in the include list
+    :var exclude: If available, will stop processing if the pod status is in the exclude list
+
+    :example include: ["Pending"]
+    :example exclude: ["Evicted"]
```
Avi-Robusta, line 94: the status's for phase are only: pending, running, succeded, failed, unknown
Evicted isn't possible for phase
https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle#pod-phase

playbooks/robusta_playbooks/alerts_integration.py
```diff
@@ -84,6 +85,39 @@ def silence_alert(alert: PrometheusKubernetesAlert, params: SilenceAlertParams):
...
+    """
+    Stop execution based on pod statuses.
+    """
+    pod = event.get_pod()
+    if not pod:
+        logging.info("Cannot run pod_status_silencer with no pod. skipping")
+        return
+
+    if params.include:  # Stop unless pod status in include list
+        if pod.status.phase not in params.include:
```
Avi-Robusta, line 112: are you sure you want to use phase for the status's
Its very high level. 
Is that what you want to filter or do you want to filter out more specific status like crashloop, oomkill, imagepull?
because allot of them will be contained in the the phase running like crashloop for example:

`containerStatuses:
  - containerID: containerd://656cd4aca1ea7cba7821dfe572908308709f3570898fd35518fb085fb026651c
    image: docker.io/library/busybox:latest
    lastState:
 ...
    state:
      waiting:
        message: back-off 20s restarting failed container=crashpod pod=crashpod-64b874bbbc-8rqts_default(cb100c54-b174-4d87-b48d-1e54c1c375b4)
        reason: CrashLoopBackOff
  phase: Running
`

arikalon1: The user asked for phase. Not sure if "Evicted" is a phase

playbooks/robusta_playbooks/alerts_integration.py
arikalon1: You have a point.
I'm not sure "Evicted" is a phase.
Going to change the example

playbooks/robusta_playbooks/job_actions.py
```diff
@@ -236,6 +236,23 @@ def job_info_enricher(event: JobEvent):
...
+    """
+    job = event.get_job()
+    if not job:
+        logging.info("Cannot run delete_job with no job. skipping")
+        return
+
+    # After deletion the metadata is empty. Saving the name and namespace
+    name = job.metadata.name
+    namespace = job.metadata.namespace
+    job.delete()
```
Avi-Robusta, line 252: FYI kubernetes could return an error here in job.delete, I think it might be printed in the logs so i don't think you need to add anything here
```
      if not 200 <= res.code <= 299:
          raise KubernetesException("Kubernetes returned error " + str(res.code))
```
maybe if it comes up we can investigate and add an enricher like "Job not deleted because of X"

Avi-Robusta (approved):

playbooks/robusta_playbooks/job_actions.py
```diff
@@ -236,6 +236,23 @@ def job_info_enricher(event: JobEvent):
     event.add_enrichment([table_block])
 
 
+@action
+def delete_job(event: JobEvent):
```
Avi-Robusta, line 240: maybe add to the light actions list, this could be useful for frontend

arikalon1: good idea

arikalon1: done

PR #811, opened 2023-04-04
Hpa support

Avi-Robusta:
uses features from PR
https://github.com/robusta-dev/robusta/pull/803
will merge to master after this PR
We list the replias+ events + if there are metrics-server issues we alert the user how to resolve since HPA requires metrics server 
for full metrics server installation
Screen Shot 2023-04-04 at 17 21 41

for a broken apiservice
Screen Shot 2023-04-04 at 17 23 01


arikalon1:
@Avi-Robusta In all the screenshots desired=running.
Is this how it will look on a real alert? (isn't the alert is when desired != running ?)

Avi-Robusta:
@arikalon1 you are right, it takes like 15 minutes to fire a real alert so I was using simulated ones, here is a real one
Screen Shot 2023-04-09 at 11 28 42

playbooks/robusta_playbooks/prometheus_simulation.py
```diff
@@ -13,6 +13,7 @@ class PrometheusAlertParams(ActionParams):
     :var deployment_name: Deployment name, for a simulated deployment alert.
     :var container_name: Container name, for adding a label on container.
     :var job_name: Job name, for a simulated Job alert.
+    :var hpa: Job name, for a simulated HorizontalPodAutoscaler alert.
```
RoiGlinik, line 16: typo?

src/robusta/core/reporting/consts.py
```diff
@@ -1,3 +1,4 @@
+import logging
```
RoiGlinik, line 1: I think its slipped here?

playbooks/robusta_playbooks/autoscaler.py
```diff
@@ -1,17 +1,25 @@
+import json
```
RoiGlinik, line 1: Also not sure this is needed

RoiGlinik (approved):
Left some minor comments , but can merge after checking them.

playbooks/robusta_playbooks/autoscaler.py
Avi-Robusta: removed

PR #815, opened 2023-04-06
Improve support for running robusta locally

aantn:


run_runner_locally.sh
```diff
@@ -0,0 +1,74 @@
...
+poetry install --extras=all
+echo "All dependencies installed"
+
+
+echo "Setting up local runner environment"
+mkdir -p deployment/playbooks/defaults
+ln -fsw $(pwd)/playbooks/robusta_playbooks/ ./deployment/playbooks/defaults
+ln -fsw $(pwd)/playbooks/pyproject.toml ./deployment/playbooks/defaults
+
+echo "Checking if runner can listen on port 5006"
```
arikalon1, line 44: I would use port 5000.
I think the use of 5006 was because port 5000 was taken on a developer machine.
Typically it should be available.

run_runner_locally.sh
```diff
@@ -0,0 +1,74 @@
...
+    exit 1
+fi
+
+echo "Checking for active_playbooks.yaml"
+# warning: this doesn't actually run independent of the cluster - e.g. scheduler dal uses cluster state
+if [ ! -f "$PWD/deployment/playbooks/active_playbooks.yaml" ]; then
+  echo -e "${RED}Error: You need an active_playbooks.yaml to run Robusta locally!${NC}"
+  echo "This file must be located at ./deployment/playbooks/active_playboks.yaml"
+  echo "You can fix by extracting this file from an existing cluster with Robusta. Run:"
+  echo -e "\tkubectl get secret robusta-playbooks-config-secret -o jsonpath='{.data.active_playbooks\.yaml}'  | base64 --decode > ./deployment/playbooks/active_playbooks.yaml"
```
arikalon1, line 56: This command will not be executed, right?
The comment on the readme suggest it's running automatically ("The runner needs a configuration file to run and it extracts it from your existing cluster.")
Would be nice to just run it, and set the cluster name for you. (pass a cluster name to the script)
Was there a reason for not doing it?

src/robusta/cli/playbooks_cmd.py
```diff
@@ -334,13 +335,29 @@ def trigger(
...
     req_body = {"action_name": action_name, "action_params": action_params}
 
-    _post_in_runner_pod(
-        namespace=namespace,
-        api_path="trigger",
-        req_body=req_body,
-        req_name="trigger action",
-        dry_run=dry_run,
+    cmd = (
+        f"curl -X POST http://localhost:5000/api/trigger "
```
arikalon1, line 339: I think this doesn't match the 5006

src/robusta/integrations/git/git_repo.py
```diff
@@ -12,7 +12,7 @@
 from robusta.core.model.env_vars import GIT_MAX_RETRIES
 
 GIT_DIR_NAME = "robusta-git"
-REPO_LOCAL_BASE_DIR = os.path.join(os.environ.get("REPO_LOCAL_BASE_DIR", "/app"), GIT_DIR_NAME)
+REPO_LOCAL_BASE_DIR = os.path.abspath(os.path.join(os.environ.get("REPO_LOCAL_BASE_DIR", "/app"), GIT_DIR_NAME))
```
arikalon1, line 15: can you explain?

arikalon1 (approved):
very nice!
We should have done this long ago

Left some minor comments

arikalon1:
one more comment.
Because you ran the runner locally, it changed your git user to "Robusta Runner"
Now, parts of the commits are registered under the "Robusta Runner" which didn't sign the CLA...

src/robusta/integrations/git/git_repo.py
aantn: I don't recall what the issue was, but added it to fix an issue with relative paths.

playbooks/robusta_playbooks/prometheus_simulation.py
```diff
@@ -3,6 +3,7 @@
 
 import requests
 from robusta.api import ActionParams, AlertManagerEvent, ExecutionBaseEvent, action
+from robusta.core.model.env_vars import PORT
```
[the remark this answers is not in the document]
aantn: @arikalon1 right now env_vars aren't available there. In the name of speed I'm going to merge as is, but feel free to change later.

run_runner_locally.sh
aantn: Fixed

run_runner_locally.sh
aantn: I don't want to do it automatically because I've run commands with the wrong kubectx before. I thought it's safer to just ask you to do it manually.

But open to improving it later. In the name of speed, just merging as is.

PR #818, opened 2023-04-09
returning error result

Avi-Robusta:
for prometheus_enricher returns internal prometheus exceptions in the finding

### DIFF
--- a/playbooks/robusta_playbooks/prometheus_enrichments.py
+++ b/playbooks/robusta_playbooks/prometheus_enrichments.py
@@ -2,11 +2,13 @@ import logging
 from datetime import datetime, timedelta
 from typing import Optional, Tuple, Union
 
+from prometheus_api_client import PrometheusApiClientException
 from robusta.api import (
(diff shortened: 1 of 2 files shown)


playbooks/robusta_playbooks/prometheus_enrichments.py
```diff
@@ -51,13 +53,18 @@ def prometheus_enricher(event: ExecutionBaseEvent, params: PrometheusQueryParams
...
+        prometheus_result = run_prometheus_query(
+            prometheus_params=params,
+            promql_query=params.promql_query,
+            starts_at=starts_at,
+            ends_at=ends_at,
+        )
+        event.add_enrichment(
+            [PrometheusBlock(data=prometheus_result, query=params.promql_query)],
+        )
+    except PrometheusApiClientException as e:
```
RoiGlinik, line 66: @Avi-Robusta I think the query class might not be able to work with that. Wrote to you about it

RoiGlinik (approved):

PR #826, opened 2023-04-12
added list_resource_names action

arikalon1:
added support for NetworkPolicy and PersistentVolumeVlaim hikaru objects


playbooks/robusta_playbooks/k8s_resource_enrichments.py
```diff
@@ -238,3 +241,26 @@ def get_resource_yaml(event: KubernetesResourceEvent):
...
+    :var kind: The k8s resource kind. Must be one of: [node,deployment,statefulset,daemonset,job,persistentvolume,persistentvolumeclaim,service,configmap,networkpolicy].
+    :var namespace: For namespaced k8s resources. List names for the specified namespace. If omitted, all namespaces will be used
+    """
+
+    kind: str
+    namespace: Optional[str]
+
+
+@action
+def list_resource_names(event: ExecutionBaseEvent, params: NamedResourcesParams):
```
Avi-Robusta, line 257: @arikalon1 should we add this to light actions? if it just lists the names it should be okay.

src/robusta/core/discovery/resource_names.py
```diff
@@ -0,0 +1,75 @@
...
+    "configmap": ResourceLister(
+        list_all=client.CoreV1Api().list_config_map_for_all_namespaces,
+        list_namespaced=client.CoreV1Api().list_namespaced_config_map,
+    ),
+}
+
+
+class ResourceNameLister:
+    @staticmethod
+    def list_resource_names(kind: str, namespace: str = None) -> List[str]:
```
Avi-Robusta, line 58: well done!

src/robusta/core/discovery/resource_names.py
```diff
@@ -0,0 +1,75 @@
...
+                error=ErrorCodes.RESOURCE_NOT_SUPPORTED, msg=f"Listing names of {kind} is not supported"
+            )
+        lister = LISTERS[kind.lower()]
+        if namespace:
+            namespace_lister = lister.list_namespaced
+            if not namespace_lister:
+                raise ActionException(
+                    error=ErrorCodes.ILLEGAL_ACTION_PARAMS,
+                    msg=f"Listing names of {kind} for a specific namespace is not supported",
+                )
```
Avi-Robusta, line 70: Listing the resources could be a heavy process and cause oomkills like we had with discovery so we switched to batch loading. Depending on what amount of resources we are loading you might want to change to batch loading the names.

Avi-Robusta (approved):
Looks good,
only thoughts are that `list_resource_names` looks like it would be a good light action.
And potentially loading the resources might cause an oomkill like we had with discovery so we might want to change it to batch load the resources. 
(OOMKill here might not be applicable here because here we are only loading one resource type and in discovery we loaded multiple so it wont be the same load, so your call)

playbooks/robusta_playbooks/k8s_resource_enrichments.py
arikalon1: yes! Good catch
It's a must

src/robusta/core/discovery/resource_names.py
arikalon1: thanks

src/robusta/core/discovery/resource_names.py
arikalon1: I think it's ok to start without batch loading

Avi-Robusta:
@arikalon1 just verify that we have no issues listing NetworkPolicy since I saw we didn't do any changes in runners ClusterRole

PR #828, opened 2023-04-14
added check for requests in node_pods_capacity_enricher

Avi-Robusta:
added amount of pods with requests in node_pods_capacity_enricher
Screen Shot 2023-04-14 at 10 26 38


arikalon1 (approved):
LGTM

PR #832, opened 2023-04-17
popeye action hardening

RoiGlinik:
image is just adjustable from env var
quote args
fix params baseclass

### DIFF
--- a/playbooks/robusta_playbooks/popeye.py
+++ b/playbooks/robusta_playbooks/popeye.py
@@ -131,7 +131,7 @@ def popeye_scan(event: ExecutionBaseEvent, params: PopeyeParams):
     Displays a popeye scan report.
     """
 
-    sanitize_args = " ".join([f"'{arg}'" for arg in shlex.split(params.args)])
+    sanitize_args = shlex.join(shlex.split(params.args))
     spec = PodSpec(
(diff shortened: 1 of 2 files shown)


playbooks/robusta_playbooks/popeye.py
```diff
@@ -129,16 +131,17 @@ def popeye_scan(event: ExecutionBaseEvent, params: PopeyeParams):
     Displays a popeye scan report.
     """
 
+    sanitize_args = " ".join([f"'{arg}'" for arg in shlex.split(params.args)])
```
RoiGlinik, line 134: @arikalon1  had to change it to this after some extra tests

LeaveMyYard (approved):

playbooks/robusta_playbooks/popeye.py
RoiGlinik: Yea ill do a quick check tomorrow , it might be better to change it to **shlex.join**. and replace my **f"'{arg}'"** with the shlex.quote.

LeaveMyYard: If you will, could you also do the same for krr playbook? It has similar code there

RoiGlinik: yea np

PR #833, opened 2023-04-17
KRR scan action

LeaveMyYard:
Robusta playbook for running the KRR scan inside the cluster


playbooks/robusta_playbooks/krr.py
```diff
@@ -0,0 +1,195 @@
...
+    MarkdownBlock,
+    ProcessParams,
+    RobustaJob,
+    ScanReportBlock,
+    ScanReportRow,
+    ScanType,
+    action,
+    to_kubernetes_name,
+)
+from robusta.core.model.env_vars import RELEASE_NAME
```
arikalon1, line 25: please use from robusta.api

playbooks/robusta_playbooks/krr.py
```diff
@@ -0,0 +1,195 @@
...
+
+
+class KRRObject(BaseModel):
+    cluster: Optional[str]
```
arikalon1, line 29: are all the fields really manadatory?

arikalon1 (approved):
Looks good
Left small comments

playbooks/robusta_playbooks/krr.py
LeaveMyYard: This is just exactly the format we get from the Job

LeaveMyYard: This is just exactly the format we get from the Job

PR #839, opened 2023-04-24
Fixed graph queries with multiple prometheus metrics jobs

Avi-Robusta:
If prometheus had multiple jobs or a job not named kubelet the graphs from robusta would be empty since the job is specified in the query.
Now we get all the graphs and check for the relevant job metric graph. (we prefer kubelet if it exists)
Additionally `container_cpu_usage_seconds_total:sum_irate` doesn't always exist in every prometheus so it is replaced too.

To duplicate this issue add this to generated_values.yaml

```
kube-prometheus-stack:
  prometheus:
      prometheusSpec:
        additionalScrapeConfigs:
          - job_name: 'robusta-kubernetes-cadvisor'
            scheme: https
            metrics_path: /metrics/cadvisor
            tls_config:
              ca_file: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
            bearer_token_file: /var/run/secrets/kubernetes.io/serviceaccount/token
            kubernetes_sd_configs:
            - role: node
            relabel_configs:
            - action: labelmap
              regex: __meta_kubernetes_node_label_(.+)
```

### DIFF
--- a/src/robusta/core/playbooks/prometheus_enrichment_utils.py
+++ b/src/robusta/core/playbooks/prometheus_enrichment_utils.py
@@ -80,12 +80,9 @@ def get_resolution_from_duration(duration: timedelta) -> int:
 
 
 def get_target_name(series: PrometheusSeries) -> Optional[str]:
-    if series.metric.get("container"):
-        return series.metric["container"]
-    elif series.metric.get("pod"):
-        return series.metric["pod"]
-    elif series.metric.get("node"):
-        return series.metric["node"]
+    for label in ["container", "pod", "node"]:
+        if label in series.metric:
+            return series.metric[label]
     return None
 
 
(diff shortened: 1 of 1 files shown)


Avi-Robusta:
@aantn

RoiGlinik (approved):

src/robusta/core/playbooks/prometheus_enrichment_utils.py
```diff
@@ -78,6 +79,39 @@ def get_resolution_from_duration(duration: timedelta) -> int:
...
+
+def get_series_job(series: PrometheusSeries) -> Optional[str]:
+    return series.metric.get("job")
+
+
+def filter_prom_jobs_results(series_list_result: Optional[List[PrometheusSeries]]) -> Optional[List[PrometheusSeries]]:
+    if not series_list_result or len(series_list_result) == 1:
+        return series_list_result
+
+    target_names = set([get_target_name(series) for series in series_list_result if get_target_name(series)])
```
LeaveMyYard, line 100: There is a set comprehension (does not have `:`, like a dict comprehension):
```python
{get_target_name(series) for series in series_list_result if get_target_name(series)}
```

Issue #841, opened 2023-04-26
MS Teams webhook returning status 400


mxk6n:
## Describe the bug
My MS Teams webhook suddenly stopped working recently.

When checking the logs of the robusta runner I got the following error:

```
ERROR    Failed to send message to teams. error: Webhook message delivery failed with error: Microsoft Teams endpoint returned HTTP error 400 with ContextId MS-CV=<CONTEXT_ID> message: <MESSAGE_DATA>
```

I then took `<MESSAGE_DATA>` and tried to post it to the webhook manually to see what the problem might be:

```
curl -X POST -H "Content-Type: application/json" -d @./message-data.json <WEBHOOK_URL>
```

which basically resulted in the same error as in the logs above (no further info).

First I thought it might be the message limit for MS Teams webhooks (28kB) that could be the problem, but it wasn't.

In the end it seemed like the severity emoji in the title ("🔴") was the problem:

```
{
  "type": "message",
  "attachments": [
    {
      "contentType": "application/vnd.microsoft.card.adaptive",
      "contentUrl": null,
      "content": {
        "$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
        "type": "AdaptiveCard",
        "version": "1.2",
        "msTeams": {
          "width": "full"
        },
        "body": [
          {
            "type": "TextBlock",
            "text": "🔴 HIGH - Job <JOB_NAME> in namespace <NAMESPACE> failed",
            "size": "extraLarge",
            "wrap": true
          }
        ]
      }
    }
  ]
}
```

did not work, while the following went through without any problems:

```
{
  "type": "message",
  "attachments": [
    {
      "contentType": "application/vnd.microsoft.card.adaptive",
      "contentUrl": null,
      "content": {
        "$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
        "type": "AdaptiveCard",
        "version": "1.2",
        "msTeams": {
          "width": "full"
        },
        "body": [
          {
            "type": "TextBlock",
            "text": "HIGH - Job <JOB_NAME> in namespace <NAMESPACE> failed",
            "size": "extraLarge",
            "wrap": true
          }
        ]
      }
    }
  ]
}
```

Does anyone have an idea on how to fix this or knows a workaround for this issue?

Thanks in advance 🙏 

## Robusta Config

```
customPlaybooks:
- triggers:
  - on_job_failure: {}
      # namespace_prefix: robusta
  actions:
    - create_finding:
[...]

PR #844, opened 2023-04-26
Fixed the error 400 being thrown for when emoji are sent via msteams …

ganeshrvel:
…api.

### DIFF
--- a/src/robusta/cli/main.py
+++ b/src/robusta/cli/main.py
@@ -172,7 +172,7 @@ def gen_config(
                 ms_teams_sink=MsTeamsSinkParams(
                     name="main_ms_teams_sink",
                     webhook_url=msteams_webhook,
-                    use_emoji=True
+                    use_emoji=False
                 )
             )
         )
--- a/src/robusta/cli/main.py
+++ b/src/robusta/cli/main.py
@@ -172,6 +172,7 @@ def gen_config(
                 ms_teams_sink=MsTeamsSinkParams(
                     name="main_ms_teams_sink",
                     webhook_url=msteams_webhook,
+                    use_emoji=True
                 )
             )
         )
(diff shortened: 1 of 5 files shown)


src/robusta/cli/main.py
```diff
@@ -172,6 +172,7 @@ def gen_config(
                 ms_teams_sink=MsTeamsSinkParams(
                     name="main_ms_teams_sink",
                     webhook_url=msteams_webhook,
+                    use_emoji=False
```
arikalon1, line 175: I think this isn't required if you put default value on the pydantic class
Also, this will only effect new clients.
We want it to work to all existing clients

src/robusta/core/sinks/msteams/msteams_sink.py
```diff
@@ -8,8 +8,9 @@ class MsTeamsSink(SinkBase):
     def __init__(self, sink_config: MsTeamsSinkConfigWrapper, registry):
         super().__init__(sink_config.ms_teams_sink, registry)
         self.webhook_url = sink_config.ms_teams_sink.webhook_url
+        self.use_emoji = False if (sink_config.ms_teams_sink.use_emoji is None) else sink_config.ms_teams_sink.use_emoji
```
arikalon1, line 11: This could be simplified to:
`self.use_emoji = sink_config.ms_teams_sink.use_emoji`
if the sink parameter has default value

src/robusta/core/sinks/msteams/msteams_sink_params.py
```diff
@@ -4,6 +4,7 @@
 
 class MsTeamsSinkParams(SinkBaseParams):
     webhook_url: str
+    use_emoji: bool
```
arikalon1, line 7: I would add a default value:
`use_emoji: bool = False`
Someone can override, otherwise it's False

arikalon1 (approved):
very nice
left some minor comments

ganeshrvel:
closing this as https://github.com/robusta-dev/robusta/issues/841 was fixed by microsoft itself

PR #852, opened 2023-05-07
Hikaru + Kubernetes client full update

Avi-Robusta:
Hikaru and the kubernetes client package were updated 
Preliminary testing showed no issues.
Advanced testing needed before final merge (In progress...)

Major internal changes:
- New events api/Object is being used, referencing the old object fields in the code might cause issues
  - new K8s api /apis/events.k8s.io/v1/namespaces/{namespace}/events
  - old K8s api /api/v1/namespaces/{namespace}/events

Notes:
- patch.py file was checked and obsolete patches were removed


arikalon1 (approved):
Nice  work!

Need to document it properly on the release notes, because some of the changes here aren't backward compatible

PR #854, opened 2023-05-08
Code for prometheus authorization with azure

RoiGlinik:
azure specific authorization code

### DIFF
--- a/src/robusta/integrations/prometheus/utils.py
+++ b/src/robusta/integrations/prometheus/utils.py
@@ -1,6 +1,8 @@
 import logging
-from typing import TYPE_CHECKING, List, Optional, Dict
+import os
+from typing import TYPE_CHECKING, Dict, List, Optional
 
+import requests
 from cachetools import TTLCache
 from requests.exceptions import ConnectionError, HTTPError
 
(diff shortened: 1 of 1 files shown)


src/robusta/integrations/prometheus/utils.py
```diff
@@ -9,6 +11,53 @@
...
+                    data={
+                        "grant_type": "client_credentials",
+                        "client_id": os.environ.get("AZURE_CLIENT_ID"),
+                        "client_secret": os.environ.get("AZURE_CLIENT_SECRET"),
+                        "resource": "https://prometheus.monitor.azure.com",
+                    },
+                )
+            except Exception:
+                logging.exception("Unexpected error when trying to generate azure access token.")
+                return
```
arikalon1, line 48: Can you explicitly return False?

src/robusta/integrations/prometheus/utils.py
```diff
@@ -9,6 +11,53 @@
 from robusta.core.model.env_vars import PROMETHEUS_SSL_ENABLED, SERVICE_CACHE_TTL_SEC
 from robusta.utils.service_discovery import find_service_url
 
+
+class PrometheusAuthorization:
+    bearer_token: str = ""
+    azure_authorization: bool = (
+        os.environ.get("AZURE_CLIENT_ID", "")
```
arikalon1, line 18: `os.environ.get("AZURE_CLIENT_ID", "")` appears twice

src/robusta/integrations/prometheus/utils.py
```diff
@@ -43,6 +89,13 @@ def check_prometheus_connection(prom: "PrometheusConnect", params: dict = None):
             # This query should return empty results, but is correct
             params={"query": "example", **params},
         )
+
+        if response.status_code == 401:
```
arikalon1, line 93: I'm a little confused by this section.
We had a failure due to authorization 

We have to options:
1. Azure auth
2. Other auth

On both I think we'd like to raise an error, right? (just return nothing hides the error)

I would think you should just run `PrometheusAuthorization.request_new_token()`, then raise_for_status anyway

Regardless, this implementation suggests we'll have a Prometheus error every time the token expires, which isn't the best, right?
Can't we refresh the token before it expires?

src/robusta/integrations/prometheus/utils.py
```diff
@@ -9,6 +11,53 @@
...
+            return {"Authorization": (f"Bearer {cls.bearer_token}")}
+        else:
+            return {}
+
+    @classmethod
+    def request_new_token(cls) -> bool:
+        if cls.azure_authorization:
+            try:
+                res = requests.post(
+                    url=f"https://login.microsoftonline.com/{os.environ.get('AZURE_TENANT_ID')}/oauth2/token",
```
arikalon1, line 37: Might be better to allow override for the url and resource, incase they change it and we want a quick workaround (via env var probably)
Just a nice to have

arikalon1 (approved):
Looks good
Left some comments

PR #865, opened 2023-05-12
Add instructions on EBS CSI driver. 

aantn:
Also move openshift to later in the list.

cc @arikalon1


arikalon1 (approved):
LGTM

I think installing on OpenShift also often needs to be configured the it's managed Prometheus, that often need to configure an authentication token.

I would consider adding it later on.

PR #866, opened 2023-05-12
Show how to link argocd to robusta

aantn:
@pavangudiwada can you test and review?

cc @arikalon1


pavangudiwada (changes requested):
Modified some details and tested it. Works fine. @aantn Take a look and lmk if other changes are needed. 

image

PR #872, opened 2023-05-16
Update playbook-job-failure.rst

aantn:
@pavangudiwada is the image here wrong? I think its an image for a builtin Prometheus alert about stuck Jobs, not Job failure.


aantn:
@pavangudiwada note that I also moved the image up top, to give motivation for trying the tutorial.

PR #876, opened 2023-05-16
Allow setting custom values for liveness and startup probes

svloginov:
We have a big-ish cluster and robusta-runner was restarting all the time, because it was failing to answer the health-check.
After I manually edited the deployment (added bigger timeout, set thresholds, etc), the issue disappeared.


svloginov:
Hello [arikalon1](https://github.com/arikalon1),
Yes, we have updated to 0.10.16, saw runner dying several times, added more memory, saw it dying some more, and then I ended up changing the deployment manually to set more relaxed parameters for the liveness check.

svloginov:
> Thanks @svloginov ,can you share what configuration you used for the `startupProbe` to solve it?

Funnily, I can't :) Before the investigation started properly, the first thing I've done was to rollback to 0.10.15 and I haven't updated back yet.

The problem for me was in the liveness anyway.

arikalon1 (approved):
Thanks for you PR @svloginov

PR #878, opened 2023-05-16
Extend OpenShift SCC support

jmcshane:
While the current installation documentation works to install the product in an OpenShift cluster, on one hand `anyuid` is far too broad for the standard product install and on the other hand key privileged playbooks fail on the cluster even when granted `anyuid`. The right way to tackle this problem is to provide a separation between the standard product operation mode and the privileged debugging mode.

One current challenge is that the `runner` project spins up the debug pod using its own service account. If this was split into a `runner-debug` service account, the Kubernetes privileges could be removed from `runner-debug` while the privileged host capabilities could be limited to this serviceaccount that only runs pods for debug purposes. See the `create_debugger_pod` method for the current request that is being sent to the API server: https://github.com/robusta-dev/robusta/blob/master/src/robusta/integrations/kubernetes/custom_models.py#L215.

With these two SCCs defined in the PR, users can install the current product without privileged access using the baseline instructions. Then, if debugger features are desired, the privileged mode can be enabled.

I see that the generated values file is created by your `robusta gen-config` CLI. This set of inputs could be added via an `--(no-)enable-openshift` option set.

### DIFF
--- a/docs/setup-robusta/supported-clusters.rst
+++ b/docs/setup-robusta/supported-clusters.rst
@@ -11,15 +11,68 @@ Distribution-specific instructions are below.
 OpenShift
 ========================================
 
-OpenShift is supported. Robusta needs to be granted permissions to run:
+OpenShift is supported via the Helm chart. 
 
-.. code-block:: bash
-   :name: cb-oc-adm-policy-add
+.. warning::
 
-    oc adm policy add-scc-to-user anyuid -z robusta-forwarder-service-account
-    oc adm policy add-scc-to-user anyuid -z robusta-runner-service-account
+      Do **not** install Robusta in the ``default`` OpenShift project.
 
-The above is very permissive. More restrictive policies can be applied.
+Update the ``generated_values.yaml`` file to enable OpenShift support:
+
+.. code-block:: yaml
+
+    openshift:
+      enabled: true
+      createScc: true
+    
+
+For users that do not have access to create their own SCCs in the cluster, see the
+`openshift-scc-baseline.yaml <../../helm/robusta/templates/openshift-scc-baseline.yaml>`_ file for the baseline SCC for the product. This SCC can be
+created in advance by administrators and referenced using the following set of values:
+
+.. code-block:: yaml
+
+    openshift:
+      enabled: true
+      sccName: my-admins-scc
+
+A test installation in OpenShift can use the existing SCC ``anyuid``.
+
+OpenShift Prometheus Stack
+--------------------------
+
+OpenShift comes with an existing Prometheus installation. In order to point
+to the preconfigued stack, update the ``globalConfig`` in the Helm values with
+the following:
+
+.. code-block:: yaml
+
+    globalConfig:
+      prometheus_url: http://prometheus-operated.openshift-monitoring.svc.cluster.local:9090
+
+Robusta Advanced Debug Features
+--------------------------------
+
+In order to support the ``python_debugger``, ``java_debugger`` and ``node_disk_analyzer``
+playbooks, permission to run a far more privileged container needs to be granted to
+the ``runner`` service account. This container has ``SYS_ADMIN`` capabilities and must
+run as root on the node.
+
+To support these features in a production environment, you may want to only temporarily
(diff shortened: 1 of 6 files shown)


jmcshane:
To test this PR in AWS:

1. Download the `oc` and `openshift-install` binaries from https://github.com/okd-project/okd/releases/tag/4.12.0-0.okd-2023-04-16-041331
2. Set up an AWS CLI profile
3. Configure a route53 hosted zone for the API Server endpoint with appropriate NS delegation before running the installation
4. Run `openshift-install create cluster` and follow the prompts (the install process takes between 30-40 minutes)
5. Run `robusta gen-config --no-enable-prometheus-stack`
6. Prepare the values file for openshift and installation from the repository helm chart (instead of the release version). We will first test the unprivileged mode by adding the following values to the `generated–values.yaml` file:

```yaml
runner:
  image: us-central1-docker.pkg.dev/genuine-flight-317411/devel/robusta-runner:0.10.16

openshift:
  enabled: true
  createScc: true

globalConfig:
  prometheus_url: http://prometheus-operated.openshift-monitoring.svc.cluster.local:9090
```

7. Prepare the namespace for the helm chart install `oc new-project robusta`
8. Clone this branch, from the root of the repository run `helm install robusta ./helm/robusta -f generated_values.yaml --set clusterName=my-test-openshift-cluster`
9. Validate basic cluster playbooks, I used the following to ensure API server actions were still operational:

```bash
robusta playbooks trigger node_running_pods_enricher name=<node-name>
robusta playbooks trigger node_allocatable_resources_enricher name=<node-name>
```

10. Test that persistent volume storage read/write and grafana renderer image continue to work after enforing fsgroup restrictions by updating the helm values:

```yaml
grafanaRenderer:
  enableContainer: true
playbooksPersistentVolume: true

```

11. Ensure you can write playbooks to the persistent volume and that they persist after restart as described in [the docs](https://docs.robusta.dev/master/developer-guide/actions/playbook-repositories.html#loading-your-own-playbook-repository)
12. Test the privileged mode that enables the debugger pod to be installed in the namespace by updating the values file:

```yaml
openshift:
  createPrivilegedScc: true
```

13. Update the helm install: `helm upgrade robusta ./helm/robusta -f generated_values.yaml --set clusterName=my-test-openshift-cluster`
14. Test the debug workflows using a python debugger on a python sample pod

```
oc run python --image octopusdeploy/sample-python-webapp:0.0.1
robusta playbooks trigger python_debugger name=python namespace=robusta
```

15. Delete the OpenShift cluster with `openshift-install destroy cluster`

jmcshane:
Having worked through this whole workflow, I see there is an error with the python debug script. I see that you are setting up `gdb` to run on the host, but all I could confirm was that `debug-toolkit` started: `/usr/local/bin/python3.8 /usr/local/bin/debug-toolkit debugger 24815 --port 5678`. The error is

```
2023-05-17 14:38:50.686 ERROR    could not parse json:
----------------
ERROR! Timed out after 120 seconds

  File "/usr/local/bin/debug-toolkit", line 6, in <module>
    sys.exit(app())

  File "/app/src/debug_toolkit/main.py", line 213, in debugger
    inject_string(pid, payload, trampoline=True, trampoline_timeout=timeout_seconds, verbose=verbose)

  File "/app/src/debug_toolkit/main.py", line 167, in inject_string
    return do_trampoline_injection(pid, payload, verbose, trampoline_timeout)

  File "/app/src/debug_toolkit/main.py", line 137, in do_trampoline_injection
    with open(abs_done_path, "r") as f:

FileNotFoundError: [Errno 2] No such file or directory: '/proc/62042/cwd/2100666c6179b700d23594c0a83ad51f_done.txt'
```

The pid `62042` is the pid of the python pod from step 14 above. I confirmed that I would be able to write that file myself in the debug-toolkit pod by running `exec` into the debug pod. It may be a specific combination of the versions I'm using here, but I'll turn this over to your team for now.

The pod is created successfully, so the SCC restrictions are correctly applied based on the pod spec you provide.

arikalon1 (approved):
Thank you for this PR @jmcshane 
This is excellent!

I'll followup internally about the debugger pod issues.

PR #881, opened 2023-05-17
format pdf output units for krr

Avi-Robusta:
[Krr report.pdf](https://github.com/robusta-dev/robusta/files/11499079/Krr.report.pdf)
used to be just bytes, now its formatted just like the table and UI

### DIFF
--- a/src/robusta/core/model/pods.py
+++ b/src/robusta/core/model/pods.py
@@ -1,6 +1,6 @@
 import logging
 from enum import Enum
-from typing import Dict, List, Optional
+from typing import Dict, List, Literal, Optional, Union
 
 from hikaru.model import Container, ContainerState, ContainerStatus, Pod
 from pydantic import BaseModel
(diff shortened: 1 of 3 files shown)


src/robusta/core/model/pods.py
```diff
@@ -24,6 +24,23 @@
     "Ei": 1024 * 1024 * 1024 * 1024 * 1024,
 }
 
+
+def format_unit(x: Union[float, int], /, *, base: Literal[1024, 1000] = 1024) -> str:
```
arikalon1, line 28: Should this support value conversion for cpu and memory?
If so, I think there's an issue here.
For value `1.5` I get `1`

The other think is a matter of preference:
I think 
2441Ki should be 2.4Mi
2384Mi should be 2.3Gi

@Avi-Robusta @LeaveMyYard thoughts?

arikalon1 (commented):
I think there's a small bug here

arikalon1 (approved):
LLGTM

PR #884, opened 2023-05-18
Minor improvements to Azure configuration docs

aantn:
cc @arikalon1 

Let me know if I missed any other improvements we should make.


RoiGlinik:
@aantn I will merge for now

PR #885, opened 2023-05-18
Added prometheus params from robusta to KRR args

Avi-Robusta:


playbooks/robusta_playbooks/krr.py
```diff
@@ -140,14 +138,14 @@ def krr_scan(event: ExecutionBaseEvent, params: KRRParams):
...
-
+    header = PrometheusAuthorization.get_authorization_headers(params)["Authorization"]
     spec = PodSpec(
         serviceAccountName=params.serviceAccountName,
         containers=[
             Container(
                 name=to_kubernetes_name(IMAGE),
                 image=IMAGE,
-                command=["/bin/sh", "-c", f"python krr.py {params.strategy} {params.args_sanitized} -q -f json"],
+                command=["/bin/sh", "-c", f"python krr.py {params.strategy} {params.args_sanitized} -p {params.prometheus_url} --prometheus-auth-header \"{header}\" -q -f json"],
```
arikalon1, line 148: The command is part of the job spec.
I think we need to pass the `auth-header` in a secure manner.

arikalon1 (commented):
Nice work.
Left one comment

playbooks/robusta_playbooks/krr.py
```diff
@@ -141,13 +143,42 @@ def krr_scan(event: ExecutionBaseEvent, params: KRRParams):
...
+
+    python_command = f"python krr.py {params.strategy} {params.args_sanitized} -p {params.prometheus_url} -q -f json "
+
+    auth_header = headers["Authorization"] if "Authorization" in headers else ""
+
+    # adding env var of auth token from Secret
+    env_var = None
+    secret = None
+    if auth_header:
+        krr_secret_name = "krr-auth-secret"
```
arikalon1, line 156: I think this may cause issues if there are multiple krr scans in parallel 
The secret creation will fail, or deleted before used etc.

Can we do the following:
For each krr scan create a unique job name
Create the job (where the container is dependent on the secret)
Create the secret with the job as owner.

When the job is done and deleted, the secret will be deleted with it.
The container will not start until the secret is created.
This also removes the need for secret read/patch permission. Only write is required. 
WDYT?

Avi-Robusta: when i tested it I kept getting a CreateContainerConfigError and the pod wouldnt run but I think I found the issue so ill try it

Avi-Robusta: but then you have the issue if the secret isnt sucessfully created than the job will exist forever

arikalon1: I tested it manually. The container had indeed the `CreateContainerConfigError` error, but once I created the secret it ran and finished successfully

Avi-Robusta: The GC didnt delete the secret when the Job was the owner but did delete it when the Job Pod was the owner

playbooks/robusta_playbooks/krr.py
```diff
@@ -25,22 +27,25 @@
     to_kubernetes_name,
 )
 
-IMAGE: str = os.getenv("KRR_IMAGE_OVERRIDE", "leavemyyard/robusta-krr:latest")
+IMAGE: str = os.getenv("KRR_IMAGE_OVERRIDE", "us-central1-docker.pkg.dev/genuine-flight-317411/devel/krr:test")
```
Avi-Robusta, line 30: remove image change

arikalon1, line 30: Should we remove this?

playbooks/robusta_playbooks/krr.py
```diff
@@ -209,6 +251,8 @@ def krr_scan(event: ExecutionBaseEvent, params: KRRParams):
                             "request": scan.recommended.requests[resource].priority,
                             "limit": scan.recommended.limits[resource].priority,
                         },
+                        "metric": scan.metrics[resource].dict(),
```
arikalon1, line 254: There will always be metrics for every resource? (no risk of null pointer error)

src/robusta/integrations/kubernetes/custom_models.py
```diff
@@ -390,8 +395,30 @@ def get_single_pod(self) -> RobustaPod:
             raise Exception(f"got more pods than expected for job: {pods}")
         return pods[0]
 
+    def create_job_owned_secret(self, secret: JobSecret):
+        """
+            This secret will be auto-deleted when the pod is Terminated
+        """
+        # Due to inconsistant GC in K8s the OwnerReference needs to be the pod and not the job (Found in azure)
+        job_pod = self.get_single_pod()
+        robusta_owner_reference = OwnerReference(apiVersion="v1",
```
arikalon1, line 404: Why do we use the pod as owner reference and not the job?
If the pod crashes, the secret will be deleted, right?

src/robusta/integrations/kubernetes/custom_models.py
```diff
@@ -390,8 +395,30 @@ def get_single_pod(self) -> RobustaPod:
...
+                                                 blockOwnerDeletion=False,
+                                                 controller=True)
+        secret = Secret(
+            metadata=ObjectMeta(name=secret.name, ownerReferences=[robusta_owner_reference]),
+            data=secret.data
+        )
+        try:
+            return secret.createNamespacedSecret(job_pod.metadata.namespace).obj
+        except Exception as e:
+            logging.error(f"Failed to create secret {secret.name}", exc_info=True)
```
arikalon1, line 417: Secret here is the k8s secret object, it doesn't have name
I think you have ambiguous names

arikalon1 (commented):
Looks good
left some comments

playbooks/robusta_playbooks/krr.py
```diff
@@ -52,10 +57,18 @@ class KRRRecommended(BaseModel):
     limits: Dict[str, KRRRecommendedInfo]
 
 
+class KRRMetric(BaseModel):
+    query: str
```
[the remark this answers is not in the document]
Avi-Robusta: something is off with git, this part was added by rebase, not by me

playbooks/robusta_playbooks/krr.py
```diff
@@ -140,14 +154,42 @@ def krr_scan(event: ExecutionBaseEvent, params: KRRParams):
...
+        # setting env variables of krr to have secret
+        env_var = [
+                  EnvVar(
+                      name=env_var_auth_name,
+                      valueFrom=EnvVarSource(secretKeyRef=SecretKeySelector(name=krr_secret_name,
+                                                                            key=prometheus_auth_secret_key)),
+                  )
+              ]
+        # adding secret env var in krr pod command
+        python_command += f'--prometheus-auth-header \"${env_var_auth_name}\"'
```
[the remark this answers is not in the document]
Avi-Robusta: from within krr?

playbooks/robusta_playbooks/krr.py
Avi-Robusta: also this is from rebase, not added by me

src/robusta/integrations/kubernetes/custom_models.py
Avi-Robusta: yes the secret will be deleted. 
in the comment above I explained that when the job is the owner the GC doesn't clean up the secret at least in azure.
I am unsure why this behavior happens in k8s but due to being short on time I put off investigating in depth.

src/robusta/integrations/kubernetes/custom_models.py
Avi-Robusta: youre right

playbooks/robusta_playbooks/krr.py
Avi-Robusta: done

arikalon1 (approved):

Avi-Robusta:
Saw bug in default master KRR v1.2.0 so i couldn't test one last time before vacation. Will test with working krr version and merge when I return.
To test

- [ ]  a cluster with no auth
- [ ]  a cluster with auth - ask roi about his azure cluster

arikalon1 (approved):
LGTM

PR #894, opened 2023-05-19
Make it dreamy to run robusta-runner locally

aantn:
@robusta-dev ❤️ @metalbear-co


run_runner_locally.sh
```diff
@@ -6,7 +6,10 @@ set -e
 
 # Settings you might want to change
 export PYTHON_BINARY=python3.9
-export PORT=5000
+export PORT=6000  # if you change this, update port_mapping in mirrord.json too
```
arikalon1, line 9: Was there a reason for changing the port?

run_runner_locally.sh
```diff
@@ -6,7 +6,10 @@ set -e
...
+export LOG_LEVEL=INFO
+#export TRACE_INCOMING_REQUESTS=false
```
arikalon1, line 11: can we just delete?

run_runner_locally.sh
```diff
@@ -58,15 +61,13 @@ if [ ! -f "$PWD/deployment/playbooks/active_playbooks.yaml" ]; then
 fi
 
 echo "Installing builtin playbooks"
-poetry run python3 -m pip install -e ./deployment/playbooks/defaults
+#poetry run python3 -m pip install -e ./deployment/playbooks/defaults
```
arikalon1, line 64: Why did you remove the local playbooks installation? Isn't that required for dev?

run_runner_locally.sh
```diff
@@ -58,15 +61,13 @@ if [ ! -f "$PWD/deployment/playbooks/active_playbooks.yaml" ]; then
...
 export PLAYBOOKS_ROOT=./deployment/playbooks
 export REPO_LOCAL_BASE_DIR=./deployment/git_playbooks
 export INSTALLATION_NAMESPACE=default
-export TRACE_INCOMING_REQUESTS=true
-export LOG_LEVEL=DEBUG
-
-poetry run python3 -m robusta.runner.main
 
+mirrord exec -f mirrord.json -- poetry run python3 -m robusta.runner.main
+#poetry run python3 -m robusta.runner.main
```
arikalon1, line 73: can we just delete?

arikalon1 (approved):
Looks good
Left some minor comments.
Feel free to fix and merge

run_runner_locally.sh
aantn: That port is already taken on my laptop, I think it happened to someone else too. If you don't mind, I'd like to jjust change the default.

run_runner_locally.sh
aantn: I want to leave it here so people see and know they can change it easily.

run_runner_locally.sh
aantn: Whoops, fixed.

run_runner_locally.sh
aantn: Yes, done.

PR #895, opened 2023-05-19
fix limits in helm template runner

daanvinken:


arikalon1 (approved):
Thanks for the PR @daanvinken 
I'm going to add a small fix on top of it, to make it backward compatible to those who doesn't set memory limits

PR #896, opened 2023-05-19
Make Mattermost client work with non-admin bots and fix logo url

daanvinken:
Discussed with Arik [here](https://robustacommunity.slack.com/archives/C02R0LSAHNJ/p1681656740413749) to maintain backwards compatibility.

This did feel a bit like writing spaghetti code due to that requirement. Any feedback / improvements is most welcome.
Both flows have been tested 👍

### DIFF
--- a/docs/configuration/sinks/mattermost.rst
+++ b/docs/configuration/sinks/mattermost.rst
@@ -39,6 +39,11 @@ Getting your mattermost webhook url
       :width: 600
       :align: center
 
+.. note::
+
+    If you are not able to use an admin bot, make sure to include `team_id` in your sink configuration.
+    Also, be aware that you won't be able to post to private channels with a non-admin bot.
+
 4. Copy the token value of the bot - it will be used to send all the messages to the channel.
 
     .. image:: /images/add_mattermost_bot_5.png
@@ -67,6 +72,7 @@ Now we're ready to configure the Mattermost sink.
             token: <YOUR BOT TOKEN> (the token we copied the first after bot creation)
             token_id: <YOUR BOT TOKEN ID> (the token id visible in bot panel)
             channel: <YOUR CHANNEL NAME> (the channel name you want to send messages to - either display name or channel name divided by hyphen (e.g. channel-name))
+            team_id: <YOUR TEAM ID> (OPTIONAL - this is only needed if your mattermost bot is not an admin)
 
 Save the file and run
 
@@ -75,7 +81,7 @@ Save the file and run
 
     helm upgrade robusta robusta/robusta --values=generated_values.yaml
 
-You should now get playbooks results in Mattermost!
+You should now get playbooks results in Mattermost! Make sure to add the bot to the required channels.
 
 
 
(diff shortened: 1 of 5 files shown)


docs/configuration/sinks/mattermost.rst
```diff
@@ -75,7 +81,7 @@ Save the file and run
 
     helm upgrade robusta robusta/robusta --values=generated_values.yaml
 
-You should now get playbooks results in Mattermost!
+You should now get playbooks results in Mattermost! Make sure to add the bot to the required channels.
```
arikalon1, line 84: `Make sure to add the bot to the required channels.` is only if you use non-admin bot?

src/robusta/integrations/mattermost/client.py
```diff
@@ -39,6 +40,24 @@ def get_token_owner_id(self) -> Optional[str]:
         response_data = response.json()
         return response_data.get("user_id")
 
+    def is_admin_bot(self):
+        endpoint = "/users/me"
+        url = self._get_full_mattermost_url(endpoint)
+        response = self._send_mattermost_request(url, HttpMethod.GET)
+        if not check_response_succeed(response):
+            logging.error("Could not connect to Mattermost with bot account")
```
arikalon1, line 48: Should we return here?
Will we have data in the response in case of error?

src/robusta/integrations/mattermost/client.py
```diff
@@ -58,21 +77,31 @@ def update_bot_logo(self, bot_id: str):
             logging.warning("Cannot update bot logo, probably bot has not enough permissions")
 
     def _init_setup(self, channel_name: str, team_name: Optional[str] = None):
```
arikalon1, line 79: Will this be easier to read? I think it the same logic
```
    def _init_setup(self, channel_name: str, team_name: Optional[str] = None):
        if self.is_admin:
            self.bot_id = self.get_token_owner_id()
            self.team_id = self.get_team_id(team_name) if team_name else None
            if self.bot_id:
                self.update_bot_settings(self.bot_id)

        self.channel_id = self.get_channel_id(channel_name)
        if not self.channel_id:
            logging.warning("No channel found, messages won't be sent")
```

arikalon1 (commented):
Thanks for the PR @daanvinken 
Left few minor comments.
WDYT?

docs/configuration/sinks/mattermost.rst
daanvinken: Ah yeah it is not required for messages, but if you have a non-admin bot you need the bot to be added in order to post files. I'll update accordingly

src/robusta/integrations/mattermost/client.py
daanvinken: Good point, do you think we should also check if we return `None` and early-exit the constructor?

daanvinken: @arikalon1 Hoping to get this into the upcoming release 😎

arikalon1: I think we should just assume we're not admin.
If it can work in a non-admin mode, no reason not to.

PR #910, opened 2023-05-27
Improve alert manager grouping configuration

arikalon1:
Fix matchers in the docs to match any severity (instead of a predefined list)


aantn (commented):
LGTM

aantn (approved):

PR #911, opened 2023-05-28
Msteams improved tables

RoiGlinik:
new table example
fixed time render
fixed rows height with using table blocks
image

old table example
image


arikalon1 (approved):
LGTM

PR #917, opened 2023-05-31
Add support for Azure Managed Identities

### LINKED ISSUE #848: Support for Azure managed Prometheus
**Is your feature request related to a problem?**
I use Azure Managed Prometheus that doesn't allow to change alertmanager config

**Describe the solution you'd like**
It would be super neat to be able to call robusta webhook from Azure Monitor (as Action Group Id or something)

**Describe alternatives you've considered**
Installing Prometheus in the cluster but it's redundant. An Azure function that translates the azure alert into a robusta webhook is another alternative 

**Additional context**
https://learn.microsoft.com/en-us/azure/azure-monitor/essentials/prometheus-rule-groups#create-prometheus-rules

EppO:
Along with Service Principal Authentication, here is the support for Managed Identities. Implementation is sending a request to the Azure Instance Metadata Service (IMDS) to get the access token. No client secret is required. 
Managed Identity still needs Monitoring Data Reader role on the Azure Monitor Workspace to be able to query the prometheus metrics.

Other changes in the code were triggered by Black, Flake8 or PyRight.

closes #848

### DIFF
--- a/src/robusta/integrations/prometheus/utils.py
+++ b/src/robusta/integrations/prometheus/utils.py
@@ -17,6 +17,9 @@ from robusta.core.model.env_vars import PROMETHEUS_SSL_ENABLED, SERVICE_CACHE_TT
 from robusta.utils.service_discovery import find_service_url
 
 AZURE_RESOURCE = os.environ.get("AZURE_RESOURCE", "https://prometheus.monitor.azure.com")
+AZURE_METADATA_ENDPOINT = os.environ.get(
+    "AZURE_METADATA_ENDPOINT", "http://169.254.169.254/metadata/identity/oauth2/token"
+)
 AZURE_TOKEN_ENDPOINT = os.environ.get(
     "AZURE_TOKEN_ENDPOINT", f"https://login.microsoftonline.com/{os.environ.get('AZURE_TENANT_ID')}/oauth2/token"
 )
@@ -25,10 +28,8 @@ AZURE_TOKEN_ENDPOINT = os.environ.get(
 class PrometheusAuthorization:
     bearer_token: str = ""
     azure_authorization: bool = (
-        os.environ.get("AZURE_CLIENT_ID", "")
-        or os.environ.get("AZURE_TENANT_ID", "")
-        or os.environ.get("AZURE_CLIENT_SECRET", "")
-    )
+        os.environ.get("AZURE_CLIENT_ID", "") != "" and os.environ.get("AZURE_TENANT_ID", "") != ""
+    ) and (os.environ.get("AZURE_CLIENT_SECRET", "") != "" or os.environ.get("AZURE_USE_MANAGED_ID", "") != "")
 
     @classmethod
     def get_authorization_headers(cls, params: Optional[PrometheusParams] = None) -> Dict:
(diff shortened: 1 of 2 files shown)


src/robusta/integrations/prometheus/utils.py
```diff
@@ -226,8 +247,7 @@ def find_prometheus_url(cls) -> Optional[str]:
                 "app=prometheus-prometheus",
                 "app.kubernetes.io/name=vmsingle",
                 "app.kubernetes.io/name=victoria-metrics-single",
-                "app.kubernetes.io/name=vmselect"
-                "app=vmselect",
+                "app.kubernetes.io/name=vmselect" "app=vmselect",
```
EppO, line 250: is it correct? looks like there was a missing comma originally

src/robusta/integrations/prometheus/utils.py
```diff
@@ -21,9 +28,9 @@
 class PrometheusAuthorization:
     bearer_token: str = ""
     azure_authorization: bool = (
-        os.environ.get("AZURE_CLIENT_ID", "")
-        or os.environ.get("AZURE_TENANT_ID", "")
-        or os.environ.get("AZURE_CLIENT_SECRET", "")
+        os.environ.get("AZURE_CLIENT_ID", "") != ""
```
RoiGlinik, line 31: Can you add another parenthesis to separate between the two checks.

src/robusta/integrations/prometheus/utils.py
```diff
@@ -39,16 +46,30 @@ def get_authorization_headers(cls, params: Optional[PrometheusParams] = None) ->
...
-                    data={
-                        "grant_type": "client_credentials",
-                        "client_id": os.environ.get("AZURE_CLIENT_ID"),
-                        "client_secret": os.environ.get("AZURE_CLIENT_SECRET"),
-                        "resource": AZURE_RESOURCE,
-                    },
-                )
+                if os.environ.get("AZURE_USE_MANAGED_ID"):
+                    res = requests.post(
+                        url=AZURE_METADATA_ENDPOINT,
```
RoiGlinik, line 51: This request was not working for me . I had to use .
```
                    res = requests.get(
                        url=AZURE_METADATA_ENDPOINT,
                        headers={"Metadata": "true"},
                        params={
                            "api-version": "2018-02-01",
                            "client_id": os.environ.get("AZURE_CLIENT_ID"),
                            "resource": AZURE_RESOURCE,
                        },
```

EppO: you're right it's a GET request, not POST

EppO:
Discussed with @RoiGlinik on Slack, the Managed Identity binding is still TBD:

- using kubelet's Managed Identity
- using [Azure workload Identity](https://learn.microsoft.com/en-us/azure/aks/workload-identity-overview)

Creation of the alerts are also to be documented (another PR?), Azure doesn't provision the [alerts](https://learn.microsoft.com/en-us/azure/azure-monitor/containers/container-insights-metric-alerts?tabs=arm-template%2Cazure-portal#enable-prometheus-alert-rules), it's up to the user to create them. The Action group definition (webhook triggered by the alert pointing to Robusta endpoint) needs also to be documented.

EppO:
Workload identity use case will be implemented in a separate PR. docs regarding alerts creation and forwarding to Robusta will be improved in a separate PR as well.

PixelRobots:
I have some bicep code for creating the Azure alerts if it is helpful.

EppO:
> I have some bicep code for creating the Azure alerts if it is helpful. 

Once this PR is merged, I'll open a new one for the alert docs. You can add your Bicep code to the upcoming PR or create another one if you prefer. 
On my side I'll write a terraform example as I'm more fluent in TF code. 
Thanks for the input!

RoiGlinik (approved):

PR #922, opened 2023-06-02
Added the support for Query string in the Prometheus_url field

ganeshrvel:


sumeet-zuora:
Do we know when this would be released

sumeet-zuora:
This PR would really help our team to adopt Robusta, can we get some traction on this.. not sure what is the priority on this

pablos44 (changes requested):
Hey, it is much simpler and more flexible to have a separate parameter for query params than gluing them to `prometheus_url`. let's add `prometheus_url_query`  param for this purpose.

src/robusta/utils/common.py
```diff
@@ -37,3 +38,11 @@ def duplicate_without_fields(obj: HikaruBase, omitted_fields: List[str]):
             pass  # in case the field doesn't exist on this object
 
     return duplication
+
+
+def parse_query_string(query_string: str) -> Dict[str, str]:
+    if not query_string:
+        return {}
+    query_params = parse_qs(query_string)
+    return {key: value[0] for key, value in query_params.items()}
```
pablos44, line 47: It can happen that there is the same key repeated a few times in a query. So value[0] is not always sufficient

pablos44 (approved):

PR #924, opened 2023-06-04
Pytest improvements

aantn:


src/robusta/integrations/kubernetes/custom_models.py
```diff
@@ -358,6 +360,23 @@ def read(name: str, namespace: str) -> "RobustaPod":
...
         return Pod.readNamespacedPod(name, namespace).obj
 
+    @staticmethod
+    def wait_for_pod_ready(pod_name: str, namespace: str, timeout: int = 60) -> "RobustaPod":
+        """
+        Waits for the pod to be in Running state
+        """
+        for _ in range(timeout):  # retry for up to timeout seconds
+            try:
+                pod = RobustaPod().read(pod_name, namespace)
```
RoiGlinik, line 370: The () is probably unnecessary here for the static method, but ill merge with it.

RoiGlinik (approved):

src/robusta/integrations/kubernetes/custom_models.py
aantn: @RoiGlinik you need it. I tried without and had issues.

I think `read` is an instance method.

PR #926, opened 2023-06-05
New playbooks model

pablos44:
1. Playbooks are now organized as a map instead of a list. This allows to override of existing playbooks or disable them
2. Enrichments aren't shared anymore between playbooks. So playbook is responsible for handling the event, enriching it, and sending it to a sink. 
3. Platbooks now have priority which defines an order of execution. Additionally, a playbook can halt an event from being processed by other playbooks. This is used to implement silencing playbooks by setting them 0 priority and halting them


src/robusta/core/playbooks/playbooks_event_handler_impl.py
```diff
@@ -32,21 +32,18 @@ def handle_trigger(self, trigger_event: TriggerEvent) -> Optional[Dict[str, Any]
             return
 
         execution_response = None
-        execution_event: Optional[ExecutionBaseEvent] = None
```
arikalon1, line 35: Removing this leaves a lint warning in the code

src/robusta/core/playbooks/playbooks_event_handler_impl.py
```diff
@@ -32,21 +32,18 @@ def handle_trigger(self, trigger_event: TriggerEvent) -> Optional[Dict[str, Any]
...
             fired_trigger = self.__get_fired_trigger(trigger_event, playbook.triggers, playbook.get_id())
             if fired_trigger:
-                execution_event = None
+                
                 try:
                     execution_event = fired_trigger.build_execution_event(trigger_event, sink_findings)
-                    # sink_findings needs to be shared between playbooks.
-                    # build_execution_event returns a different instance because it's running in a child process
-                    execution_event.sink_findings = sink_findings
+                    logging.debug(f"playbook.name {playbook.name},build_execution_event {execution_event}")
```
arikalon1, line 43: I think this would be too spammy. Do we really need it?

src/robusta/core/playbooks/playbooks_event_handler_impl.py
```diff
@@ -32,21 +32,18 @@ def handle_trigger(self, trigger_event: TriggerEvent) -> Optional[Dict[str, Any]
...
                 except Exception:
                     logging.error(
-                        f"Failed to build execution event for {trigger_event.get_event_description()}, Event: {trigger_event}",
-                        exc_info=True,
+                        f"Failed to build execution event for {trigger_event.get_event_description()}, Event: {trigger_event}"
```
arikalon1, line 46: Why did you remove the logging of the stack trace?

src/robusta/runner/config_loader.py
```diff
@@ -245,26 +245,24 @@ def __prepare_runtime_config(
...
         else:
             logging.warning("No active playbooks configured")
 
-        playbooks_registry = PlaybooksRegistryImpl(
-            active_playbooks,
-            actions_registry,
-            runner_config.global_config,
-            sinks_registry.default_sinks,
-        )
+        try:
```
arikalon1, line 254: Is it better to just let the error be thrown here?
If we return undefined playbooks registry it will fail in many other places

src/robusta/model/playbook_definition.py
```diff
@@ -1,44 +1,40 @@
...
-    sinks: List[str] = None
+    name: str
+    sinks: List[str] = None  # compatibility
     triggers: List[Trigger]
-    actions: List[Dict[str, Any]]
-    _actions: List[PlaybookAction] = PrivateAttr()
-    stop: bool = False
-    _playbook_id: str = PrivateAttr()
+    actions: List[PlaybookAction] = []
+    tags: Dict[str, str] = {}
```
arikalon1, line 17: I think we don't use tags anywhere.
If that's the case, maybe it's better to remove it

src/robusta/model/playbook_definition.py
```diff
@@ -1,44 +1,40 @@
...
 
-    def __init__(self, **data: Any):
-        super().__init__(**data)
-        self._actions = []
-        for action in self.actions:
-            if len(action.keys()) != 1:
-                raise Exception(f"Action must have a single name: {action.keys()}")
+    # TODO: better to be in root_validator of the PlaybookAction
+    @validator('actions', pre=True, each_item=True)
+    def create_action(cls, action_data):
```
arikalon1, line 25: I think this is a little confusing.
We call it for each list item, but get the list each time, and assume the iterator is on the same position.

Would it be simpler to run once, and validate the whole list?

arikalon1 (commented):
Nice work!
Added some comments.
In addition, there's another piece of ugly code that can be removed. 
It handle managing finding per sink, which isn't required anymore  [here](https://github.com/robusta-dev/robusta/blob/4f3d9e03f7b8000772f60bf952b591110c251486/src/robusta/core/model/events.py#L95)
and [here](https://github.com/robusta-dev/robusta/blob/4f3d9e03f7b8000772f60bf952b591110c251486/src/robusta/core/model/events.py#L73)

We also need to fix the docs/add new docs that describes the change

helm/robusta/values.yaml
```diff
@@ -76,323 +76,496 @@ disableCloudRouting: false
...
+
+
+
+
+###################################
+# playbook map, playbooksMap
+###################################
+playbooksMap:
+  # priority builtin playbooks for running before all playbooks
+  #don't remove it. used for internal processing !!!  !
```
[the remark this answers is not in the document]
pablos44: I moved this playbook from code(config_loader.py). Removing it will influence normal code execution.
```
active_playbooks = [
            PlaybookDefinition(
                triggers=[{"on_kubernetes_any_resource_all_changes": {}}],
                actions=[{"cluster_discovery_updates": {}}],
            )
        ]
```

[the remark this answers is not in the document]
arikalon1: ohh got it.
Maybe the comment should be:
`This playbook required for Robusta to function properly. Do not disable or remove it.`

helm/robusta/values.yaml
```diff
@@ -76,323 +76,496 @@ disableCloudRouting: false
...
-  actions:
-  - popeye_scan: {}
-  sinks:
-    - "robusta_ui_sink"
+###################################
+# playbook map, platformPlaybooksMap
+###################################
+platformPlaybooksMap:
+  # name: k8sWarningEventCreateReportRobusta
+  k8sWarningEventCreateReportRobusta:
```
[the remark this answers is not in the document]
pablos44: The default is infinite. No reason for a halt. This runs at the end. Do you think differently?

helm/robusta/values.yaml
```diff
@@ -76,323 +76,496 @@ disableCloudRouting: false
...
+  k8sAnyResourceAllChangesDiscovery:
+    triggers:
+    - on_kubernetes_any_resource_all_changes: {}
+    actions:
+    - cluster_discovery_updates: {}
+    priority: 0
+
+  # playbooks for prometheus silencing
+  # name: prometheusAllAlertSilencing
+  prometheusAllAlertSilencing:
```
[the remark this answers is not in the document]
pablos44: Don't you like the comment itself ? the `# name: ....` pattern ? I use it for easy indexing. For instance, searching this pattern gives you a list of all names.

src/robusta/model/playbook_definition.py
pablos44: Not sure I got you. The each_item flag just saves me from writing an extra loop. The action data is a dict object

arikalon1:
not going to merge it right now.
Closing it for now

PR #930, opened 2023-06-07
Persistent events

RoiGlinik:


src/robusta/core/sinks/robusta/dal/supabase_dal.py
```diff
@@ -537,3 +548,29 @@ def publish_cluster_nodes(self, node_count: int):
...
+            event["account_id"] = self.account_id
+            event["cluster_id"] = self.cluster
+            event["namespace"] = block.namespace
+            if finding_id:
+                event["finding_id"] = str(finding_id)
+            event.setdefault("kind", block.kind)
+            event.setdefault("name", block.resource_name)
+            db_events.append(event)
+
+        res = self.client.table(RESOURCE_EVENTS).insert(db_events, upsert=True).execute()
```
Avi-Robusta, line 565: will upserting override the previous events row stored in the table?
like if we get the events for deployment A today and again tomorow will there only be the newer events?

Avi-Robusta, line 565: not sure upsert is needed since the primary key is defined as all the fields
so it would register any differences as a different row

src/robusta/core/sinks/robusta/dal/model_conversion.py
```diff
@@ -97,6 +97,8 @@ def to_evidence_json(
                 structured_data.append({"type": "list", "data": block.items})
             elif isinstance(block, PrometheusBlock):
                 structured_data.append({"type": "prometheus", "data": block.data.json(), "metadata": block.metadata})
+            elif isinstance(block, EventsBlock):
```
Avi-Robusta, line 100: will this prevent us from creating findings from any of the events actions since they all get the events from get_resource_events_table? can you verify we get events in other findings in the platform still?

Avi-Robusta (approved):
looks good, no big changes 
but verify that the findings in the ui that are supposed to get events still get them

src/robusta/core/sinks/robusta/dal/model_conversion.py
RoiGlinik: This will only prevent the creation of events evidence. 
Finding will be created and extra code from the UI will be added to fetch the events. 

There is an extra annotation to mute the finding but it has to be added specifically   

I will run some extra integration tests once the UI is ready as well

src/robusta/core/sinks/robusta/dal/supabase_dal.py
```diff
@@ -537,3 +548,29 @@ def publish_cluster_nodes(self, node_count: int):
...
+    def persist_events_block(self, block: EventsBlock, finding_id):
+        db_events = []
+        for row in block.render_rows():
+            event = dict(zip(block.headers, row))
+            event["account_id"] = self.account_id
+            event["cluster_id"] = self.cluster
+            event["namespace"] = block.namespace
+            if finding_id:
+                event["finding_id"] = str(finding_id)
+            event.setdefault("kind", block.kind)
```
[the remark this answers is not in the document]
RoiGlinik: There is an option to call events with related pods, in this case the events are already filled with this info.
So I only override it if its empty

src/robusta/core/sinks/robusta/dal/supabase_dal.py
RoiGlinik: Yes on different findings it wont help, the upsert should only happen on a no finding case ( with muted finding for on_kubernetes_warning_event_create )

src/robusta/core/sinks/robusta/dal/supabase_dal.py
```diff
@@ -537,3 +548,29 @@ def publish_cluster_nodes(self, node_count: int):
...
+            msg = f"Failed to persist resource events error: {res.get('data')}"
+            logging.error(msg)
+            self.handle_supabase_error()
+            raise Exception(msg)
+
+    def persist_platform_blocks(self, enrichment: Enrichment, finding_id):
+        for block in enrichment.blocks:
+            if isinstance(block, EventsBlock):
+                self.persist_events_block(block, finding_id)
+            # TODO should we add scans here
```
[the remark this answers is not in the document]
RoiGlinik: Ill check about merging the flows.

src/robusta/core/reporting/blocks.py
```diff
@@ -434,3 +436,41 @@ def grade(self):
...
         else:
             return "F"
+
+
+class EventsBlock(TableBlock):
+    """
+    Table display of a Events, Persists the events on the Robusta Platform.
+
+    Note: Wider tables appears as a file attachment on Slack, because they aren't rendered properly inline
+    """
```
RoiGlinik, line 446: @pablos44 , this is the issue for the blocks classes.
Pydantic does not support positional arguments, only named ones.  seems like v2 might support it in the future.

So for every constructor we have ( around 14 ) we need to add the param name in the constructor.
```
    alert.add_enrichment(
        [
            TableBlock(
                <missing> rows=[[k, v] for (k, v) in labels.items()],
                <missing> headers=["label", "value"],
                table_name="*Alert labels*",
            ),
        ],
        annotations={SlackAnnotations.ATTACHMENT: True},
    )
```

Some extra info.
https://github.com/pydantic/pydantic/issues/116

I would prefer to create a new PR after this one with the changes on all constructor calls.

helm/robusta/values.yaml
```diff
@@ -361,7 +361,8 @@ platformPlaybooks:
       exclude: ["NodeSysctlChange"]
   actions:
   - event_report: {}
-  - event_resource_events: {}
+  - event_resource_events:
+      platform_mute_finding: true
```
arikalon1, line 365: why do we need this?

playbooks/robusta_playbooks/event_enrichments.py
```diff
@@ -82,7 +83,13 @@ def event_resource_events(event: EventChangeEvent):
         obj.namespace,
     )
     if events_table:
-        event.add_enrichment([events_table], {SlackAnnotations.ATTACHMENT: True})
+        event.add_enrichment(
+            [events_table],
+            {
+                SlackAnnotations.ATTACHMENT: True,
+                EnrichmentAnnotation.PLATFORM_MUTE_FINDING: params.platform_mute_finding,
```
arikalon1, line 90: what is the case that this is false?

src/robusta/core/model/base_params.py
```diff
@@ -41,8 +41,12 @@ class ResourceChartResourceType(Enum):
 class ActionParams(DocumentedModel):
     """
     Base class for all Action parameter classes.
+
+    :var platform_mute_finding: platform sink will process the alert but will not create a finding.
     """
 
+    platform_mute_finding: bool = False
```
arikalon1, line 48: Is this flag relevant to all ActionParams?
Wouldn't it be better to add it in the EventParams?

src/robusta/core/reporting/blocks.py
```diff
@@ -434,3 +436,41 @@ def grade(self):
...
+
+    namespace: str
```
arikalon1, line 448: Why do we need it here?
We always have it on the finding
It's not a part of the enrichment really, is it?

src/robusta/core/sinks/robusta/dal/model_conversion.py
arikalon1: I think @Avi-Robusta is right.
When the UI calls the action, finding is not created in the DB.
It's called with `no sinks`. So the data is just returned by the websocket response

src/robusta/core/sinks/robusta/dal/supabase_dal.py
```diff
@@ -142,6 +143,16 @@ def persist_scan(self, enrichment: Enrichment):
                 raise Exception(msg)
 
     def persist_finding(self, finding: Finding):
+        mute_finding = any(
```
arikalon1, line 146: Why do we mute the finding?
We want the timeline to still behave the same, don't we?

src/robusta/core/sinks/robusta/dal/supabase_dal.py
```diff
@@ -537,3 +548,29 @@ def publish_cluster_nodes(self, node_count: int):
             self.handle_supabase_error()
 
         logging.info(f"cluster nodes: {UPDATE_CLUSTER_NODE_COUNT} => {data}")
+
+    def persist_events_block(self, block: EventsBlock, finding_id):
+        db_events = []
+        for row in block.render_rows():
+            event = dict(zip(block.headers, row))
```
arikalon1, line 555: This is a little bug prone
If someone accidentally add another column to table it will fail.
I think it's better to use some class that matches the row
It's also hard to understand what are the DB columns

arikalon1 (commented):
nice work
left some comments

pablos44 (approved):

PR #941, opened 2023-06-21
Save KRR strategy data to database

LeaveMyYard:
Save strategy data to a database (see https://github.com/robusta-dev/krr/pull/78)

Saving this data to each scan, as in future it will contain different information for different objects (HPA detection and strategy).

### DIFF
--- a/playbooks/robusta_playbooks/krr.py
+++ b/playbooks/robusta_playbooks/krr.py
@@ -86,8 +86,8 @@ class KRRResponse(BaseModel):
     scans: List[KRRScan]
     score: int
     resources: List[ResourceType] = ["cpu", "memory"]
-    description: Optional[str] = None # Old KRR versions do not return this field.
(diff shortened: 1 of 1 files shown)


playbooks/robusta_playbooks/krr.py
```diff
@@ -77,11 +77,17 @@ def priority(self) -> int:
...
+    name: str
+    settings: dict[str, Any]
+
+
 class KRRResponse(BaseModel):
     scans: List[KRRScan]
     score: int
     resources: List[ResourceType] = ["cpu", "memory"]
-    description: Optional[str] = None
+    description: Optional[str] = None # Old KRR versions do not return this field.
```
Avi-Robusta, line 89: I would just write which version of KRR this is added in or else all the new fields will have this comment

Avi-Robusta (approved):
looks good, one small comment

PR #942, opened 2023-06-22
Prometheus block is jsonified twice

RoiGlinik:
@Avi-Robusta  @mershal  Can this be changed safely on FE as well without backward compatible code?


arikalon1:
@RoiGlinik How would the frontend work?
Can we add some version attribute, so the UI knows how to parse it? (`version: 1.0`, if missing, use old method)

mershal:
@RoiGlinik Can you elaborate more? I'm not sure what this change is.

RoiGlinik:
@mershal This code converts the runner finding/action response to json. This change is about the prometheus blocks. Prometheus block were jsonified twice as a result, they require two json.parse calls. 
Its going to break the parsing for UI for graphs and every prometheus enricher code. 
@mershal  could it be solved  by parsing once and checking if its an object or string ?

PR #944, opened 2023-06-22
Handled a case where if the prom flags are unavailable then prometheu…

ganeshrvel:
…s was flagged as false.

### DIFF
--- a/src/robusta/core/sinks/robusta/prometheus_health_checker.py
+++ b/src/robusta/core/sinks/robusta/prometheus_health_checker.py
@@ -4,7 +4,8 @@ import time
 
 from pydantic import BaseModel
 
-from robusta.core.exceptions import NoPrometheusUrlFound, NoAlertManagerUrlFound, PrometheusFlagsConnectionError
+from robusta.core.exceptions import NoPrometheusUrlFound, NoAlertManagerUrlFound, PrometheusFlagsConnectionError, \
+    PrometheusNotFound, VictoriaMetricsNotFound, AlertsManagerNotFound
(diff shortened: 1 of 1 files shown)


otherguy (approved):
Nice! Can confirm that this fixes Prometheus integration on our GKE managed Prometheus instance.

src/robusta/core/sinks/robusta/prometheus_health_checker.py
```diff
@@ -53,16 +54,20 @@ def prometheus_connection_checks(self, global_config: dict):
...
             if prometheus_flags:
                 self.status.prometheus_retention_time = prometheus_flags.get('retentionTime', "")
             else:
                 self.status.prometheus_retention_time = ""
 
-            self.status.prometheus = True
-
         except Exception as e:
-            self.status.prometheus = False
+            if isinstance(e, NoPrometheusUrlFound) \
```
Avi-Robusta, line 66: I would just remove the exception PrometheusFlagsConnectionError at the end of get_prometheus_flags since it is known acceptable behavior and maybe print to the logs instead that flags could not be queried

Avi-Robusta (changes requested):
I would remove the PrometheusFlagsConnectionError since it it feels weird to throw a exception on acceptable behavior.

Avi-Robusta (approved):

src/robusta/core/sinks/robusta/prometheus_health_checker.py
ganeshrvel: The reason we have the PrometheusFlagsConnectionError is to handle and rate limit VM and prom exceptions separately.

PR #949, opened 2023-06-27
Node actions

RoiGlinik:


playbooks/robusta_playbooks/node_actions.py
```diff
@@ -0,0 +1,76 @@
...
+
+    try:
+        client.CoreV1Api().patch_node(node.metadata.name, {"spec": {"unschedulable": False}})
+        event.add_enrichment([MarkdownBlock(f"Node {node.metadata.name} uncordoned")])
+    except Exception as e:
+        raise ActionException(ErrorCodes.ACTION_UNEXPECTED_ERROR, f"Failed to uncordoned node {node.metadata.name} {e}")
+
+
+@action
+def drain(event: NodeEvent):
```
Avi-Robusta, line 50: I would add a comment about the function
also I would add these to the docs or talk to pavan about this

Avi-Robusta:
can you add a screenshot here for the release?

playbooks/robusta_playbooks/node_actions.py
Avi-Robusta: this function leaves the node in a cordoned state, so no new pods will be scheduled, i would add this to the comment

playbooks/robusta_playbooks/node_actions.py
```diff
@@ -0,0 +1,76 @@
...
+
+from robusta.api import ActionException, ErrorCodes, MarkdownBlock, NodeEvent, action
+
+
+@action
+def cordon(event: NodeEvent):
+    """
+    Cordon a node
+    """
+    node = event.get_node()
```
Avi-Robusta, line 14: do we have a way of detecting if a node is cordoned/unschedulable from the frontend? If not than i could see using these functions causing undetectable issues from the robusta UI

playbooks/robusta_playbooks/node_actions.py
```diff
@@ -0,0 +1,76 @@
...
+    api = client.CoreV1Api()
+    pods: V1PodList = api.list_pod_for_all_namespaces(watch=False, field_selector=f"spec.nodeName={node.metadata.name}")
+
+    msg = ""
+    for pod in pods.items:
+        name = pod.metadata.name
+        namespace = pod.metadata.namespace
+
+        owner_list: List[V1OwnerReference] = pod.metadata.owner_references if pod.metadata.owner_references else []
+        if any(owner.kind == "DaemonSet" for owner in owner_list):
```
Avi-Robusta, line 63: why do we skip daemonset owned pods? (also add a comment here explaining why)

playbooks/robusta_playbooks/node_actions.py
```diff
@@ -0,0 +1,76 @@
...
+            msg += f"Skipping {namespace}/DaemonSet/{name}\n"
+            continue
+
+        try:
+            body = client.V1Eviction(metadata=client.V1ObjectMeta(name=name, namespace=namespace))
+            api.create_namespaced_pod_eviction(name, namespace, body, pretty="True")
```
Avi-Robusta, line 69: do we have to check this result? what happens if we are deleting a pod that already terminated while we are doing this like a pod of a job?

playbooks/robusta_playbooks/node_actions.py
Avi-Robusta: also how long does this take on a fully loaded node? just an estimate?

Avi-Robusta (approved):
Looks good,
I had allot of questions I wanted clarity on.
And there were a few things I wanted you to leave comments on in the code.
Also add screenshots for the release, and make sure you add the actions to the docs
We might need to add some code to discovery to show when a node is unschedulable so we don't accidentally make something unschedulable and not show it in the table

playbooks/robusta_playbooks/node_actions.py
```diff
@@ -0,0 +1,76 @@
...
+        raise ActionException(ErrorCodes.ACTION_UNEXPECTED_ERROR, f"Failed to uncordoned node {node.metadata.name} {e}")
+
+
+@action
+def drain(event: NodeEvent):
+    cordon(event)
+
+    node = event.get_node()
+    api = client.CoreV1Api()
+    pods: V1PodList = api.list_pod_for_all_namespaces(watch=False, field_selector=f"spec.nodeName={node.metadata.name}")
```
Avi-Robusta, line 55: if this throws an exception for any reason, we leave the node in a cordoned state

playbooks/robusta_playbooks/node_actions.py
RoiGlinik: Yes visible, it will have a node.kubernetes.io/unschedulable=None:NoSchedule taint

playbooks/robusta_playbooks/node_actions.py
RoiGlinik: Daemonsets have the toleration to run on unscheduled nodes, So the drain was command not meant to evict them in that way. 
Also, a Daemonset runs on possibly all nodes, so it won't need a "reschedule".

playbooks/robusta_playbooks/node_actions.py
RoiGlinik: I think the action is auto generated, will double check

RoiGlinik: also takes like few seconds for 150 pods

aantn:
Like Avi said, please make sure to add to the docs.

PR #953, opened 2023-06-28
Add info to KRRRecommended

LeaveMyYard:
[This update](https://github.com/robusta-dev/krr/pull/80) introduced info for recommendations in KRR, currently used for explaining undefined recommendations ("?"):
- HPA Detected
- No data

Added them to the playbook for Robusta UI to utilize


Avi-Robusta (approved):
do we want to add info to the pdf?

LeaveMyYard:
> do we want to add info to the pdf?

I think we have future plans on improving those PDFs, so we might leave that for then

PR #956, opened 2023-06-28
fix: deployment resources.limits.cpu referencing wrong value

mehowthe:
Currently forwarder resources are defined as below: 
```
resources:
  requests:
    cpu: {{ .Values.kubewatch.resources.requests.cpu }}
    memory: {{ if .Values.isSmallCluster }}"64Mi"{{ else }}{{ .Values.kubewatch.resources.requests.memory | quote }}{{ end }}
  limits:
    memory: {{ if .Values.isSmallCluster }}"64Mi"{{ else }}{{ .Values.kubewatch.resources.requests.memory | quote }}{{ end }}
    {{ if .Values.kubewatch.resources.limits.cpu }}cpu: {{ .Values.kubewatch.resources.limits.cpu | quote }}{{ end }}
```
`Values.kubewatch.resources.requests.memory` is used both for requests and for limits, which isn't a desired behavior

### DIFF
--- a/helm/robusta/templates/forwarder.yaml
+++ b/helm/robusta/templates/forwarder.yaml
@@ -41,13 +41,13 @@ spec:
           capabilities: {}
           privileged: false
           readOnlyRootFilesystem: false
-          runAsUser: 1000    
+          runAsUser: 1000
         resources:
           requests:
             cpu: {{ .Values.kubewatch.resources.requests.cpu }}
             memory: {{ if .Values.isSmallCluster }}"64Mi"{{ else }}{{ .Values.kubewatch.resources.requests.memory | quote }}{{ end }}
           limits:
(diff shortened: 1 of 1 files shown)


arikalon1:
Hi @mehowthe , thanks for the PR

We don't have a memory limit in the default values. so I think this won't work if someone isn't defining memory limit in his own private values file.

Adding a default memory limit to the default values file, might break if someone defined a bigger memory request (because then he'll have request > limit)

Can you make it backward compatible like we did for the runner [here](https://github.com/robusta-dev/robusta/blob/2060543bd64f88b770d97565e2bc0de2ca778175/helm/robusta/templates/runner.yaml#L111)?

arikalon1:
This was fixed in #1296
Thanks for the PR @mehowthe 
Closing it

PR #959, opened 2023-06-29
webhook

RoiGlinik:
add option to use headers dict
add some error logs


RoiGlinik:
@Avi-Robusta  I hope to do a minor fix here, so dont review yet

Avi-Robusta (approved):

PR #960, opened 2023-07-02
Added memory limits for scan jobs

Avi-Robusta:


Avi-Robusta:
can set the limits from generated_values.yaml env var
Screen Shot 2023-07-02 at 14 12 05

PR #968, opened 2023-07-04
Moved Manage secrets from configuration to Installation

pavangudiwada:


aantn (approved):
LGTM. Do you need to remove `configuration/configuration-secrets` from `docs/index.rst` as well? (Under Configuration, where it was previously)?

PR #972, opened 2023-07-05
Workload rollout restart

RoiGlinik:


Avi-Robusta (changes requested):
I think this is the function that is called at `kubectl rollout restart`
https://github.com/kubernetes/kubectl/blob/release-1.26/pkg/polymorphichelpers/objectrestarter.go#L32

- it does not do a rollout if the deployment is paused so we should also add that logic in your action
(i checked and this function logically is the same from version 1.16)

Patch question - does it only add the field or does it overwrite the entire annotations object with the body passed
 it looks like the code from kubectl gets the object spec, adds annotations if it doesnt exist and overwrites the field restartedAt, where as we pass an object with only the fields we add so i wanted to now how it functions
 
 other than that, looks good

playbooks/robusta_playbooks/workload_actions.py
```diff
@@ -0,0 +1,59 @@
...
+                        "kubectl.kubernetes.io/restartedAt": datetime.datetime.utcnow()
+                        .replace(tzinfo=pytz.UTC)
+                        .isoformat()
+                    }
+                }
+            }
+        }
+    }
+    try:
+        if isinstance(resource, Deployment):
```
Avi-Robusta, line 47: check if its paused and dont do if it is, see comment

Avi-Robusta (approved):

PR #973, opened 2023-07-05
Added HTTP PUT custom playbook action

dmitriishuvalov:
we need to support also HTTP PUT custom playbook action in addition to existing HTTP GET and HTTP POST. 
It can very useful when we need to use REST API webhooks.


RoiGlinik:
@dmitriishuvalov  looks good, 
Can you also apply this minor fix to your function?
https://github.com/robusta-dev/robusta/pull/958/files

PR #974, opened 2023-07-05
fix link for investigating jobs

Avi-Robusta:


src/robusta/core/reporting/base.py
```diff
@@ -224,14 +224,24 @@ def _map_service_to_uri(self):
...
-            "namespaces": f'["{self.subject.namespace}"]' if self.subject.namespace else None,
-            "kind": self.service.resource_type if self.service else None,
-            "name": self.service.name if self.service else None,
-            "names": f'["{self.aggregation_key}"]' if self.aggregation_key else None,
-        }
+        kind = self.service.resource_type if self.service else None
+        if kind and kind.lower() == "job":
+            params = {
+                "account": account_id,
+                "cluster": f'"{cluster_name}"' if cluster_name else None,
```
RoiGlinik, line 231: Is the string interpolation necessary for cluster , namespace and service in the job flow ?

RoiGlinik:
@Avi-Robusta  Can be merged after this question

src/robusta/core/reporting/base.py
Avi-Robusta: yea, i ran it without by accident and it was super annoying to find out why it didnt work

PR #982, opened 2023-07-09
avoid namespace service and kind labels on node alerts

RoiGlinik:


src/robusta/core/reporting/base.py
```diff
@@ -272,18 +272,21 @@ def __str__(self):
 
     def get_prometheus_silence_url(self, account_id: str, cluster_name: str) -> str:
         labels: Dict[str, str] = {"alertname": self.aggregation_key, "cluster": cluster_name, "account": account_id}
-        if self.subject.namespace:
-            labels["namespace"] = self.subject.namespace
+        kind: Optional[str] = self.subject.subject_type.value
+        if kind and kind == "node":
+            pass
+        else:
+            if self.subject.namespace:
```
Avi-Robusta, line 279: couldnt 
```
        if kind and kind == "node":
            pass
        else:
            if self.subject.namespace:
```
    be simplified as 
`if kind != "node" and self.subject.namespace:`

src/robusta/core/reporting/base.py
```diff
@@ -272,18 +272,21 @@ def __str__(self):
 
     def get_prometheus_silence_url(self, account_id: str, cluster_name: str) -> str:
         labels: Dict[str, str] = {"alertname": self.aggregation_key, "cluster": cluster_name, "account": account_id}
-        if self.subject.namespace:
-            labels["namespace"] = self.subject.namespace
+        kind: Optional[str] = self.subject.subject_type.value
+        if kind and kind == "node":
+            pass
+        else:
```
Avi-Robusta, line 278: I would replace
```
        if kind and kind == "node":
            pass
        else:
```
with
` if kind != "node":`

src/robusta/core/reporting/base.py
```diff
@@ -272,18 +272,21 @@ def __str__(self):
...
+            if self.subject.namespace:
+                labels["namespace"] = self.subject.namespace
 
-        if self.silence_labels and self.silence_labels.get("service"):
-            labels["service"] = self.silence_labels["service"]
+            if self.silence_labels and self.silence_labels.get("service"):
```
Avi-Robusta, line 282: can't you have service for a node?

src/robusta/core/reporting/base.py
```diff
@@ -272,18 +272,21 @@ def __str__(self):
...
 
-        # In prometheus job is related to the scrape target.
-        # Kubernetes jobs are stored in job_name.
-        kind: Optional[str] = self.subject.subject_type.value
-        if kind and self.subject.name:
-            kind = "job_name" if kind == "job" else kind
-            labels[kind] = self.subject.name
+            # In prometheus, job is related to the scrape target.
+            # Kubernetes jobs are stored in job_name.
+            if kind and self.subject.name:
```
Avi-Robusta, line 287: are you sure you want to indent this, 
labels[kind] = self.subject.name
wont be run for nodes

RoiGlinik: Yes we want node alerts to be general, without a node name

src/robusta/core/reporting/base.py
RoiGlinik: Yes I like it, ill change it

PR #983, opened 2023-07-10
Corlogix managed prom support

Avi-Robusta:


arikalon1 (commented):
@Avi-Robusta Don't we need to somehow pass the cluster on the query?
Isn't that a managed Prometheus that has metrics from multiple clusters?

How do they know which cluster the metrics are from?

Avi-Robusta:
> @Avi-Robusta Don't we need to somehow pass the cluster on the query? Isn't that a managed Prometheus that has metrics from multiple clusters?
> 
> How do they know which cluster the metrics are from?

Im going to do that in a separate PR, its not a unique to corlogix issue it is the same for all the centralized prometheus from what I can see

RoiGlinik (approved):

PR #984, opened 2023-07-10
Fix/prometheus operator reconcile error

ganeshrvel:


playbooks/robusta_playbooks/alerts_integration.py
```diff
@@ -439,3 +443,76 @@ def stack_overflow_enricher(alert: PrometheusKubernetesAlert):
...
+    """
+
+    api = client.CoreV1Api()
+    matching_pods: List[V1Pod] = []
+    for selector in params.selectors:
+        try:
+            pods: V1PodList = api.list_pod_for_all_namespaces(label_selector=selector)
+            if pods.items:
+                matching_pods = pods.items
+                break
```
Avi-Robusta, line 471: this will only get the first selectors pods and not all of them. is that the behavior you want?
specifically with this issue there couldn't there be multiple prometheus operators so might want from all of them?
I would replace it with 
`matching_pods.extend(pods.items)
`

ganeshrvel: So @arikalon1 and I discussed about this on the call. The idea here is to use the first matching label selector just like we do for finding prometheus urls.

src/robusta/integrations/kubernetes/custom_models.py
```diff
@@ -166,23 +166,23 @@ def exec(self, shell_command: str, container: str = None) -> str:
...
-        previous=None,
-        tail_lines=None,
-        regex_replacer_patterns: Optional[List[NamedRegexPattern]] = None,
-        regex_replacement_style: Optional[RegexReplacementStyle] = None,
-        filter_regex: Optional[str] = None,
+    @staticmethod
+    def find_logs(
+            name: str,
+            namespace: str,
+            container=None,
```
Avi-Robusta, line 173: what happens to the function when the container is None?

src/robusta/integrations/kubernetes/custom_models.py
```diff
@@ -166,23 +166,23 @@ def exec(self, shell_command: str, container: str = None) -> str:
...
+            previous=None,
+            tail_lines=None,
+            regex_replacer_patterns: Optional[List[NamedRegexPattern]] = None,
+            regex_replacement_style: Optional[RegexReplacementStyle] = None,
+            filter_regex: Optional[str] = None,
     ) -> str:
         """
         Fetch pod logs, can replace sensitive data in the logs using a regex
         """
-        if container is None:
```
Avi-Robusta, line 181: I would re-add this and remove 
`container = self.spec.containers[0].name
` 
from the other to places in this PR

src/robusta/integrations/kubernetes/custom_models.py
Avi-Robusta: if the function fails I might add 
```
if container is None:
 raise exception...
```

playbooks/robusta_playbooks/alerts_integration.py
```diff
@@ -439,3 +443,76 @@ def stack_overflow_enricher(alert: PrometheusKubernetesAlert):
...
+        except Exception as e:
+            if not (isinstance(e, exceptions.ApiException) and e.status == 404):
+                raise ActionException(
+                    ErrorCodes.ACTION_UNEXPECTED_ERROR, f"Failed to list pods for foreign log enricher.\n{e}"
+                )
+
+    if not matching_pods:
```
Avi-Robusta, line 478: I would replace the for lines with

```
if not matching_pods:
    logging.warning(f"[foreign_logs_enricher] failed to find any matching pods for the selectors: {params.selectors}")
     return

for pod in matching_pods:
```
we dont need the second if and it limits the indentaions

playbooks/robusta_playbooks/alerts_integration.py
```diff
@@ -439,3 +443,76 @@ def stack_overflow_enricher(alert: PrometheusKubernetesAlert):
...
+            backoff_seconds: int = 2
+            regex_replacement_style = (
+                RegexReplacementStyle[params.regex_replacement_style] if params.regex_replacement_style else None
+            )
+
+            for _ in range(tries - 1):
+                if container is None:
+                    container = pod.spec.containers[0].name
+
+                log_data = RobustaPod.find_logs(
```
Avi-Robusta, line 502: why do we need RobustaPod.find_logs when we can just use pod.get_logs() here?

playbooks/robusta_playbooks/alerts_integration.py
```diff
@@ -439,3 +443,76 @@ def stack_overflow_enricher(alert: PrometheusKubernetesAlert):
...
+        except Exception as e:
+            if not (isinstance(e, exceptions.ApiException) and e.status == 404):
+                raise ActionException(
+                    ErrorCodes.ACTION_UNEXPECTED_ERROR, f"Failed to list pods for foreign log enricher.\n{e}"
+                )
+
+    if not matching_pods:
+        logging.warning(f"[foreign_logs_enricher] failed to find any matching pods for the selectors: {params.selectors}")
+
+    if matching_pods:
```
Avi-Robusta, line 481: it seems like all of this is the same code as `logs_enricher`
I would move all to its own function and use that function for here and logs enricher.
Something like this: 
`def pod_log_enricher(pod: Pod, params: LogEnricherParams) -> Optional[str]:
`
so we don't duplicate code

Avi-Robusta (changes requested):
Hey, good work,
it needs a bit of refactoring to remove the duplicated code.

src/robusta/integrations/kubernetes/custom_models.py
ganeshrvel: I have checked the necessity for a valid container name and it is not mandatory. It didnt throw any exceptions.
But in the file block we add the log name like this: `FileBlock(f"{pod.metadata.name}/{container}.log"` for which i have added some conditional file naming logic

playbooks/robusta_playbooks/alerts_integration.py
ganeshrvel: you are right, i missed the return statement

playbooks/robusta_playbooks/alerts_integration.py
ganeshrvel: Yup, i have deduplicated the code now.

Avi-Robusta (approved):
Looks allot better, great work.

playbooks/robusta_playbooks/alerts_integration.py
```diff
@@ -439,3 +452,46 @@ def stack_overflow_enricher(alert: PrometheusKubernetesAlert):
...
         ]
     )
+
+
+class ForeignLogParams(LogEnricherParams):
+    """
+    :var selectors: List of specific label selectors to retrieve logs from
+    """
+
+    selectors: List[str]
```
Avi-Robusta, line 462: Personally I would rename to 
`label_selectors`
so users dont accidentally put field selectors there without reading the docs

PR #986, opened 2023-07-11
Centralized prometheus support

Avi-Robusta:
To specify the label for the cluster add this to the generated_values.yaml

```
runner:
  - name: PROMETHEUS_CLUSTER_LABEL_NAME # this is optional, by default it is 'cluster'
    value: 'cluster' # some users use the value cluster_name or env
  - name: PROMETHEUS_CLUSTER_LABEL_VALUE
    value: 'eks-prom-2'
```


src/robusta/core/sinks/robusta/dal/model_conversion.py
```diff
@@ -97,7 +97,7 @@ def to_evidence_json(
                 structured_data.append({"type": "list", "data": block.items})
             elif isinstance(block, PrometheusBlock):
                 structured_data.append(
-                    {"type": "prometheus", "data": block.data.dict(), "metadata": block.metadata, "version": 1.0}
+                    {"type": "prometheus", "data": block.data.json(), "metadata": block.metadata, "version": 1.0}
```
arikalon1, line 100: @Avi-Robusta is that on purpose? 
@RoiGlinik fixed it few days ago (turning it from json to dict)

src/robusta/core/playbooks/prometheus_enrichment_utils.py
```diff
@@ -235,32 +247,32 @@ def create_resource_enrichment(
     ChartOptions = namedtuple("ChartOptions", ["query", "values_format"])
     combinations: Dict[ResourceKey, Optional[ChartOptions]] = {
         (ResourceChartResourceType.CPU, ResourceChartItemType.Pod): ChartOptions(
-            query='sum(irate(container_cpu_usage_seconds_total{namespace="$namespace", pod=~"$pod"}[5m])) by (pod, job)',
+            query='sum(irate(container_cpu_usage_seconds_total{namespace="$namespace", pod=~"$pod"$additional_labels}[5m])) by (pod, job)',
             values_format=ChartValuesFormat.CPUUsage,
         ),
         (ResourceChartResourceType.CPU, ResourceChartItemType.Node): ChartOptions(
-            query='instance:node_cpu_utilisation:rate5m{job="node-exporter", instance=~"$node_internal_ip:[0-9]+", cluster=""} != 0',
+            query='instance:node_cpu_utilisation:rate5m{job="node-exporter", instance=~"$node_internal_ip:[0-9]+"$additional_labels} != 0',
```
arikalon1, line 254: Why do we add the `additional_labels` here to the queries?
Wouldn't it be simpler to treat all the queries the same, and do the same replacement in a single place?

src/robusta/core/sinks/robusta/dal/model_conversion.py
Avi-Robusta: The UI graphs stopped working after the change in staging. I'll check again to see if it has been supported in the past two days

src/robusta/core/playbooks/prometheus_enrichment_utils.py
Avi-Robusta: good call

Avi-Robusta:
The docs are in a different PR since there is an open PR to refactor the docs
https://github.com/robusta-dev/robusta/pull/990/files

src/robusta/integrations/prometheus/utils.py
```diff
@@ -92,8 +92,12 @@ def request_new_token(cls) -> bool:
 
 
 def get_prometheus_connect(prometheus_params: PrometheusParams) -> "PrometheusConnect":
+    import urllib3
     from prometheus_api_client import PrometheusConnect
 
+    # urllib3 will spam a warning if this is disabled
```
RoiGlinik, line 98: @Avi-Robusta  lets do a fix later that turn/off this flag based on the prom_url

RoiGlinik (approved):

PR #988, opened 2023-07-13
Upgrade supabase-py to new supabase 1.0.3 python package

RoiGlinik:
* New package requires python 3.8
* Uses new exception and ApiResponse interface from the package.
* Fixes most of our old patches.
* Github action matrix test on python cli will now continue even if one specific version fail.
* Used https://github.com/supabase-community/supabase-py/issues/185 snippet to handle rls and tokens after sign up
* New rpc works sync, rpc patch still exists because of rpcs return value errors (errors with ApiResponse interface and bool return value).


.github/workflows/test_installation.yaml
```diff
@@ -29,6 +29,7 @@ jobs:
       needs: build_package
       runs-on: ubuntu-latest
       strategy:
+        fail-fast: false
         matrix:
           python-version: [ "3.7", "3.8", "3.9", "3.10" ]
```
arikalon1, line 34: it will now break on 3.7, right?

pablos44 (approved):
Great job!

.github/workflows/test_installation.yaml
pablos44: There is some issue however with building cli

src/robusta/core/sinks/robusta/dal/supabase_dal.py
```diff
@@ -340,48 +300,48 @@ def publish_jobs(self, jobs: List[JobInfo]):
...
+        try:
+            self.client.table(JOBS_TABLE).upsert(db_jobs).execute()
+        except Exception as e:
+            logging.error(f"Failed to persist jobs {jobs} error: {e}")
             self.handle_supabase_error()
-            status_code = res.get("status_code")
-            raise Exception(f"publish jobs failed. status: {status_code}")
+            raise
 
     def remove_deleted_job(self, job: JobInfo):
```
Avi-Robusta, line 310: hey, just double check that deleted jobs are out of the staging DB

.github/workflows/test_installation.yaml
RoiGlinik: Yes, I removed it from the text matrix

PR #992, opened 2023-07-16
add k8s version info in discovery

RoiGlinik:
Added the git version as its more verbose (contains patch and more)
example EKS
 provider': 'EKS', 'k8s_version': 'v1.25.11-eks-a5565ad'
example AKS
'provider': 'AKS', 'k8s_version': 'v1.25.6'

### DIFF
--- a/src/robusta/core/discovery/discovery.py
+++ b/src/robusta/core/discovery/discovery.py
@@ -409,6 +409,12 @@ class Discovery:
         except Exception:
             logging.error("Failed to count jobs", exc_info=True)
 
+        k8s_version: str = None
+        try:
+            k8s_version = client.VersionApi().get_code().git_version
+        except Exception:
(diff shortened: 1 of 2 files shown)


src/robusta/core/discovery/discovery.py
```diff
@@ -409,6 +409,12 @@ def discover_stats() -> ClusterStats:
         except Exception:
             logging.error("Failed to count jobs", exc_info=True)
 
+        k8s_version: str = None
+        try:
+            k8s_version = client.VersionApi().get_code().git_version
```
Avi-Robusta, line 414: can this change without restarting runner?
It might be something we only need to check once instead of calling it every five minutes

pablos44: it is possible to update Kubernetes version without restart

Avi-Robusta (approved):

PR #994, opened 2023-07-18
Labels matchers

arikalon1:


src/robusta/core/reporting/base.py
```diff
@@ -106,18 +106,28 @@ def __str__(self):
...
+    def attribute_matches(self, attribute: str, expression: Union[str, List[str], Dict, List[Dict]]) -> bool:
+        value: Union[str, Dict[str, str]] = self.attribute_map[attribute]
+        if isinstance(expression, str) or isinstance(expression, Dict):
+            return Filterable.__value_match(value, expression)
         else:  # expression is list of values
-            return value in expression
+            return any([Filterable.__value_match(value, single_exp) for single_exp in expression])
+
+    @staticmethod
+    def __value_match(value: Union[str, Dict[str, str]], expression: Union[str, Dict]) -> bool:
```
Avi-Robusta, line 123: will we ever have a legal match where value and expression are not of the same instance/type?

src/robusta/core/reporting/base.py
```diff
@@ -106,18 +106,28 @@ def __str__(self):
...
-        value = self.attribute_map[attribute]
-        if isinstance(expression, str):
-            return bool(re.match(expression, value))
+    def attribute_matches(self, attribute: str, expression: Union[str, List[str], Dict, List[Dict]]) -> bool:
+        value: Union[str, Dict[str, str]] = self.attribute_map[attribute]
+        if isinstance(expression, str) or isinstance(expression, Dict):
+            return Filterable.__value_match(value, expression)
         else:  # expression is list of values
-            return value in expression
+            return any([Filterable.__value_match(value, single_exp) for single_exp in expression])
```
Avi-Robusta, line 120: is that logically any different than this
```
        else isinstance(value, str):  # expression is list of values
            return value in expression
```
other than the warnings?
it seems a bit over complicated

src/robusta/core/reporting/base.py
```diff
@@ -106,18 +106,28 @@ def __str__(self):
...
+
+    @staticmethod
+    def __value_match(value: Union[str, Dict[str, str]], expression: Union[str, Dict]) -> bool:
+        if isinstance(value, str):
+            return bool(re.match(expression, value))
```
Avi-Robusta, line 125: will this throw an exception if 
`isinstance(expression, Dict)`

src/robusta/core/reporting/base.py
Avi-Robusta, line 120: is that logically any different than this
```
        else isinstance(value, str):  # expression is list of values
            return value in expression
```
other than the warnings?
it seems a bit over complicated

src/robusta/integrations/kubernetes/autogenerated/events.py
```diff
@@ -1051,28 +1269,29 @@ def get_subject(self) -> FindingSubject:
...
+    "clusterrole": ClusterRoleChangeEvent,
+    "clusterrolebinding": ClusterRoleBindingChangeEvent,
+    "job": JobChangeEvent,
+    "namespace": NamespaceChangeEvent,
+    "serviceaccount": ServiceAccountChangeEvent,
+    "persistentvolume": PersistentVolumeChangeEvent,
+    "persistentvolumeclaim": PersistentVolumeClaimChangeEvent,
+    "networkpolicy": NetworkPolicyChangeEvent,
+    "configmap": ConfigMapChangeEvent,
+    "ingress": IngressChangeEvent,
```
Avi-Robusta, line 1296: since we added labels + annotations to all of our events we should verify all of these kubernetes types have the metadata field and its manditory so we don't throw an exception when creating them

src/robusta/integrations/prometheus/models.py
```diff
@@ -118,36 +118,56 @@ def get_alert_subject(self) -> FindingSubject:
...
+            annotations = self.job.metadata.annotations
         elif self.hpa:
             subject_type = FindingSubjectType.TYPE_HPA
             name = self.hpa.metadata.name
+            labels = self.hpa.metadata.labels
+            annotations = self.hpa.metadata.annotations
 
-        return FindingSubject(name, subject_type, namespace, node_name, container)
+        # Add alert labels and annotations. On duplicates, alert labels/annotations are taken
+        labels = {**labels, **self.alert.labels}
```
Avi-Robusta, line 167: what happens here if the labels or annotations are None? (i saw my hpa has neither)

src/robusta/core/sinks/sink_base_params.py
```diff
@@ -13,4 +14,42 @@ class SinkBaseParams(BaseModel):
 
     @root_validator
     def env_values_validation(cls, values: Dict):
-        return replace_env_vars_values(values)
+        updated = replace_env_vars_values(values)
```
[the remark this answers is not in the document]
arikalon1: we do. If someone wants to use it
Any problem with that @Avi-Robusta ?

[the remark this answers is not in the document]
Avi-Robusta: no, just curious

src/robusta/integrations/prometheus/models.py
arikalon1: it's an empty dict.

src/robusta/core/reporting/base.py
arikalon1: This is a validation
If someone you get here with an illegal input, we handle it properly (instead of exception, we say the input is invalid)

src/robusta/core/reporting/base.py
arikalon1: it's different, it's not always a string
The comparison is different according to the matcher type

arikalon1: it's logically different

Avi-Robusta (approved):
Looks good, just had a few questions

src/robusta/core/reporting/base.py
```diff
@@ -106,18 +106,28 @@ def __str__(self):
...
+        elif isinstance(value, Dict) and isinstance(expression, Dict):  # value is Dict[str, str], expression is a Dict
+            return expression.items() <= value.items()
```
pablos44, line 127: Why when both are dict this matches  expression.items() <= value.items() ?

arikalon1: it checks that one dict is included in the other
We parse the string requirement into a dict, and then check for inclusion

src/robusta/core/reporting/base.py
arikalon1: fixed

PR #998, opened 2023-07-20
EKS managed prometheus

Avi-Robusta:
working version

fixed check connection query


RoiGlinik (approved):
@Avi-Robusta  Looks good. I think that on next provider we will have to do some design changes to the prometheus area

Avi-Robusta:
redundant, added in unified prometheus

PR #1014, opened 2023-07-31
RSA keys from existing secret

drustan:
The goal is to implement the ability to retrieve the RSA keys from an existing secret


drustan:
Hi! Is there any reason this is not reviewed / accepted?

arikalon1 (approved):
Sorry for the late reply @drustan 
Thanks for your contribution

jmalloc:
Any chance of a new tag that includes these changes? It took me a while to work out why this wasn't working, because the documentation for it is already live :wink:

arikalon1:
You're right @jmalloc 
The docs contains the change, but it wasn't released yet.
It will be included in 0.10.26
It should be released within a week or so

PR #1015, opened 2023-08-01
Added `max_log_file_limit_kb` in `slack_sink`

ganeshrvel:
Added `max_log_file_limit_kb` in `slack_sink`

max_log_file_limit_kb: 1000 # (Optional) The maximum allowed file size for log files (in kilobytes) uploaded to the Slack channel. Default: 1000.

### DIFF
--- a/src/robusta/core/reporting/blocks.py
+++ b/src/robusta/core/reporting/blocks.py
@@ -76,6 +76,31 @@ class FileBlock(BaseBlock):
         """
         super().__init__(filename=filename, contents=contents)
 
+    def truncate_content(self,  max_file_size_bytes: int) -> bytes:
+        """
+        Truncates the log file by removing lines from the beginning until its size is within the given limit.
+        """
+        content_length = len(self.contents)
+
+        if content_length <= max_file_size_bytes:
+            return self.contents
+
+        lines = self.contents.decode('utf-8')
+        lines = lines.splitlines()
+        byte_length_newline = len('\n'.encode('utf-8'))
+
+        truncated_lines: List[str] = []
+        for idx, line in enumerate(lines):
+            line_content_length = len(line.encode('utf-8'))
+            line_content_length = line_content_length + byte_length_newline
+
+            content_length -= line_content_length
+            if (content_length + byte_length_newline) <= max_file_size_bytes:
+                truncated_lines = lines[idx:]
+                break
+
+        return "\n".join(truncated_lines).encode('utf-8')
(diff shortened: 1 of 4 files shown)


src/robusta/core/reporting/blocks.py
```diff
@@ -76,6 +76,31 @@ def __init__(self, filename: str, contents: bytes):
         """
         super().__init__(filename=filename, contents=contents)
 
+    def truncate_content(self,  max_file_size_bytes: int) -> bytes:
+        """
+        Truncates the log file by removing lines from the beginning until its size is within the given limit.
+        """
+        content_length = len(self.contents)
```
Avi-Robusta, line 83: this is only correct for utf-8 strings
you should either convert it to utf-8 before you calculate the content_length and return the converted value or you should use this calculation for length here
`content_length = sys.getsizeof(self.contents) - sys.getsizeof("")`
(just double check to be sure)
below here. its okay to use len because it is utf-8 encoded

src/robusta/core/reporting/blocks.py
```diff
@@ -76,6 +76,31 @@ def __init__(self, filename: str, contents: bytes):
...
+        if content_length <= max_file_size_bytes:
+            return self.contents
+
+        lines = self.contents.decode('utf-8')
+        lines = lines.splitlines()
+        byte_length_newline = len('\n'.encode('utf-8'))
+
+        truncated_lines: List[str] = []
+        for idx, line in enumerate(lines):
+            line_content_length = len(line.encode('utf-8'))
```
Avi-Robusta, line 94: you could replace this with one line 
`line_content_length = len(line.encode('utf-8')) + byte_length_newline`

src/robusta/core/reporting/blocks.py
```diff
@@ -76,6 +76,31 @@ def __init__(self, filename: str, contents: bytes):
...
+            line_content_length = line_content_length + byte_length_newline
+
+            content_length -= line_content_length
+            if (content_length + byte_length_newline) <= max_file_size_bytes:
```
Avi-Robusta, line 98: why add the additional byte_length_newline here? we add it above too?

Avi-Robusta (dismissed):
Looks good,
Requested some small changes

arikalon1 (approved):
Looks good
nice work

arikalon1 (approved):

arikalon1 (approved):

PR #1016, opened 2023-08-02
Unified prometheus changes

Avi-Robusta:
refactored prometheus code to prometrix repo + added aws support approved code from a different pr
https://pypi.org/project/prometrix/
Approved PR for prometrix - https://github.com/kotlickya/prometrix/pull/3
https://github.com/kotlickya/prometrix

Added support to run KRR from robusta on all supported prometheus versions
Created additional secrets in the command line for krr when running on coralogix and EKS managed prometheus


src/robusta/utils/common.py
```diff
@@ -1,8 +1,8 @@
 from collections import defaultdict
-from typing import List, Dict
+from typing import Dict, List
+from urllib.parse import parse_qs
```
RoiGlinik, line 3: parse_qs , dict and defaultdict can be removed from import

pyproject.toml
```diff
@@ -64,6 +63,8 @@ cryptography = "^36.0.0"
 PyJWT = "2.4.0"
 fpdf2 = "^2.7.1"
 attrs = "^23.1.0"
+botocore = "^1.31.3"
```
RoiGlinik, line 66: is it still necessary ?

RoiGlinik (approved):
left some minor import questions. going to test it on azure

pyproject.toml
Avi-Robusta: good catch

src/robusta/integrations/prometheus/utils.py
```diff
@@ -26,195 +24,82 @@
...
-            f"{prom.url}/api/v1/status/flags",
-            verify=prom.ssl_verification,
-            headers=prom.headers,
-            # This query should return empty results, but is correct
-            params={},
+    # coralogix config
+    if CORALOGIX_PROMETHEUS_TOKEN:
+        return CoralogixPrometheusConfig(**baseconfig, prometheus_token=CORALOGIX_PROMETHEUS_TOKEN)
+    # Azure config
+    if os.environ.get("AZURE_USE_MANAGED_ID"):
```
RoiGlinik, line 69: The azure auth app flow is missing here, need to use something like that to support both flows.
`if os.environ.get("AZURE_USE_MANAGED_ID") or os.environ.get("AZURE_CLIENT_SECRET"):`

PR #1019, opened 2023-08-04
Added patch for errors like this: Duplicated timeseries in Collector

ganeshrvel:
Added patch for errors like this: `Duplicated timeseries in Collector Registry: {'services_updated'}`


arikalon1 (approved):
nice work

PR #1021, opened 2023-08-08
Discovery preformance

Avi-Robusta:
Notes:
needed to use os._exit(1) to kill robusta process

`_thread.interrupt_main()` still keeps other threads alive but calls handle_sigint in a parent thread
(~10 threads killed, the rest are still alive image taken after interrupt_main called and see from logs robusta still runs)
Screen Shot 2023-08-08 at 9 15 59
sys.exit will only exit the current child thread so its not an option from watchdog


Avi-Robusta:
replaced os.exit with sending the signal directly to the parent process -> sys.exit is called on correct thread.
Robusta container is restarted, stack dump present in previous logs

src/robusta/core/sinks/robusta/robusta_sink.py
```diff
@@ -467,6 +474,24 @@ def __should_run_history(self) -> bool:
             logging.error("Failed to check run history condition", exc_info=True)
             return False
 
+    def __discovery_watchdog(self):
+        logging.info("Cluster discovery watchdog initialized")
+        is_initial_health_check = True
+        while self.__active:
```
arikalon1, line 480: Can we simplify it by initializing the `last_send_time` to current time, inside `__update_cluster_status`?
This way you only need this:
```
            if not self.is_healthy():
                logging.info("Unhealthy discovery, restarting runner")
                # sys.exit and thread.interrupt_main doest stop robusta
                os.kill(os.getpid(), signal.SIGINT)
                return
```

src/robusta/core/sinks/robusta/robusta_sink.py
```diff
@@ -467,6 +474,24 @@ def __should_run_history(self) -> bool:
...
+            # if the initial thread is stuck it will register as "Healthy"
+            initial_discovery_stuck = not is_initial_health_check and self.last_send_time == 0
+            # if is not healthy we will get the stack dump from sigint handler
+            if initial_discovery_stuck:
+                StackTracer.dump(traces=1)
```
arikalon1, line 485: I think we always want few dumps.
Other wise you might not be able to tell if the thread is stuck (if for example, there's a weird scenario of an endless loop)

src/robusta/core/sinks/robusta/robusta_sink.py
```diff
@@ -467,6 +474,24 @@ def __should_run_history(self) -> bool:
...
+            if not self.is_healthy() or initial_discovery_stuck:
+                logging.info("Unhealthy discovery, restarting runner")
+                # sys.exit and thread.interrupt_main doest stop robusta
+                os.kill(os.getpid(), signal.SIGINT)
+                return
+            is_initial_health_check = False
+            time.sleep(DISCOVERY_CHECK_THRESHOLD_SEC)
```
arikalon1, line 492: I think we need a shorter sleep period.
Other wise, the actual recovery time is 2*DISCOVERY_CHECK_THRESHOLD_SEC

arikalon1 (approved):
Nice work

I think we might be able to simplify it a little, left some comments

src/robusta/core/sinks/robusta/robusta_sink.py
Avi-Robusta: its set before __update_cluster_status,
this is more for the rare case that the thread gets stuck before there, like in has_cluster_findings

src/robusta/core/sinks/robusta/robusta_sink.py
Avi-Robusta: DISCOVERY_CHECK_THRESHOLD_SEC its default 90 minutes, 
I would make its own env var, how many minutes do you want the default to be?

PR #1024, opened 2023-08-08
Improved the guess_cached_resource method.

ganeshrvel:
- Update: Only for replica-sets and pods we will use partial name match for other resource kinds we will use full name matching

### DIFF
--- a/src/robusta/core/discovery/top_service_resolver.py
+++ b/src/robusta/core/discovery/top_service_resolver.py
@@ -51,20 +51,29 @@ class TopServiceResolver:
     # TODO remove this guess function
     # temporary try to guess who the owner service is.
     @classmethod
-    def guess_service_key(cls, name: str, namespace: str) -> str:
-        resource = cls.guess_cached_resource(name, namespace)
+    def guess_service_key(cls, name: str, namespace: str, kind: str) -> str:
+        resource = cls.guess_cached_resource(name, namespace, kind=kind)
         return resource.get_resource_key() if resource else ""
 
     # TODO remove this guess function
     # temporary try to guess who the owner service is.
     @classmethod
-    def guess_cached_resource(cls, name: str, namespace: str) -> Optional[TopLevelResource]:
+    def guess_cached_resource(cls, name: str, namespace: str, kind: str,) \
+            -> Optional[TopLevelResource]:
         if name is None or namespace is None:
             return None
 
+        kind = kind.lower()
+
         for cached_resource in cls.__namespace_to_resource[namespace]:
-            if name.startswith(cached_resource.name):
-                return cached_resource
+            if kind == "pod" \
+                    or kind == "replicaset":
+                if name.startswith(cached_resource.name):
+                    return cached_resource
+            else:
+                if name == cached_resource.name:
+                    return cached_resource
+
         return None
 
     @classmethod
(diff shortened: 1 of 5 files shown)


src/robusta/core/discovery/top_service_resolver.py
```diff
@@ -51,20 +51,29 @@ def store_cached_resources(cls, resources: List[TopLevelResource]):
...
             return None
 
+        kind = kind.lower()
+
         for cached_resource in cls.__namespace_to_resource[namespace]:
-            if name.startswith(cached_resource.name):
-                return cached_resource
+            if kind == "pod" \
+                    or kind == "replicaset":
+                if name.startswith(cached_resource.name):
```
Avi-Robusta, line 71: maybe we can make this better 
like for pod named app-12346-sdfsgsdf
it will match with deployments named app-1, app-12 and app-123456
(with replica sets its less of an issue)
Pods have a replica set in the yaml, is there a way to use that?
Screen Shot 2023-08-09 at 9 33 04

Avi-Robusta: maybe do something like this 
```
def get_pod_owner_reference(cls, name: str, namespace: str, kind: str)
if kind == "pod":
 robusta_pod = RobustaPod.find_pod(name, namespace)
 return robusta_pod.owner_references[0] if pobusta_pod.owner_references else none
```
and than in the function guess_cached_resource
```
...
 owner_reference = get_pod_owner_reference()
 if owner_reference:
    kind = owner_reference.kind.lower()
    name = owner_reference.name
for cached_resource in cls.__namespace_to_resource[namespace]:
...
```

Avi-Robusta (changes requested):
Hey, 
We still have a bug here, I wrote some psudo code to make it better
Also sinc with tomer and shahal to see if this solves their issue on the timeline

src/robusta/core/discovery/top_service_resolver.py
ganeshrvel: done

src/robusta/core/discovery/top_service_resolver.py
```diff
@@ -51,20 +53,52 @@ def store_cached_resources(cls, resources: List[TopLevelResource]):
...
     @classmethod
-    def guess_service_key(cls, name: str, namespace: str) -> str:
-        resource = cls.guess_cached_resource(name, namespace)
+    def guess_service_key(cls, name: str, namespace: str, kind: str) -> str:
+        resource = cls.guess_cached_resource(name, namespace, kind=kind)
         return resource.get_resource_key() if resource else ""
 
+    @classmethod
+    def get_pod_owner_reference(cls, name: str, namespace: str) -> OwnerReference:
+        robusta_pod = RobustaPod.find_pod(name, namespace)
```
arikalon1, line 62: @Avi-Robusta @ganeshrvel I think doing this is a very bad idea.
This will make every call on a pod, to call the API server.
This is called for every finding.
It must be done within memory, without calling the k8s api server

arikalon1 (changes requested):
@Avi-Robusta @ganeshrvel I think we need to come with a better solution for this
Lets discuss

src/robusta/core/discovery/top_service_resolver.py
Avi-Robusta: Also arik mentioned even if we get the pod that we only have the replicaset owner,  since we want the owner of the replicaset like the deployment or job

PR #1026, opened 2023-08-09
Use unqualified names for Alertmanager & service discovery URL

### LINKED ISSUE #1025: Runner & prometheus stack alertmanager uses wrong URL when GKE VPC scope DNS is used
**Describe the bug**
With [GKE VPC scope DNS](https://cloud.google.com/kubernetes-engine/docs/how-to/cloud-dns#vpc_scope_dns), the cluster DNS uses a custom suffix instead of `svc.cluster.local`. 

**To Reproduce**
1. Deploy Robusta on GKE with VPC scope DNS enabled, with Prometheus stack enabled
2. Check alertmanager & runner logs

**Expected behavior**
DNS resolution working properly

**Screenshots**

Alertmanager Logs

```
ts=2023-08-09T08:43:16.945Z caller=notify.go:732 level=warn component=dispatcher receiver=robusta integration=webhook[0] msg="Notify attempt failed, will retry later" attempts=1 err="Post \"http://robusta-runner.monitoring.svc.cluster.local/api/alerts\": dial tcp: lookup robusta-runner.monitoring.svc.cluster.local on 169.254.20.10:53: no such host"
```

yahya-hrln:
Fixes #1025

### DIFF
--- a/helm/robusta/values.yaml
+++ b/helm/robusta/values.yaml
@@ -13,6 +13,9 @@ sinksConfig: []
 clusterName: ""
 clusterZone: ""
 
+global:
+  clusterDomain: "cluster.local"
+
 automountServiceAccountToken: true
 
 # see https://docs.robusta.dev/master/user-guide/configuration.html#global-config and https://docs.robusta.dev/master/configuration/additional-settings.html#global-config
@@ -540,7 +543,7 @@ kube-prometheus-stack:
         - name: 'null'
         - name:  'robusta'
           webhook_configs:
-            - url: 'http://robusta-runner.{{ .Release.Namespace }}.svc.cluster.local/api/alerts'
+            - url: 'http://robusta-runner.{{ .Release.Namespace }}.svc.{{ .Values.global.clusterDomain }}/api/alerts'
               send_resolved: true
     alertmanagerSpec:
       resources:
--- a/src/robusta/utils/service_discovery.py
+++ b/src/robusta/utils/service_discovery.py
@@ -4,6 +4,8 @@ from kubernetes import client
 from kubernetes.client import V1ServiceList
 from kubernetes.client.models.v1_service import V1Service
 
+from robusta.core.model.env_vars import CLUSTER_DOMAIN
+
 
 def find_service_url(label_selector):
     """
@@ -18,6 +20,6 @@ def find_service_url(label_selector):
     name = svc.metadata.name
     namespace = svc.metadata.namespace
     port = svc.spec.ports[0].port
-    url = f"http://{name}.{namespace}.svc.cluster.local:{port}"
+    url = f"http://{name}.{namespace}.svc.{CLUSTER_DOMAIN}:{port}"
     logging.info(f"discovered service with label-selector: `{label_selector}` at url: `{url}`")
     return url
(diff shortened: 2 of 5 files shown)


helm/robusta/values.yaml
```diff
@@ -511,7 +511,7 @@ kube-prometheus-stack:
         - name: 'null'
         - name:  'robusta'
           webhook_configs:
-            - url: 'http://robusta-runner.{{ .Release.Namespace }}.svc.cluster.local/api/alerts'
+            - url: 'http://robusta-runner.{{ .Release.Namespace }}/api/alerts'
```
arikalon1, line 514: @yahya-hrln Can you please make it configurable, so it won't change the default behavior for all users?
i.e., put the DNS suffix in a helm value, defaulted to `.svc.cluster.local`

My concern is, that some users have defined network policies for the existing url, and this change might break it for them

src/robusta/utils/service_discovery.py
```diff
@@ -18,6 +18,6 @@ def find_service_url(label_selector):
     name = svc.metadata.name
     namespace = svc.metadata.namespace
     port = svc.spec.ports[0].port
-    url = f"http://{name}.{namespace}.svc.cluster.local:{port}"
+    url = f"http://{name}.{namespace}:{port}"
```
arikalon1, line 21: Same here,
I would pass the helm value to the runner, using env variable, and use it here

arikalon1 (commented):
Sorry for the late review @yahya-hrln 
Left small comment

anakaiti:
@arikalon1 updated

helm/robusta/values.yaml
```diff
@@ -540,7 +543,7 @@ kube-prometheus-stack:
         - name: 'null'
         - name:  'robusta'
           webhook_configs:
-            - url: 'http://robusta-runner.{{ .Release.Namespace }}.svc.cluster.local/api/alerts'
+            - url: 'http://robusta-runner.{{ .Release.Namespace }}.svc.{{ .Values.global.clusterDomain }}/api/alerts'
```
arikalon1, line 546: The url you'd like on your case is `http://robusta-runner.default/api/alerts` right?
What will you put in `clusterDomain` for your cluster? is it ""
I think that will result in `http://robusta-runner.default.svc./api/alerts` 
Is it working this way?

src/robusta/utils/service_discovery.py
```diff
@@ -18,6 +20,6 @@ def find_service_url(label_selector):
     name = svc.metadata.name
     namespace = svc.metadata.namespace
     port = svc.spec.ports[0].port
-    url = f"http://{name}.{namespace}.svc.cluster.local:{port}"
+    url = f"http://{name}.{namespace}.svc.{CLUSTER_DOMAIN}:{port}"
```
arikalon1, line 23: same here

arikalon1 (commented):
left a small comment @yahya-hrln 
WDYT

helm/robusta/values.yaml
anakaiti: In my case, I'll set it to my GKE cluster's custom suffix. So it will still use an FQDN instead of an unqualified name.

arikalon1 (approved):
Looks good
Thanks for your PR @yahya-hrln

PR #1031, opened 2023-08-12
Add alerts processing pool as optional

arikalon1:


Avi-Robusta (approved):
looks good, maybe in the future if we detect it has more than X events per a second, we can enable it automatically.
Also I was curious why we use ProcessPoolExecutor instead of ThreadPoolExecutor, because its better for CPU bound? ( I saw thats what we did in the previous versions)

Avi-Robusta:
- [ ] add something in the docs or send something to pavan to add

PR #1032, opened 2023-08-13
supporting secret strings for telegram chatID

Avi-Robusta:


src/robusta/core/sinks/telegram/telegram_client.py
```diff
@@ -9,8 +10,11 @@
 
 
 class TelegramClient:
-    def __init__(self, chat_id: int, thread_id: int, bot_token: str):
-        self.chat_id = chat_id
+    def __init__(self, chat_id: Union[int, str], thread_id: int, bot_token: str):
+        if isinstance(chat_id, int):
```
RoiGlinik, line 14: I think the if could be omitted

RoiGlinik (approved):

PR #1035, opened 2023-08-13
Fixed a bug where Pod and Replicaset gets filtered out in the resource_events_diff

ganeshrvel:
Fixed a bug where Pod and Replicaset gets filtered out in the resource_events_diff


RoiGlinik:
@ganeshrvel  thats good, but ill fix it on the open jobs pr to avoid conflict

PR #1040, opened 2023-08-20
Support patterns on Slack channel_override

arikalon1:


RoiGlinik (approved):
@arikalon1 pr is ok overall. I believe it could simplified a bit with template safe_substitute ( format is a bit different), which is what used in the template enricher.

PR #1047, opened 2023-08-22
forward imagePullSecrets

JuanRamino:
Add imagePullSecrets to forward deployment


arikalon1 (approved):
Thanks for the PR @JuanRamino 

We just realized it's missing, so this is spot on :)

PR #1051, opened 2023-08-24
Automatic issue reply

pavangudiwada:


.github/workflows/new-contributors-autoreply.yaml
```diff
@@ -0,0 +1,15 @@
+name: Automated message for issues
+on:
+  issues:
+    types:
+      - opened
+jobs:
+  build:
+    name: Welcome new contributor 🎉
+    permissions: write-all
```
arikalon1, line 9: This seems way too permissive. 
Why are you giving all those permissions to the job?
https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

It's best to give minimal permissions.
Since our repo is public, anyone can reply to an issue.
Can we just give None?

arikalon1 (commented):
Looks good
I think we need to give less permissions

.github/workflows/new-contributors-autoreply.yaml
pavangudiwada: Yea you're right. We can change it to ` issues: write`. We can't use None, because the GitHub action can't reply. This feature is for this GitHub action to automatically reply when a new contributor/user creates an issue.

arikalon1 (approved):
LGTM

PR #1052, opened 2023-08-24
Custom annotations

Avi-Robusta:


Avi-Robusta:
Actions that need to be tested with new parameter (more were changed but have the same flow)
- [ ] krr_scan
- [ ] popeye_scan
- [ ] debugger_stack_trace
- [ ] disk_benchmark
- [ ] jdk actions (either one)
- [ ] node_disk_analyzer
- [ ] node_dmesg_enricher

aantn:
@Avi-Robusta don't forget to document this + add to the default values.yaml (people look at it to see which values are available)

PR #1067, opened 2023-08-31
Add info field to KRR playbook

LeaveMyYard:
Add info field to KRR playbook passing info field (HPA detected, no data, not enough data) to be displayed in UI


arikalon1 (approved):
LGTM

PR #1070, opened 2023-09-05
Discovery stackdump fix

Avi-Robusta:


src/robusta/core/discovery/discovery.py
```diff
@@ -54,9 +61,49 @@ class Config:
...
+                pipe.write(signal)
+                pipe.close()
+        except Exception:
+            logging.error("error writing pipe", exc_info=True)
+
+    @staticmethod
+    def stack_dump_on_signal():
+        try:
+            with open(DISCOVERY_PIPE) as pipe:
+                while True:
```
Avi-Robusta, line 98: have a function that waits until a signal is returned and returns the signal

src/robusta/core/discovery/discovery.py
```diff
@@ -54,9 +61,49 @@ class Config:
...
+        except Exception:
+            logging.error(
+                "Failed to create discovery pipes",
+                exc_info=True,
+            )
+
+    @staticmethod
+    def send_signal_to_pipe(signal):
+        try:
+            with open(DISCOVERY_PIPE, mode="w") as pipe:
```
Avi-Robusta, line 87: double check that this can never be blocking, (I had issues with other types of pipes i opened)

src/robusta/core/discovery/discovery.py
```diff
@@ -54,9 +61,49 @@ class Config:
...
+                logging.info("Sending signal to discovery thread")
+                pipe.write(signal)
```
Avi-Robusta, line 89: also check that this cant be blocking (waiting for the other side of the pipe to be open)

src/robusta/core/discovery/discovery.py
Avi-Robusta: in practice if the pipe wasnt open there was no issue, but check in docs

arikalon1: @Avi-Robusta what do you mean by blocking?
We don't want it to effect the discovery process on any occasion

arikalon1 (approved):
LGTM

PR #1072, opened 2023-09-07
Remove chart flag from create source helm

martynbristow:
`flux create source helm ...` should not have a `--chart` flag, it throws an error...

```
└─(05:22:49 on main ✹ ✭)──> flux create source helm robusta \                                                                                          ──(Thu,Sep07)─┘
   --url https://robusta-charts.storage.googleapis.com \
   --namespace robusta \
   --chart robusta \
   --export >> robusta.yaml
✗ unknown flag: --chart```

Also --namespace flag is missing in command

### DIFF
--- a/docs/setup-robusta/gitops/flux.rst
+++ b/docs/setup-robusta/gitops/flux.rst
@@ -73,6 +73,7 @@ Now lets use the HelmRepository to create a HelmRelease of Robusta.
 .. code-block:: bash
 
         flux create helmrelease robusta \
+            --namespace robusta \
             --chart=robusta \
             --source=HelmRepository/robusta \
             --chart-version="VERSION" \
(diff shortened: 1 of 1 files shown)


pavangudiwada:
Hi @martynbristow thanks for the PR. The chart repository has two charts, robusta and kubewatch. So I believe `--chart robusta` is needed to specify which chart we want to use.

Regarding the namespace, you might be right. I'll test it.

martynbristow:
When using --chart, an error is thrown.
Only when it's removed can you generate the yaml with flux. The behaviour is consistent with the flux documentation.

Let me grab you a screenshot to prove.

You are correct, you need to provide a name, but doesn't require the flag.

martynbristow:
As per: https://fluxcd.io/flux/cmd/flux_create_source_helm/
No --chart flag exists

pavangudiwada:
@martynbristow You are correct. `flux create source helm` does not have `--create`. It is similar to `helm repo add`, so the exact chart name does not matter until we run `flux create helmrelease` (helm install). 

The addition of  `--namespace robusta` is also valid. Thank you!

pavangudiwada (approved):
Fixes issues as expected.

martynbristow:
It looks to be failing due to a missing slack token.
There looks to be a flag to skip the test but perhaps it isn't working.
Is anyone able to advise ?

PR #1073, opened 2023-09-07
Fix error on node delete

Avi-Robusta:
in 
```
operation == K8sOperationType.DELETE
   ...
   new_info.deleted = True
```
would throw the exception
UnboundLocalError: local variable 'new_info' referenced before assignment


arikalon1 (approved):
LGTM

PR #1074, opened 2023-09-08
add support for extra volumes and volume mounts

gigawhat:
Add support for extra volumes and volume mounts to the runner.

We need to use extra volumes for setting up a scratch/temp directory. We run our k8s pods in clusters that require a readonly filesystem and `Matplotlib` library requires a temp directory it can write to. By adding support for additional volumes and mounts we are able to create an `emptyDir`, mount it to the runner container and then we set the `TMPDIR` env var to the `emptyDir` mount point.


arikalon1 (approved):
Nice work

arikalon1:
Thanks for your contribution @gigawhat

PR #1081, opened 2023-09-17
Increase supabase db timeout

arikalon1:
Small comments fix
Change log print to info, when failing to get Prometheus flags(not supported on some distros)


Avi-Robusta (approved):

Avi-Robusta:
looks good, its weird that the default from supabase was 5 seconds

PR #1082, opened 2023-09-18
Algolia docs

LeaveMyYard:
Reworked docs search to use Algolia search

### DIFF
--- a/docs/_templates/partials/header.html
+++ b/docs/_templates/partials/header.html
@@ -85,22 +85,25 @@
                     {{ alt.name }}
                   </a>
                 </li>
-                {% endfor %}
+              {% endfor %}
             </ul>
           </div>
         </div>
       </div>
     {% endif %}
 
-    <!-- Button to open search modal -->
-    {% if "search" in config["plugins"] %}
-      <label class="md-header__button md-icon" for="__search">
-        {% include ".icons/material/magnify.svg" %}
-      </label>
-
-      <!-- Search interface -->
-      {% include "partials/search.html" %}
-    {% endif %}
+    <!-- Algolia search -->
+    <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@docsearch/css@3" />
+    <div id="docsearch"></div>
+    <script src="https://cdn.jsdelivr.net/npm/@docsearch/js@3"></script>
+    <script>
+      docsearch({
+        container: '#docsearch',
+        appId: 'EEWU58CJVX',
+        indexName: 'robusta',
+        apiKey: 'e77d73e819e13d5b207909ee81891826',
+      });
+    </script>
 
     <!-- Repository information -->
     {% if config.repo_url %}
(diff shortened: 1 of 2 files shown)


docs/crawler.js
```diff
@@ -0,0 +1,103 @@
...
+              lvl1: ["article h2", "h2"],
+              lvl2: ["article h3", "h3"],
+              lvl3: ["article h4", "h4"],
+              lvl4: ["article h5", "h5"],
+              lvl5: ["article h6", "h6"],
+              content: ["article p, article li, article code"],
+              pageRank: (() => {
+                if (url.pathname.includes("/installation/index.html")) {
+                  return 10;
+                } else if (url.pathname.includes("/triggers/kubernetes.html")) {
```
aantn, line 32: @LeaveMyYard we need to prioritize all trigger and action pages, not just `/triggers/kubernetes.html` and `/actions/event-enrichment.html`

docs/crawler.js
```diff
@@ -0,0 +1,112 @@
+// This is the code example for the crawler configuration file for https://crawler.algolia.com/
```
aantn, line 1: How is this file used? It doesn't seem to be included anywhere.

docs/_templates/partials/header.html
```diff
@@ -85,22 +85,25 @@
...
-    <!-- Button to open search modal -->
-    {% if "search" in config["plugins"] %}
-      <label class="md-header__button md-icon" for="__search">
-        {% include ".icons/material/magnify.svg" %}
-      </label>
-
-      <!-- Search interface -->
-      {% include "partials/search.html" %}
-    {% endif %}
+    <!-- Algolia search -->
```
aantn, line 95: Would it be possible to make the fonts in the search results much larger? It's a bit hard for me to read. (I know we said not to worry about design, but I've changed my mind on this one aspect.)

docs/crawler.js
```diff
@@ -0,0 +1,112 @@
...
+new Crawler({
```
aantn, line 2: See if you can strip out the `¶` sign at the end of search results. If necessary, contact Algolia.

aantn (approved):
Approving with comments. This is already an improvement over existing search so I am merging despite the comments. Please open a follow up PR fixing comments.

aantn:
@pavangudiwada FYI

docs/crawler.js
LeaveMyYard: So the code is used at `https://crawler.algolia.com`
Because it defines on how a crawler indexes the docs and how the search after that is working - I descided to leave a reference in our codebase to not to loose it in the future

docs/_templates/partials/header.html
LeaveMyYard: I think we can make it even the same as it was before, but we will need to update our CSS files
If you want - I can do that, but it will take some time
Or also yes, I can increate just the text size also

docs/crawler.js
aantn: Good thinking, can you add a comment to the file making that explicit. (This file needs to be copied to crawler.algolia.com to work - its stored here as reference...)

PR #1087, opened 2023-09-20
Report pod-owned pods

Avi-Robusta:


src/robusta/core/discovery/discovery.py
```diff
@@ -497,6 +511,20 @@ def is_pod_ready(pod) -> bool:
     return False
 
 
+def should_report_pod(pod: Union[Pod, V1Pod]) -> bool:
+    if isinstance(pod, V1Pod):
```
arikalon1, line 515: I think you should start with `is_pod_finished`
If it is, return false.
Otherwise, continue to the rest (instead of checking `is_pod_finished` in multiple places)

arikalon1 (approved):
Looks good
Left one comment

PR #1088, opened 2023-09-20
Count pods in running state in discovery

RoiGlinik:
Add pod count to cluster nodes tables

Using discovery and not remaining_item_count  in this case

image


arikalon1 (approved):
looks good

PR #1089, opened 2023-09-21
added ignore when no caches

Avi-Robusta:


src/robusta/core/sinks/robusta/robusta_sink.py
```diff
@@ -154,6 +154,9 @@ def __reset_caches(self):
         self.__helm_releases_cache = None
         self.__namespaces_cache: Dict[str, NamespaceInfo] = {}
 
+    def __is_caches_initiated(self) -> bool:
```
RoiGlinik, line 157: @Avi-Robusta 
I think we still need to add a check for jobs and helm cache for not null,
Otherwise, the same bug can happen again on discovery failure.

Or maybe not using the null as a state for (jobs/helm), not sure I see the reason for it

Avi-Robusta: you cant have a cluster without nodes, services or namespaces but you can have a cluster with no jobs or helm charts, so when checking if the jobs cache is initialized if we initalized it with {} we would have no way when rechecking the cache object  it if there are no jobs or the cache is not initialised

RoiGlinik (approved):

PR #1090, opened 2023-09-21
Ingress playbook change

Avi-Robusta:
Screen Shot 2023-09-21 at 15 44 51

Needs change in frontend for new timeline, old timeline works well


arikalon1 (approved):
very nice

PR #1096, opened 2023-09-22
More algolia improvements

LeaveMyYard:
- Increased the text size in algolia popup
- Removed paragraph symbols from the search results


aantn:
Can we increase the font size a tiny bit more? After that, please merge.

aantn (approved):

PR #1100, opened 2023-09-27
Set cluster inactive on SIGINT and active on start

LeaveMyYard:
- Set `ClustersStatus.active = true` when starting the runner
- Set `ClustersStatus.active = false` when getting SIGINT (gracefull stopping)

Requires https://github.com/robusta-dev/robusta-storage/pull/36 migration


src/robusta/core/playbooks/playbooks_event_handler_impl.py
```diff
@@ -336,6 +337,11 @@ def is_healthy(
 
     def handle_sigint(self, sig, frame):
         logging.info("SIGINT handler called")
+
         if not self.is_healthy():  # dump stuck trace only when the runner is unhealthy
             StackTracer.dump()
+
+        for sink in self.registry.get_sinks().get_sinks_by_type(RobustaSink):
+            sink.dal.set_cluster_active(False)
```
Avi-Robusta, line 345: I would add this function in SinkBase
```
  def set_cluster_active(self, is_active: bool):
      pass
```
   and then in robusta_sink
```
    def set_cluster_active(self, is_active: bool):
      self.dal.set_cluster_active(is_active)
```
and then you don't need the function `get_sinks_by_type`, 
and then if we want to signal it in any other sink we can (like in slack)

src/robusta/core/sinks/robusta/dal/supabase_dal.py
```diff
@@ -527,3 +527,16 @@ def __update_token_patch(self, event, session):
         logging.debug(f"Event {event}, Session {session}")
         if session and event == "TOKEN_REFRESHED":
             self.client.postgrest.auth(session.access_token)
+
+    def set_cluster_active(self, active: bool) -> None:
+        try:
+            (
+                self.client.table(CLUSTERS_STATUS_TABLE)
+                .update({"active": active})
```
arikalon1, line 535: I think it would be also good to update the `updated_at` column to `now()`, so it's easy to understand when it happened last

src/robusta/runner/main.py
```diff
@@ -45,6 +45,8 @@ def main():
         logging.info(f"Running alerts workers pool of {ALERT_BUILDER_WORKERS}")
 
     Web.init(event_handler, loader)
+
+    event_handler.set_cluster_active(True)
```
arikalon1, line 49: I think it's a redundant db update.
When the server starts, the RobustaSink updates ClusterStatus. ([here](https://github.com/robusta-dev/robusta/blob/acfadcbe8270474c2bac741d0be69403fc7a9720/src/robusta/core/sinks/robusta/dal/supabase_dal.py#L416))
We can just add there `active: true` and remove this call (which can also fail and interfere with server startup)

arikalon1 (approved):
Looks good
Left 2 minor comments

src/robusta/runner/main.py
LeaveMyYard: That's correct, but from what Avi proposed, I made it so any sink can override this method and process being active as they want
So It resulted in two requests

PR #1102, opened 2023-09-27
fix: use correct api version/type

oscgu:
This makes it possible to deploy the servicemonitor when using an existing prometheus instance


RoiGlinik:
Thanks @oscg great catch!

PR #1105, opened 2023-10-03
Version mismatch playbook

Avi-Robusta:
Screen Shot 2024-02-21 at 16 22 17

### DIFF
--- a/playbooks/robusta_playbooks/version_mismatch_enricher.py
+++ b/playbooks/robusta_playbooks/version_mismatch_enricher.py
@@ -1,5 +1,6 @@
 import logging
 import re
+from functools import total_ordering
 from typing import List, Tuple
 
 from robusta.api import (
@@ -22,6 +23,7 @@ class VersionMismatchParams(PrometheusParams):
     version_mismatch_query: str = 'count by (git_version, node) (label_replace(kubernetes_build_info{job!~"kube-dns|coredns"}, "git_version", "$1", "git_version", "(v[0-9]*.[0-9]*).*"))'
 
 
+@total_ordering
 class Semver:
     """
     This is a class for comparing semantic version strings
@@ -102,7 +104,6 @@ def version_mismatch_enricher(alert: PrometheusKubernetesAlert, params: VersionM
     if len(kubernetes_api_versions) == 0:
         logging.error(f"Missing api server results for version_mismatch_enricher.")
         return
-
     kubernetes_api_version = _semver_max(kubernetes_api_versions)
     nodes_by_version = [
         [metric.get("node"), metric.get("git_version")] for metric in metrics if metric.get("node") is not None
--- a/playbooks/robusta_playbooks/version_mismatch_enricher.py
+++ b/playbooks/robusta_playbooks/version_mismatch_enricher.py
@@ -1,8 +1,9 @@
 import logging
 import re
 from functools import total_ordering
-from typing import List, Tuple
+from typing import Dict, List, Tuple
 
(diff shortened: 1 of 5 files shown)


Avi-Robusta:
Screen Shot 2024-02-21 at 16 22 17

playbooks/robusta_playbooks/version_mismatch_enricher.py
```diff
@@ -0,0 +1,100 @@
...
+            return False
+
+        # some prometheus's remove specific labels due to space constraints
+        if len([metric for metric in self.raw_metrics if metric.get("node") is not None]) == 0:
+            logging.error(f"Prometheus is missing the 'node' label, unable to detect node versions.")
+            return False
+
+        # you can have multiple versions of the api server in the metrics at once
+        # i.e. at the time of kubernetes upgrade both will show up in the metrics for several minutes
+        if len(self.api_versions) == 0:
```
arikalon1, line 48: Can we make this more flexible, and support the case we don't have the api server version?
Like, if for example, we have 2 node pools, with different versions

playbooks/robusta_playbooks/version_mismatch_enricher.py
```diff
@@ -0,0 +1,100 @@
...
+    if query_result.result_type == "error" or query_result.vector_result is None:
+        logging.error(f"version_mismatch_enricher failed to get prometheus results.")
+        return
+
+    build_infos = BuildInfoResults(query_result)
+    if not build_infos.is_results_valid():
+        logging.error(f"Invalid prometheus results for version_mismatch_enricher.")
+        return
+
+    if len(build_infos.api_versions) > 1:
```
arikalon1, line 80: I think I would also support the case we don't know the api server version (i.e. len(api_versions) == 0 )

playbooks/robusta_playbooks/version_mismatch_enricher.py
```diff
@@ -0,0 +1,100 @@
...
+        api_version_string = f"has reported versions {', '.join(build_infos.api_versions)}"
+    else:
+        api_version_string = f"is version {build_infos.api_versions[0]}"
+
+    # in the case where a node is of a higher version than the api server
+    alert.add_enrichment(
+        [
+            MarkdownBlock(f"Automatic {alert.alert_name} investigation:"),
```
arikalon1, line 88: I would remove this.
I think we don't need to state we're investigating, and just show the results

src/robusta/core/playbooks/prometheus_enrichment_utils.py
```diff
@@ -350,9 +351,38 @@ def create_chart_from_prometheus_query(
...
+    For graphs use run_prometheus_query_range which uses the prometheus query_range api
+    """
+    try:
+        prom = get_prometheus_connect(prometheus_params)
+        query = __add_additional_labels(query, prometheus_params)
+        prom_params = {"timeout": PROMETHEUS_REQUEST_TIMEOUT_SECONDS}
+        prom.check_prometheus_connection(prom_params)
+        results = prom.safe_custom_query(query=query, params=prom_params)
+        return PrometheusQueryResult(results)
+    except Exception as e:
```
arikalon1, line 377: I think I would not surround with try/except
We have a global code to handle Prometheus errors

Also, someone calling the api might not expect having the error as a field of the result

arikalon1 (approved):
nice work
left few minor comments

Avi-Robusta:
testing

PR #1112, opened 2023-10-08
cofiguring kubewatch from values file

Avi-Robusta:
able to set the kubewatch configs from values file with 
```
kubewatch:
  config:
    resource:
      deployment: false
      replicationcontroller: false  # 0.10.12 disabled because not supported on the runner
      replicaset: false
...
```


ganeshrvel (approved):

Avi-Robusta:
changed base image to bash to allow for kubectl exec bin/sh into the pod 
it does not have apt-get or dpkg but kubectl cp works

PR #1118, opened 2023-10-09
Update sinks-development.rst

aantn:


RobertSzefler (approved):
+1 for the cleanup but I think maybe it would be useful not to remove all the mentions of `src/robusta/integrations` - that was actually helpful when I started working on the mail sink. OTOH when someone looks inside any sink code in `src/robusta/core/sinks` it's pretty obvious that code from `src/robusta/integrations` is referenced, so idk

aantn:
You will have the most context on what is (un)clear, useful, and missing here.

PR #1121, opened 2023-10-12
Added Ingress triggers to the docs

pavangudiwada:
image

https://docs.robusta.dev/fix-docs-ingress-triggers/playbook-reference/triggers/kubernetes.html#ingress-triggers


docs/playbook-reference/triggers/_k8s-generic-triggers.jinja
```diff
@@ -3,7 +3,11 @@
 {{ resource_name }} Triggers
 ************************************
 
+{% if resource_name == "Ingress" %}
```
aantn, line 6: Nice catch :)

aantn (approved):
+1

PR #1125, opened 2023-10-13
Fix mattermost channel selection

daanvinken:
Fixing my own mistake. The implemented channel search uses a search term, so if we have multiple hits it could pick the wrong channel.


arikalon1 (approved):
Looks good
Thanks for your contribution @daanvinken

PR #1128, opened 2023-10-16
SMTP sink

RobertSzefler:
HTML layouts seem to be OK :)


LeaveMyYard (approved):

src/robusta/integrations/mail/sender.py
```diff
@@ -0,0 +1,198 @@
...
+    return obj
+
+
+class MailTransformer(Transformer):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.file_blocks = []
+
+    def block_to_html(self, block: BaseBlock) -> str:
+        # TODO should we additionally support ScanReportBlock here?
```
arikalon1, line 30: Yes, I think we should
Any reason not to?
We can convert it to PDF and send it as a file attachment (like we do on Slack)
I think users will use it

src/robusta/integrations/mail/sender.py
```diff
@@ -0,0 +1,198 @@
...
+    def send_finding_via_email(self, finding: Finding, platform_enabled: bool):
+        blocks: List[BaseBlock] = []
+
+        if finding.title:
+            status: FindingStatus = (
+                FindingStatus.RESOLVED if finding.title.startswith("[RESOLVED]") else FindingStatus.FIRING
+            )
+            blocks.append(self.__create_finding_header(finding, status))
+        else:
+            # TODO is this correct? Is it possible for the finding to have no title at all?
```
arikalon1, line 64: I think it's not possible, and `title` is mandatory

src/robusta/integrations/mail/sender.py
```diff
@@ -0,0 +1,198 @@
...
+            for file_block in transformer.file_blocks:
+                # This is awkward, but it's the standard way to handle
+                # attachments in apprise - by providing local filesystem
+                # names.
+                f = tempfile.NamedTemporaryFile()
+                attachment_files.append(f)
+                f.write(file_block.contents)
+                attachments.add(f.name)
+            ap_obj.add(self.mailto)
+            logging.info(f"MailSender: sending title={finding.title}, body={html_body}")
```
arikalon1, line 93: This might be too spammy, maybe change it to debug?

arikalon1 (approved):
Great work @RobertSzefler 

Left some minor comments

docs/configuration/sinks/mail.rst
```diff
@@ -0,0 +1,30 @@
...
+------------------------------------------------
+
+.. admonition:: Add this to your generated_values.yaml
+
+    .. code-block:: yaml
+
+        sinksConfig:
+        - mail_sink:
+            name: mail_sink
+            mailto: mailtos://user:password@server&from=a@x&to=b@y,c@z
```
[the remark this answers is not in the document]
RobertSzefler: It [looks like](https://stackoverflow.com/a/33149072) having a `:` in a value is ok as long as it's not followed by a space (the YAML spec is weird), and indeed it works ok with PyYAML, but to be extra sure I added quotes and a warning in the docs.

src/robusta/integrations/mail/sender.py
RobertSzefler: removed the "else" part

src/robusta/integrations/mail/sender.py
```diff
@@ -0,0 +1,129 @@
...
+    LinkProp,
+    MarkdownBlock,
+)
+from robusta.core.reporting.consts import EnrichmentAnnotation
+from robusta.core.sinks.transformer import Transformer
+
+
+class MailTransformer(Transformer):
+    def __init__(self, *args, **kwargs):
+        super(MailTransformer).__init__(*args, **kwargs)
```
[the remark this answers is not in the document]
RobertSzefler: fixed

tests/test_mail_sink.py
```diff
@@ -0,0 +1,83 @@
...
+    finding = Finding(
+        title=title,
+        description="Lorem ipsum",
+        aggregation_key="1234",
+        add_silence_url=True,
+    )
+    with patch("robusta.integrations.mail.sender.apprise") as mock_apprise:
+        sink.write_finding(finding, platform_enabled=True)
+    mock_apprise.Apprise.return_value.add.assert_called_once_with("mailtos://user:password@example.com?from=a@x&to=b@y")
+    expected_body = (
```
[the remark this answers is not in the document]
RobertSzefler: fixed

src/robusta/integrations/mail/sender.py
RobertSzefler: changed

src/robusta/integrations/mail/sender.py
RobertSzefler: In fact it looks like the Slack sink is the *only* sink that handles ScanReportBlocks, so I was not sure if it's important at all. Will add.

RobertSzefler: Added support for ScanReportBlocks via PDF files, just as in the Slack sink. Additionally added a check & a warning in `robusta.core.sinks.transformer`

playbooks/pyproject.toml
```diff
@@ -17,7 +17,8 @@ include_trailing_comma = true
 python = "^3.7.1"
 CairoSVG = "^2.5.2"
 Flask = "^2.0.2"
-prometheus-api-client = "^0.4.2"
+#prometheus-api-client = "^0.4.2"
```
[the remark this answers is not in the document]
RobertSzefler: fixed

PR #1129, opened 2023-10-17
Add custom labels on pods

michMartineau:
We would like to add custom labels on forwarder and runner pods. We will use them in order to show which team is responsible for this pods.


arikalon1 (approved):
Looks good
Thanks for your contribution @michMartineau

Issue #1117, opened 2023-10-09
UnicodeEncode error with webhook sink when message contain a non-ASCII character


Tassatux:
**Describe the bug**
UnicodeEncode error with webhook sink when finding message contain a non-ASCII character, like an emoji added by cpu_throttling enricher.

**To Reproduce**
1. Configure a webhook sink
2. Send Prometheus CPUThrottlingHigh alert

**Expected behavior**
The message should be encoded in `utf-8` so bytes will be provided to `requests.post()` instead of unicode.

By default, if a python string (unicode) is provided to the http.client lib (used by requests), it will be [encoded in latin-1](https://github.com/python/cpython/blob/3.12/Lib/http/client.py#L164) following [RFC 2616](https://datatracker.ietf.org/doc/html/rfc2616#section-3.7.1).

**Traceback**

```
\u001b[32m2023-10-04 15:16:02.999 INFO     Loaded k8s pod netbox-prometheus-sd-fcc8c5c95-mv72k for alert CPUThrottlingHigh\u001b[0m
UnicodeEncodeError: 'latin-1' codec can't encode character '\\U0001f4d8' in position 194: Body ('📘') is not valid Latin-1. Use body.encode('utf-8') if you want to send it encoded in UTF-8.\u001b[0m
    raise UnicodeEncodeError(
  File \"/usr/local/lib/python3.9/http/client.py\
    body = _encode(body
  File \"/usr/local/lib/python3.9/http/client.py\
    self._send_request(method
  File \"/usr/local/lib/python3.9/http/client.py\
    super(HTTPConnection
  File \"/usr/local/lib/python3.9/site-packages/urllib3/connection.py\
    conn.request(method
  File \"/usr/local/lib/python3.9/site-packages/urllib3/connectionpool.py\
    httplib_response = self._make_request(
  File \"/usr/local/lib/python3.9/site-packages/urllib3/connectionpool.py\
    resp = conn.urlopen(
  File \"/usr/local/lib/python3.9/site-packages/requests/adapters.py\
    r = adapter.send(request
  File \"/usr/local/lib/python3.9/site-packages/requests/sessions.py\
    resp = self.send(prep
  File \"/usr/local/lib/python3.9/site-packages/requests/sessions.py\
    return session.request(method=method
  File \"/usr/local/lib/python3.9/site-packages/requests/api.py\
    return request('post'
  File \"/usr/local/lib/python3.9/site-packages/requests/api.py\
    r = requests.post(self.url
[...]

PR #1130, opened 2023-10-17
added encode option

swastik959:
solves : #1117


arikalon1 (approved):
Nice work
Thanks for your contribution @swastik959

PR #1136, opened 2023-10-27
Playbook generator

aantn:
WIP. Do not merge or review yet.


scripts/main_app.py
```diff
@@ -0,0 +1,15 @@
+import streamlit as st
+from pages import demo_playbooks, playbook_builder
+from streamlit import session_state as ss
+
+if "current_page" not in st.session_state:
```
aantn, line 5: @pavangudiwada we're inconsistent in this file. Sometimes we set session_state via `ss` and sometimes we do it via `st.session_state` - can you choose one and use it consistently.

scripts/main_app.py
```diff
@@ -0,0 +1,15 @@
...
+    st.session_state["current_page"] = "demo_playbooks"
+
+if "playbook_choosen" not in st.session_state:
```
aantn, line 8: @pavangudiwada typo in `choosen` - should be `chosen``

pavangudiwada: Fixed

scripts/main_app.py
pavangudiwada: Done :heavy_check_mark:

scripts/main_app.py
```diff
@@ -0,0 +1,14 @@
+from pages import demo_playbooks, playbook_builder
+from streamlit import session_state as ss
+
+if "current_page" not in ss:
+    ss["current_page"] = "demo_playbooks"
```
aantn, line 5: @pavangudiwada one more comment on consistency: either stick with `ss["foo"]` or `ss.foo` but don't mix and match unless there is a reason.

Issue #1141, opened 2023-10-30
Slack SDK fails on large file uploads


aiell0:
**Describe the bug**
```/usr/local/lib/python3.9/site-packages/slack_sdk/web/internal_utils.py:450: UserWarning: client.files_upload() may cause some issues like timeouts for relatively large files. Our latest recommendation is to use client.files_upload_v2(), which is mostly compatible and much stabler, instead.```

**To Reproduce**
1. Have a k8s cluster with a lot of stuff on it.
2. Attempt to use krr with a slack sink to send cost reports to Slack.

**Expected behavior**
The file should be in my slack channel.

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - OS: mac OSX 
 - N/A
 - Sonoma 14.1

**Smartphone (please complete the following information):**
 - N/A

**Additional context**
Slack looks like they fixed this in their [latest release](https://github.com/slackapi/python-slack-sdk/releases/tag/v3.23.0). So theoretically all one should need to do is update the dependencies and it should work.

PR #1142, opened 2023-10-30
fix: update slack sdk to resolve large file timeout issue

aiell0:
Solve issue: https://github.com/robusta-dev/robusta/issues/1141


arikalon1:
Thanks for the PR @aiell0 

Did you test and it works?
I think we tried upgrading to the new Slack api in the past, but had an issue with it, because it required some additional permissions our app did not have.

arikalon1 (approved):
Thanks for your contribution @aiell0

PR #1144, opened 2023-11-01
k8s vs Prometheus alert distinction in Slack sink code

RobertSzefler:
Adds the distinction between K8s events and Prometheus alerts in Slack messages. Also rephrases a message in the mail sink to be consistent with the Slack code.

### DIFF
--- a/src/robusta/integrations/slack/sender.py
+++ b/src/robusta/integrations/slack/sender.py
@@ -25,7 +25,7 @@ from robusta.core.reporting.blocks import (
     TableBlock,
 )
 from robusta.core.reporting.callbacks import ExternalActionRequestBuilder
-from robusta.core.reporting.consts import EnrichmentAnnotation, SlackAnnotations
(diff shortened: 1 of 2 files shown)


src/robusta/integrations/slack/sender.py
```diff
@@ -278,18 +278,25 @@ def __send_blocks_to_slack(
...
+            status_name: str = (
+                "{status.to_emoji()} `Prometheus Alert Firing`"
+                if status == FindingStatus.FIRING
+                else "*Prometheus resolved*"
+            )
+        elif finding.source == FindingSource.KUBERNETES_API_SERVER:
+            status_name: str = "👀 *K8s event detected*"
+        else:
+            # TODO how should we format findings with other sources?
+            status_name: str = "???"
```
arikalon1, line 293: I would just name this `notification`

arikalon1 (approved):
LGTM

Added one minor comment
Please fix and merge

src/robusta/integrations/slack/sender.py
RobertSzefler: Done

PR #1147, opened 2023-11-06
Fixed broken URL, added an announcement on the top and an empty Robot.txt

pavangudiwada:


docs/_templates/base.html
```diff
@@ -35,3 +35,8 @@
   }, false);
 </script>
 {% endblock %}
+{% block announce %}
```
aantn, line 38: Can you split this to a separate PR and ask Keren for feedback on colors?

aantn (approved):

docs/_templates/base.html
pavangudiwada: Done

Issue #1103, opened 2023-09-28
Helm: Possibility to set PodSecurityContext as well as individual container SecurityContext


tibuntu:
**Is your feature request related to a problem?**
We have very strict policies for our Kubernetes clusters. Especially certain values in the SecurityContext of pods & containers are required so that pods are accepted by the API.

**Describe the solution you'd like**
It would be fantastic, if one could modify the PodSecurityContext as well as the individual container SecurityContext via the Helm-Values.

aantn:
Hi @tibuntu,
We don't have the option yet, but a PR would be welcome as long as it doesn't change the default behaviour. See here - https://github.com/robusta-dev/robusta/blob/master/helm/robusta/templates/runner-service-account.yaml

PR #1148, opened 2023-11-06
[helm] add the possibility to overwrite Pod and Container SecurityContext

lukashankeln:
Issue: #1103


arikalon1 (approved):
Looks good
Thanks for the PR @lukashankeln

PR #1149, opened 2023-11-06
Reduce memory limits

RobertSzefler:


src/robusta/cli/main.py
```diff
@@ -288,6 +288,13 @@ def gen_config(
             },
         ]
 
+    if is_small_cluster:
+        setattr(values, "prometheus", {})
```
arikalon1, line 292: I think this should be under a `kube-prometheus-stack` section

arikalon1 (commented):
left one comment

src/robusta/cli/main.py
RobertSzefler: fixed

arikalon1 (approved):
LGTM

PR #1150, opened 2023-11-07
Improve runner service monitor in order to be more configurable

michMartineau:
We would like to have a more configurable service monitor for the runner.
It is useful when prometheus is external in order to get runner metrics. 
So we would like to :
- add custom labels on the resources
- add relabeling, interval, scrapetimeout ...


arikalon1 (approved):
Thanks for the PR @michMartineau 

Looks good

PR #1152, opened 2023-11-07
Extensible Kafka Auth Config

stvnksslr:
Current kafka sink does not support authentication which makes it difficult to adopt or use in many situations. This PR allows the kafka sink to behave as it is but to have an auth block added to easily cater to most if not all the supported authentications schemas of the python kafka package.

Tested With
* Apache kafka
* AWS MSK

Not Tested
* Confluent kafka

```yaml
  sinksConfig:
  - kafka_sink:
      name: kafka_sink
      kafka_url: "localhost:9096"
      topic: "robusta-playbooks"
      auth:
          sasl_mechanism: SCRAM-SHA-512
          security_protocol: SASL_SSL
          sasl_plain_username: robusta
          sasl_plain_password: password
```

additionally updated skaffold configs to the latest version, i can easily roll this back if its not wanted or needed.

### DIFF
--- a/docs/configuration/sinks/kafka.rst
+++ b/docs/configuration/sinks/kafka.rst
@@ -16,6 +16,21 @@ Configuring the Kafka sink
             kafka_url: "localhost:9092"
             topic: "robusta-playbooks"
 
+.. admonition:: Add this to your generated_values.yaml if configuring with authentication
+
+   .. code-block:: yaml
+
+        sinksConfig:
+        - kafka_sink:
+            name: kafka_sink
+            kafka_url: "localhost:9096"
+            topic: "robusta-playbooks"
+            auth:
+                sasl_mechanism: SCRAM-SHA-512
+                security_protocol: SASL_SSL
+                sasl_plain_username: robusta
+                sasl_plain_password: password
+
 Save the file and run
 
 .. code-block:: bash
--- a/docs/configuration/sinks/kafka.rst
+++ b/docs/configuration/sinks/kafka.rst
@@ -18,6 +18,8 @@ Configuring the Kafka sink
 
 .. admonition:: Add this to your generated_values.yaml if configuring with authentication
 
+   `additional auth options <https://kafka-python.readthedocs.io/en/master/apidoc/KafkaProducer.html#kafkaproducer>`_
+
    .. code-block:: yaml
 
         sinksConfig:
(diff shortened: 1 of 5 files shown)


docs/configuration/sinks/kafka.rst
```diff
@@ -16,6 +16,21 @@ Configuring the Kafka sink
             kafka_url: "localhost:9092"
             topic: "robusta-playbooks"
 
+.. admonition:: Add this to your generated_values.yaml if configuring with authentication
```
LeaveMyYard, line 19: I would add the link to KafkaProducer docs on fields available for the __init__
Maybe to rename `auth` to something like `kwargs`? I am not sure on this one tho

skaffold.yaml
```diff
@@ -1,67 +1,77 @@
-apiVersion: skaffold/v2beta20
+apiVersion: skaffold/v4beta6
```
LeaveMyYard, line 1: Yes, please make it in a separate PR

LeaveMyYard (changes requested):

skaffold.yaml
stvnksslr: will revert

docs/configuration/sinks/kafka.rst
stvnksslr: I was also conflicted here and went back and forth a few times with something like `additonalConfigs`, `kwargs` ect.

I think spelling out the auth and the link to the different auth types is the most direct way to resolve questions with how do you connect kafka. If its too generic I can see people getting confused where they put the auth.

stvnksslr:
Found an issue with my implementation where you couldnt use the `{{ .env.SOME_SECRET }}` env substitution because the dict wasnt being validated. I have two solutions here, the one ive included in the latest commit extends replace_env_vars_values to be able to recursively search any dictionaries too. 

```python
elif isinstance(value, dict):
    env_var_value = replace_env_vars_values(value)
    if env_var_value:
        values[key] = env_var_value
```

but this change in the root validator might be more encompassing than the maintainers want, Option B would be to extend just the kafka_sink_params to do a recursive search.

```python
     Possible Option B.
     @validator("auth", pre=True, always=True)
     def auth_env_values_validation(cls, value: Dict):
         updated = replace_env_vars_values(value)
         return updated
```

@LeaveMyYard let me know what you think when you have time.

LeaveMyYard (approved):
I think the first solution could be usefull for such cases in a future, so we can keep it

PR #1154, opened 2023-11-09
[helm] add extraVolumeMounts and extraVolumes on Forwarder

michMartineau:
When automountServiceAccountToken is set to false, It is not possible to mount manually the service account token for the forwarder (It is possible with the runner).
I would like to add the possibility to mount the token volume.
```
    kubewatch:
      extraVolumes:
        - name: sa-token
          projected:
            sources:
              - serviceAccountToken:
                  path: token
                  expirationSeconds: 3607
              - configMap:
                  items:
                    - key: ca.crt
                      path: ca.crt
                  name: kube-root-ca.crt
              - downwardAPI:
                  items:
                    - fieldRef:
                        apiVersion: v1
                        fieldPath: metadata.namespace
                      path: namespace
      extraVolumeMounts:
        - name: sa-token
          mountPath: /var/run/secrets/kubernetes.io/serviceaccount/
          readOnly: true
```

### DIFF
--- a/helm/robusta/templates/forwarder.yaml
+++ b/helm/robusta/templates/forwarder.yaml
@@ -48,6 +48,9 @@ spec:
         volumeMounts:
           - name: robusta-kubewatch-config
             mountPath: /config
+          {{- with .Values.kubewatch.extraVolumeMounts }}
+          {{- toYaml . | nindent 10 }}
+          {{- end }}
         securityContext:
           allowPrivilegeEscalation: false
           capabilities: {}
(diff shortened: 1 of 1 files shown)


arikalon1 (commented):
Thanks for the PR @michMartineau 

Can you add empty defaults to `values.yaml`, so it'll be easier to understand the option exists (similar to what we have on the runner)

arikalon1 (approved):

arikalon1:
thanks for the PR @michMartineau

PR #1156, opened 2023-11-09
Enhanced prometheus queries using manual_action

ganeshrvel:
Added:
- All prometheus metrics
- Labels and values for specific metric


playbooks/robusta_playbooks/prometheus_enrichments.py
```diff
@@ -42,6 +47,52 @@ def parse_duration(
...
+    event.add_enrichment([JsonBlock(json.dumps({"metrics": result}))])
+
+
+def get_prometheus_all_available_metrics(prometheus_params: PrometheusParams) -> List[str]:
+    try:
+        prom = get_prometheus_connect(prometheus_params=prometheus_params)
+        return prom.all_metrics()
+
+    except Exception as e:
+        raise Exception(f"An error occurred while fetching all available Prometheus metrics: {e}")
```
arikalon1, line 63: If you're raising an error this way, you lose the original stack trace
I suggest logging the error, with exc_info=true, and raising the original exception
```
logging.error("An error occurred while fetching all available Prometheus metrics", exc_info=True)
raise e
```

playbooks/robusta_playbooks/prometheus_enrichments.py
```diff
@@ -42,6 +47,52 @@ def parse_duration(
...
+
+
+def get_prometheus_series(prometheus_params: PrometheusGetSeriesParams) -> dict:
+    try:
+        prom = get_prometheus_connect(prometheus_params=prometheus_params)
+        return prom.get_series(match=prometheus_params.match, end_time=prometheus_params.end_time,
+                               start_time=prometheus_params.start_time)
+
+    except Exception as e:
+        raise Exception(f"Failed to fetch Prometheus series for match criteria {prometheus_params.match} "
```
arikalon1, line 92: If you're raising an error this way, you lose the original stack trace
I suggest logging the error, with exc_info=true, and raising the original exception
```
logging.error(f"Failed to fetch Prometheus series for match criteria {prometheus_params.match} within the time range {prometheus_params.start_time} - {prometheus_params.end_time}", exc_info=True)
raise e
```

arikalon1 (commented):
nice work
left minor comments

arikalon1 (approved):
LGTM

Issue #1163, opened 2023-11-14
Improve the Jira Sink


paoloyx:
**Is your feature request related to a problem?**
Jira integration is very useful, albeit a bit basic in its current form. I would love Robusta to have some adding features like:

* Closing Jira issues when a matching solved alert is fired
* Configure Jira Sink behavior through additional params (i.e. send resolved status to Jira, re-open closed issues if new alarms arrive, etc. etc.)
* Update existing issues with new content from Prometheus alerts (if alert's fingeprint matches)
* Carry on additional fields coming from Prometheus alerts (i.e. `runbook_url` above the others)

**Describe the solution you'd like**
The overall flow is the one depicted here
![image](https://github.com/robusta-dev/robusta/assets/991842/d5d002b7-5057-49bc-9ab9-92ebabe3a27c)

In a nutshell, when a Prometheus alerts fire:

* If there is NOT a matching Jira issue, create a new Jira issue
* If there is a matching Jira issue that is not in DONE status, then:
   * Mark the issue is Done if alert is resolved AND Robusta is configured to send resolved status to Jira
   * Update issue summary/description if alert is not resolved
* If there is a matching jira issue in DONE status, then:
   * If alert is resolved, there is nothing to do as Jira issue is already done
   * If alert is not resolved then:
      * If re-open issue behavior is active, re-open DONE issue
      * Create a new issue otherwise

**Describe alternatives you've considered**
I did not found alternatives in Robusta codebase, I was more or less used to what another project does for Prometheus ([jiralert](https://github.com/prometheus-community/jiralert)) and I'd like to implement them in Robusta

aantn:
This makes a lot of sense, thank you. The diagram is especially helpful.

We will review in more depth soon.

arikalon1:
Thanks for the PR @paoloyx

PR #1164, opened 2023-11-14
Improve the Jira sink

paoloyx:
This PR improves the Jira sink as it is proposed in #1163.
It's my first PR for Robusta, in a nutshell here you will find following improvements

# Improve the [prometheus-alert](https://docs.robusta.dev/master/playbook-reference/actions/miscellaneous.html#prometheus-alert) action
I think that the [prometheus-alert](https://docs.robusta.dev/master/playbook-reference/actions/miscellaneous.html#prometheus-alert) action is already great, but adding following params could make it even more useful
* `runbook_url`: could be useful to show this URL in Jira issue. See [alertmanager]([prometheus-alert](https://docs.robusta.dev/master/playbook-reference/actions/miscellaneous.html#prometheus-alert) action) docs
* `fingeprint`: could be useful to pass a fingeprint to test Jira issue auto-solve feature that this PR implements 

# Improve the Jira sink
The logical flow is described in issue #1163 
The improvements only works for Prometheus alerts.

# Configure Jira Issue Priority
Jira issue will be created with a priority that follows the `severity` of Prometheus alert

# Please note
I was not able to add integration tests as they are already provided, for example, for Slack integration. I honestly don't know if there exist a "Jira mock" that can be used.

Manual tests successfully covered all the logical flow described in issue #1163


docs/configuration/sinks/jira.rst
```diff
@@ -27,6 +27,12 @@ Optional Settings
 * ``project_type_id_override`` : [Optional - default: None] If available, will override the ``project_name`` configuration. Follow these `instructions <https://confluence.atlassian.com/jirakb/how-to-get-project-id-from-the-jira-user-interface-827341414.html>`__ to get your project id.
 * ``issue_type_id_override`` : [Optional - default: None] If available, will override the ``issue_type`` configuration. Follow these `instructions <https://confluence.atlassian.com/jirakb/finding-the-id-for-issue-types-646186508.html>`__ to get your issue id.
 
+Following optional settings only work for Prometheus alerts:
```
paoloyx, line 30: These are additional params that Jira sink will use

src/robusta/core/sinks/jira/jira_sink_params.py
```diff
@@ -13,6 +13,10 @@ class JiraSinkParams(SinkBaseParams):
     project_name: str
     project_type_id_override: Optional[int]
     issue_type_id_override: Optional[int]
+    sendResolved: Optional[bool]
+    reopenIssues: Optional[bool]
+    doneStatusName: Optional[str] = "Done"
```
paoloyx, line 18: these statuses "Done" and "To Do" should be "default" ones in Jira, as I understand from the [docs](https://support.atlassian.com/jira-cloud-administration/docs/what-are-issue-statuses-priorities-and-resolutions/)

src/robusta/integrations/jira/client.py
```diff
@@ -54,58 +67,69 @@ def _call_jira_api(self, url, http_method: HttpMethod = HttpMethod.GET, **kwargs
...
+            )
             return None
 
-    def list_issues(self, search_params: Optional[str] = None):
-        endpoint = "search"
-        search_params = search_params or ""
-        url = self._get_full_jira_url(endpoint)
-        query = {"jql": search_params}
-        return self._call_jira_api(url, params=query) or []
+    def _get_transition_id(self, issue_id, status_name):
```
paoloyx, line 83: gets the transition id associated to a particular status i.e. "Done" or "To Do"

src/robusta/integrations/jira/client.py
```diff
@@ -54,58 +67,69 @@ def _call_jira_api(self, url, http_method: HttpMethod = HttpMethod.GET, **kwargs
...
+        transitions = self._get_transitions_for_issue(issue_id)
+        for t in transitions:
+            if t.get("name", "").lower() == status_name.lower():
+                return t.get("id", None)
+        return None
 
-    def create_issue(self, issue_data, issue_attachments=None):
-        endpoint = "issue"
+    def _get_transitions_for_issue(self, issue_id):
```
paoloyx, line 90: Gets all the transitions for a target Jira issue

src/robusta/integrations/jira/client.py
```diff
@@ -131,6 +155,122 @@ def _get_default_project_id(self):
...
+
+        payload = {
+            "fields": {"summary": summary, "description": description},
+        }
+
+        logging.debug(f"Update issue with payload: {payload}")
+        response = self._call_jira_api(url, http_method=HttpMethod.PUT, json=payload) or {}
+        logging.debug(response)
+
+    def manage_issue(self, issue_data, alert_data, issue_attachments=None):
```
paoloyx, line 238: Here the logic is implemented. See #1163

docs/configuration/sinks/jira.rst
```diff
@@ -27,6 +27,12 @@ Optional Settings
...
+* ``sendResolved`` : [Optional - default: false] If true, will transition a Jira issue to "Done" when the alert event contains a "solved" status.
```
arikalon1, line 31: Should it be "resolved" instead of "solved"?

src/robusta/integrations/jira/client.py
```diff
@@ -131,6 +155,122 @@ def _get_default_project_id(self):
...
+
+        if existing_issue and is_prometheus_alert:
+            issue_done = self._check_issue_done(existing_issue)
+
+            if issue_done:
+                if alert_resolved:
+                    logging.warn("Ignoring resolved alert that is already 'done' in Jira")
+                elif self.reopenIssues:
+                    self.transition_issue(existing_issue, self.reopenStatusName)
+                    self.update_issue(existing_issue, issue_data)
```
arikalon1, line 253: Should we add more data when reopening the issue?
Maybe just a comment that the issue was reopened at this time?

src/robusta/integrations/jira/client.py
```diff
@@ -131,6 +155,122 @@ def _get_default_project_id(self):
...
+                else:
+                    self.create_issue(issue_data, issue_attachments)
+            else:
+                if alert_resolved and self.sendResolved:
+                    self.transition_issue(existing_issue, self.doneStatusName)
+                    self.comment_issue(existing_issue)
+                else:
+                    self.update_issue(existing_issue, issue_data)
```
arikalon1, line 261: On the case `alert_resolved` is `true`, but `sendResolved` is `false` (default), we're still updating the issue, but not changing the Jira status.
Should we maybe add a comment that the issue was resolved?

src/robusta/integrations/jira/client.py
```diff
@@ -131,6 +155,122 @@ def _get_default_project_id(self):
...
+        else:
+            self.create_issue(issue_data, issue_attachments)
```
arikalon1, line 263: When the alert is resolved, and there's no previous issue, a new issue will be created on Jira (regardless of the `sendResolved` parameter)

src/robusta/integrations/jira/sender.py
```diff
@@ -200,6 +209,11 @@ def send_finding_to_jira(
                     to_paragraph(f"🎬 {video_link.name}", [{"type": "link", "attrs": {"href": video_link.url}}])
                 )
 
+        # Add runbook_url to issue markdown if present
+        if finding.subject.annotations["runbook_url"]:
```
arikalon1, line 213: This will throw KeyError in case there's no `runbook_url`
Might be better to use `finding.subject.annotations.get("runbook_url", None)`

src/robusta/integrations/jira/sender.py
```diff
@@ -219,11 +233,20 @@ def send_finding_to_jira(
             elif attr == "cluster_name":
                 labels.append(self.cluster_name)
 
-        self.client.create_issue(
+        status: FindingStatus = (
+            FindingStatus.RESOLVED if finding.title.startswith("[RESOLVED]") else FindingStatus.FIRING
+        )
+
+        # Default priority is "Major" if not a standard severity is given
+        severity = SEVERITY_JIRA_ID[finding.severity] or "Major"
```
arikalon1, line 241: Here is probably also better to use `get()` to avoid `KeyError` with unknown severities

arikalon1 (commented):
Thanks for the PR @paoloyx
Great work!

Left few minor comments

docs/configuration/sinks/jira.rst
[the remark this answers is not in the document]
paoloyx: Thanks, fixed in commits that I'm about to push

playbooks/robusta_playbooks/prometheus_simulation.py
```diff
@@ -41,6 +43,8 @@ class PrometheusAlertParams(ActionParams):
     description: str = "simulated prometheus alert"
     summary: Optional[str]
     generator_url = ""
+    runbook_url: Optional[str] = None
+    fingerprint: Optional[str] = None
```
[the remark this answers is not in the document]
paoloyx: Done, thanks!

playbooks/robusta_playbooks/prometheus_simulation.py
```diff
@@ -100,6 +105,8 @@ def prometheus_alert(event: ExecutionBaseEvent, prometheus_event_data: Prometheu
                     "endsAt": datetime.now(),
                     "startsAt": datetime.now(),
                     "generatorURL": prometheus_event_data.generator_url,
+                    "runbookURL": prometheus_event_data.runbook_url,
```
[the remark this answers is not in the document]
paoloyx: @arikalon1 the meaning is that I wanted to test the `runbookURL` with Prometheus simulation that Robusta provides. That value is put into the `annotations` dict into the `prometheus_simulation.py` code.

A real prometheus alert can have the `runbookURL` into its annotations, and so imho this field allows to test the feature.

If it's ok for you, we could leave this here, what do you think?

src/robusta/integrations/jira/client.py
```diff
@@ -54,58 +67,69 @@ def _call_jira_api(self, url, http_method: HttpMethod = HttpMethod.GET, **kwargs
...
 
-    def add_attachment(self, issue_id, issue_attachments):
-        endpoint = f"issue/{issue_id}/attachments"
-        url = self._get_full_jira_url(endpoint)
-        for attachment in issue_attachments:
-            data = MultipartEncoder(fields={"file": attachment})
-            request_headers = {"Content-type": data.content_type}
-            response = self._call_jira_api(url, http_method=HttpMethod.POST, data=data, headers=request_headers) or {}
-            logging.debug(response)
+    def _get_dict_field(self, issue, field, default_value):
```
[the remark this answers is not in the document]
paoloyx: Yes, done, thanks!

docs/configuration/sinks/jira.rst
paoloyx: Yes, thanks you, I changed it

src/robusta/integrations/jira/client.py
paoloyx: I'll add some checks, like we create the alert only if the alert is NOT solved

src/robusta/integrations/jira/sender.py
paoloyx: Done, thanks

src/robusta/integrations/jira/client.py
```diff
@@ -131,6 +153,133 @@ def _get_default_project_id(self):
...
+        logging.debug(response)
+
+    def manage_issue(self, issue_data, alert_data, issue_attachments=None):
+        existing_issue = self._get_issue_for_labels(issue_data.get("labels", [])) or []
+        status = self._get_nested_property(alert_data, "status", -1)
+        source = self._get_nested_property(alert_data, "source", None)
+        alert_resolved = status == FindingStatus.RESOLVED
+        is_prometheus_alert = source == FindingSource.PROMETHEUS
+
+        if not is_prometheus_alert:
```
paoloyx, line 240: @arikalon1 I've slightly updated the logic here as follows:

* If it's not a Prometheus alert, we create the issue on Jira without doing any processing
* Else we do our processing

src/robusta/integrations/jira/client.py
```diff
@@ -131,6 +153,133 @@ def _get_default_project_id(self):
...
+                    self.update_issue(issue_id, issue_data)
+                else:
+                    self.create_issue(issue_data, issue_attachments)
+            else:
+                if alert_resolved:
+                    if self.sendResolved:
+                        self.transition_issue(issue_id, self.doneStatusName)
+                        self.comment_issue(issue_id, "Issue was marked as resolved by the Alertmanager")
+                    else:
+                        logging.warn("Alert is resolved but 'sendResolved' is false, so we don't update Jira")
```
paoloyx, line 267: If we have a open issue and a _resolved_ alert arrives, and `sendResolved` is false, then we simply warn the user that we won't update Jira. After all, the user selected to NOT send resolved status to jira with the `senResolved` value.

src/robusta/integrations/jira/client.py
```diff
@@ -131,6 +153,133 @@ def _get_default_project_id(self):
...
+                else:
+                    self.update_issue(issue_id, issue_data)
+        else:
+            if not alert_resolved:
```
paoloyx, line 271: If we have no previous issue, we open a new one only if alert is NOT resolved

src/robusta/integrations/jira/client.py
```diff
@@ -131,6 +153,133 @@ def _get_default_project_id(self):
...
+            issue_done = self._check_issue_done(existing_issue)
+            issue_id = self._get_nested_property(existing_issue, "id", -1)
+            issue_resolution = self._get_nested_property(existing_issue, "fields.resolution.name", "")
+
+            if issue_done:
+                if alert_resolved:
+                    logging.info(
+                        f"Ignoring resolved alert that is already 'done' in Jira for issue with id '{issue_id}'"
+                    )
+                elif self.noReopenResolution and issue_resolution == self.noReopenResolution:
```
paoloyx, line 254: This is a new piece of logic that I've introduced. I think that it could be useful.
It allows anyone to stop Robusta from re-opening issues on an issue basis.

For example, let's say that I've an issue that I know I won't fix, and that issue is closed with a "Won't Fix" resolution on Jira. Then I could use this resolution with the `noReopenResolution` JiraParam to prevent Robusta from re-opening it.

playbooks/robusta_playbooks/prometheus_simulation.py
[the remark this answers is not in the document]
arikalon1: Can you explain how you use it?

We take the alert input, and put it in this class: (the schema is from [here](https://prometheus.io/docs/alerting/latest/notifications/#alert))
```
class PrometheusAlert(BaseModel):
    endsAt: datetime
    generatorURL: str
    startsAt: datetime
    fingerprint: Optional[str] = ""
    status: str
    labels: Dict[Any, Any]
    annotations: Dict[Any, Any]
```

So, I think the `runbook_url' in that location, will not be included in the notification at all.
Am I wrong on that?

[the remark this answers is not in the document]
paoloyx: @arikalon1 Yes, I've overlooked that code line, even my explanation above is wrong. We're using `runbook_url` within the annotation, so that line is redundant as you correctly pointed out.

I've removed it, thanks

PR #1170, opened 2023-11-16
Document default Prometheus labels mapping

aantn:


pavangudiwada (approved):
Works as expected.

image

PR #1172, opened 2023-11-16
Allow on_pod_conditions to be None in V1PodFailurePolicyRule

RobertSzefler:
Some monkey-patching of the k8s client to allow on_pod_conditions to be set to None for V1PodFailurePolicyRule


arikalon1 (approved):
LGTM

PR #1177, opened 2023-11-20
add support for setting pod / container security context

mitchmckenzie:
add support for setting pod / container security context

We use opa gatekeeper to enforce pod / container security policies in our clusters.  
To make the robusta deployments compliant we need to be able to set the pod / container security context.

https://open-policy-agent.github.io/gatekeeper/website/


mitchmckenzie:
security context support added yesterday 🤣   Will review what's there and submit a new PR if additional changes are needed.

PR #1178, opened 2023-11-20
add support for setting pod / container security context

mitchmckenzie:
add support for setting pod / container security context

We use opa gatekeeper to enforce pod / container security policies in our clusters.
To make the robusta deployments compliant we need to be able to set the pod / container security context.

https://open-policy-agent.github.io/gatekeeper/website/

### DIFF
--- a/helm/robusta/templates/forwarder.yaml
+++ b/helm/robusta/templates/forwarder.yaml
@@ -28,9 +28,9 @@ spec:
       imagePullSecrets:
       {{- toYaml .Values.kubewatch.imagePullSecrets | nindent 6 }}
       {{- end }}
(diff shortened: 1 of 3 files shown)


mitch-mckenzie:
The previous PR didn't set container security context for the grafana render. I also like to to keep the templates dry by removing the else statements and moving the defaults to the default values file.

mitchmckenzie:
The previous PR didn't set the container security context for the grafana render. I also like to keep the templates dry by removing the else statements and moving the defaults to the default values file.

arikalon1:
Makes sense @mitchmckenzie , thanks for the explanation

arikalon1 (approved):
Looks good
Thanks for the PR @mitchmckenzie !

PR #1181, opened 2023-11-21
prometheus_auth is not passed to config

bear-redhat:


arikalon1 (approved):
Looks good
Thanks for the PR @bear-redhat

PR #1182, opened 2023-11-22
Pod investigator enricher: move the reason-message part from enrinchments to description

RobertSzefler:


playbooks/robusta_playbooks/pod_investigator_enricher.py
```diff
@@ -162,9 +162,17 @@ def report_pod_issue(
         event.add_enrichment(blocks)
 
     if reason:
+        # Update findings' descriptions.
         if message is None:
             message = "unknown"
-        event.add_enrichment([MarkdownBlock(f"\n\n{reason}: {message}")])
+        # The code below is essentially a specialized version of ExcutionBaseEvent.override_finding_attributes
```
arikalon1, line 168: I would put this on the parent class, and a new method `extend_description` or `add_to_description`
`def add_to_description(addition: str)`

arikalon1 (commented):
Looks good
Left one comment

playbooks/robusta_playbooks/pod_investigator_enricher.py
RobertSzefler: done

arikalon1 (approved):
LGTM

Issue #1116, opened 2023-10-09
New Sink: Webhook JSON


Tassatux:
**Is your feature request related to a problem?**
I would like to send finding information to a webhook in a JSON payload to be used by external app (for instance in a n8n workflow).

**Describe the solution you'd like**
Could we have a new generic sink webhook which send Finding attributes and other extra info (like cluster name) in JSON payload. I could allow more complex integration with a machine readable payload.

**Describe alternatives you've considered**
For now, a workaround may be to use VictorOps sink in order to have a JSON payload with some finding information and a url configurable, but we still miss some info (like severity) and the fields may change in future due to VictorOps evolution.

aantn:
Hi @Tassatux,
We have a [webhook sink](https://docs.robusta.dev/master/configuration/sinks/webhook.html#webhook) already - does it send data in the right format or do you need something else

Tassatux:
Hi,

The webhook sink only send a formatted string as body, and I would like to use the raw field from finding in a serialised way (ideally JSON), so I could use it in a programmatic way.

Tassatux:
I don't have time to work on it for now, but I'll consider to give a try for the future if nobody is able to work on it in the mean time.

I guess I have to follow https://docs.robusta.dev/master/configuration/sinks/sinks-development.html right?

aantn:
Yep, that's right. Make sure to see the instructions [here](https://docs.robusta.dev/master/setup-robusta/installation/dev-setup.html) for setting up a build environment.

Instead of writing a brand new sink, you can also consider modifying the existing webhook sink and contributing changes to that.

arikalon1:
This fix is included in 0.10.26 which was released today

PR #1184, opened 2023-11-22
feat(json-webhook): Add json option to webhook sink

shlomosfez:
Option to send webhooks in JSON format for use it programatically.

In reference to: 
https://github.com/robusta-dev/robusta/issues/1116


shlomosfez:
issues with signing on license agreement as i had a wrong email address, will try to open a new PR

PR #1188, opened 2023-11-23
Sink stop feature

RobertSzefler:
A feature that allows the sink to be configured as stopping the further dissemination of findings, if a finding is able to be delivered via that sink.


arikalon1 (approved):
LGTM

PR #1207, opened 2023-12-14
fixing bug were memory limit can not be set for forwarder.

Deepak1100:
fixing bug were the memory limit can not be set for the forwarder.


arikalon1:
This was fixed in #1296
Thanks for the PR @Deepak1100 
Closing it

PR #1210, opened 2023-12-14
Feature/support special chars dynamic routing

levtomer66:
Normalizing labels and annotations to support special characters in labels/annotations


src/robusta/core/sinks/slack/slack_sink_params.py
```diff
@@ -35,19 +33,43 @@ def validate_channel_override(cls, v: str):
                 )
                 raise ValueError(err_msg)
         return v
+    
+    def normalize_placeholder(cls, s):
+        def repl(match):
+            return cls.normalize_key_string(match.group(0))
+        
+        result = s.replace(ANNOTATIONS_PREF, '').replace(LABELS_PREF, '')
+        # Use regular expression to find and replace inside "${}"
```
arikalon1, line 42: If I understand correctly, this code assume that the parameters are surrounded by `{}`, but this is not the supported format
See the docs [here](https://docs.robusta.dev/master/configuration/sinks/slack.html#dynamic-slack-channels)

Possible values for the `channel_override` are:
* channel_override: "cluster_name"
* channel_override: "annotations.anno.slack"
* channel_override: "arik-$labels.lusha-test"

In general, it's either `cluster_name`, a label or an annotation (without the `$` sign)
Or, a string template, the the parameters with a `$` sign before the parameter

src/robusta/core/sinks/slack/slack_sink_params.py
```diff
@@ -35,19 +33,43 @@ def validate_channel_override(cls, v: str):
...
-            channel = self.channel_override.replace(LABELS_PREF, "")
-            channel = Template(channel).safe_substitute(labels)
-            channel = channel.replace(ANNOTATIONS_PREF, "")
             annots = defaultdict(lambda: MISSING)
             annots.update(annotations)
-            channel = Template(channel).safe_substitute(annots)
+
+            # Normalize channel and labels/annotations keys
+            channel_normalized = self.normalize_placeholder(channel)
+            labels_annotaions_normalized = self.normalize_dict_keys(labels | annots)
```
arikalon1, line 64: I think the code here miss the edge case, where there's some value in both annotations and labels.
For example:
If you configure:
`channel_override: "labels.channel`

And you have both a `label` channel, and an `annotation` channel

Might be better to call it separately for `labels` and `annotations` to handle this potential edge case

arikalon1 (commented):
Thanks for the PR @levtomer66 
Looks good

Left 2 comments in the review

Lastly, can you sign the CLA with both github users that committed to the PR? (or create a new PR from a single user that signed the CLA?)

src/robusta/core/sinks/slack/slack_sink_params.py
levtomer66: For long sentences that includes special characters (like kubernet.io/owner, etc...) it is crucial to use the ${} pattern.
but the old pattern still works, the logic is preserved for simple $lables.test_label.

Would you like me to add to the docs an explanation ?

src/robusta/core/sinks/slack/slack_sink_params.py
levtomer66: OK got it.

src/robusta/core/sinks/slack/slack_sink_params.py
arikalon1: Got it @levtomer66 , thanks
So if I understand correctly, if your annotation or label includes `.` or `-` or `/`, you need to use the new syntax, with the `{}`, right?
I think we need to add it to the docs, otherwise users won't know about it

The location to add it is [here](https://github.com/robusta-dev/robusta/blob/d4a4c3404208da4c323ddd87ab9f6e4ab02f0479/docs/configuration/sinks/slack.rst?plain=1#L34) somewhere 

You can build the docs locally by running:
```
poetry install -E all
./docs_autobuild.sh
```

levtomer66: Exactly, I've added to the docs

src/robusta/core/sinks/slack/slack_sink_params.py
```diff
@@ -35,19 +33,45 @@ def validate_channel_override(cls, v: str):
                 )
                 raise ValueError(err_msg)
         return v
+    
+    def normalize_placeholder(cls, s):
+        def repl(match):
+            return cls.normalize_key_string(match.group(0))
+        
+        result = s.replace(ANNOTATIONS_PREF, '').replace(LABELS_PREF, '')
```
arikalon1, line 41: Hey @levtomer66 
Do we also need to change here, and call it with only one of `ANNOTATIONS_PREF` or `LABELS_PREF` each time or it works fine as it is?

levtomer66: This should be fine as it just clear the prefixes leaving a standalone '$' for the template substitute

arikalon1: got it, thanks!

arikalon1 (approved):
Great contribution, thanks @levtomer66 !

PR #1211, opened 2023-12-15
Added check_prometheus_flags in global config.

ganeshrvel:


arikalon1 (approved):
Looks good

Please add it to the docs as well

arikalon1 (approved):
Nice work

PR #1212, opened 2023-12-17
Multi resource trigger

arikalon1:


src/robusta/core/playbooks/internal/discovery_events.py
```diff
@@ -20,17 +20,18 @@
 @action
 def cluster_discovery_updates(event: KubernetesAnyChangeEvent):
```
aantn, line 21: Can we rely on two subsequent calls to cluster_discovery_updates executing in order on the same resource?

arikalon1: I think this is the case in most cases, but in theory, the order might change
I don't think k8s resource changes are that frequent (on busy clusters we get something like 10/sec on all the cluster)

src/robusta/core/discovery/top_service_resolver.py
```diff
@@ -74,3 +74,10 @@ def add_cached_resource(cls, resource: TopLevelResource):
             cls.__recent_resource_updates[resource.get_resource_key()] = CachedResourceInfo(
                 resource=resource, event_time=time.time()
             )
+
+    @classmethod
```
aantn, line 78: Not familiar with the background, but what is the motivation for this change?

I think it'll make us more likely to send Findings where the subject is not filled in (e.g. an alert fires after the relevant resource was deleted) but that was already possible and it's not necessarily much worse.

aantn (approved):

src/robusta/core/triggers/multi_resoources_trigger.py
```diff
@@ -0,0 +1,55 @@
+from typing import Any, Dict, List
+
+from robusta.core.playbooks.base_trigger import TriggerEvent
+from robusta.integrations.kubernetes.autogenerated.triggers import KubernetesAnyAllChangesTrigger
+from robusta.integrations.kubernetes.base_triggers import K8sTriggerEvent
+
+
+class MultiResourceTrigger(KubernetesAnyAllChangesTrigger):
```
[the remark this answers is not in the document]
arikalon1: This trigger is different because it's for multiple resources
I think the benefit here with both performance and usability is great

src/robusta/core/discovery/top_service_resolver.py
arikalon1: ok, reverting it

src/robusta/core/discovery/top_service_resolver.py
```diff
@@ -74,3 +74,10 @@ def add_cached_resource(cls, resource: TopLevelResource):
...
+    def remove_cached_resource(cls, resource: TopLevelResource):
```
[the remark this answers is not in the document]
arikalon1: reverting it

PR #1215, opened 2023-12-18
mention_enricher

RobertSzefler:


playbooks/robusta_playbooks/alerts_integration.py
```diff
@@ -436,3 +437,60 @@ def foreign_logs_enricher(event: ExecutionBaseEvent, params: ForeignLogParams):
 
 
 logs_enricher = action(logs_enricher)
+
+
+class MentionParams(ActionParams):
+    static_mentions: Optional[List[str]]
+    mentions_label: Optional[str]
+    message_template: Optional[str]
```
arikalon1, line 445: I would put the default here:
`message_template: str = "Hey: $mentions"`

playbooks/robusta_playbooks/alerts_integration.py
```diff
@@ -436,3 +437,60 @@ def foreign_logs_enricher(event: ExecutionBaseEvent, params: ForeignLogParams):
...
+        if isinstance(event, PrometheusKubernetesAlert):
+            # Alert labels and annotations. FindingSubject can represent
+            # e.g. a k8s pod, job, daemonset etc etc.
+            alert_subject: FindingSubject = event.get_alert_subject()
+            event_data.update(alert_subject.annotations)
+            event_data.update(alert_subject.labels)
+            event_data.update(event.alert.annotations)
+            event_data.update(event.alert.labels)
+        elif event.obj:
+            # TODO why do we ignore alert_subject here?
```
arikalon1, line 473: We don't ignore, it's the same as the subject
you can do:
```
subject = event.get_subject()
...
```

playbooks/robusta_playbooks/alerts_integration.py
```diff
@@ -436,3 +437,60 @@ def foreign_logs_enricher(event: ExecutionBaseEvent, params: ForeignLogParams):
...
+
+        # get the mentions and use it
+        mentions_value = event_data.get(params.mentions_label)
+        if mentions_value:
+            users_list = mentions_value.split(",")
+        else:
+            users_list = []
+    elif params.static_mentions:
+        # TODO shouldn't we rather use the union of the mentions gathered above and the static ones?
+        users_list = params.static_mentions
```
arikalon1, line 487: yeah, you're right
it's more flexible to use the union

playbooks/robusta_playbooks/alerts_integration.py
```diff
@@ -436,3 +437,60 @@ def foreign_logs_enricher(event: ExecutionBaseEvent, params: ForeignLogParams):
...
+    else:
+        logging.warning("mention_enricher called with neither static_mentions nor mentions_label set")
+        return
+
+    # TODO should we additionally allow for setting messsage_template per finding subject etc as is
```
arikalon1, line 492: I think that's not necessary
We have other enrichments for custom messages
This should be only for mentions I think

arikalon1 (approved):
LGTM
Left few minor comment

playbooks/robusta_playbooks/alerts_integration.py
```diff
@@ -436,3 +437,56 @@ def foreign_logs_enricher(event: ExecutionBaseEvent, params: ForeignLogParams):
 
 
 logs_enricher = action(logs_enricher)
+
+
+class MentionParams(ActionParams):
+    static_mentions: Optional[List[str]]
```
arikalon1, line 443: I would add a parameters docs section, which would be rendered in the docs:
In the slack id example, I would put the format of real Slack ids
```
    """
    :var static_mentions: List of Slack user ids/groups ids to be mentioned
    :var mentions_label: A alert label, or Kubernetes resource label, in which the value contains a comma separated ids to mention
    :var message_template: Optional. Custom mention message. Default: `"Hey: $mentions"`

    :example static_mentions: ["slack id1", "slack id2"]
    """

```

arikalon1 (approved):
nice work

PR #1217, opened 2023-12-19
Allow overriding INITIAL_SCHEDULE_DELAY_SEC via environemt variable

DrorBuhnik:
environment variables are always string, we need to parse/cast it to int before using it


arikalon1 (approved):
Good catch
Thanks for the PR @DrorBuhnik

PR #1220, opened 2023-12-21
Sink time activation intervals

RobertSzefler:
Implements the ability to selectively activate sinks basing on configured schedules.


src/robusta/core/sinks/timing.py
```diff
@@ -0,0 +1,78 @@
...
+DAY_NAMES = list(_DAY_STR_TO_NUM.keys())
+
+
+class TimeSliceBase:
+    def is_active_now(self) -> bool:
+        raise NotImplementedError()
+
+
+class TimeSlice(TimeSliceBase):
+    def __init__(self, days: List[str] = None, time_intervals: List[Tuple[str, str]] = [], timezone=DEFAULT_TIMEZONE):
```
arikalon1, line 28: I think days shouldn't have default value
If someone calls it without days, line 29 throws an error

src/robusta/core/sinks/sink_base.py
```diff
@@ -17,6 +19,21 @@ def __init__(self, sink_params: SinkBaseParams, registry):
...
         self.signing_key = global_config.get("signing_key", "")
 
+        self.time_slice = self._build_time_slice_from_params(self.params.activity)
+
+    def _build_time_slice_from_params(self, params: ActivityParams):
+        if params is None:
+            return TimeSliceAlways()
+        else:
+            timezone = params.timezone
+            time_slice = self._interval_to_time_slice(timezone, params.intervals[0])
```
arikalon1, line 29: Would it be easier to just have an array of `time_slice`, and then, the `Sink` is_active is just:
`any([slice.is_active_now() for slice in time_slice])`

If we do this, we remove the need for `TimeSliceUnion`, which kinda not explicit (both overriding the `+` operator, and the `is_active_now`)

WDYT?

src/robusta/core/sinks/timing.py
```diff
@@ -0,0 +1,78 @@
...
+        min = int(min)
+        if not (0 <= hr <= 23 and 0 <= min <= 59):
+            raise ValueError(f"Invalid time: {time_str}")
+        return hr * 3600 + min * 60
+
+    def __add__(self, other: "TimeSlice") -> "TimeSlice":
+        return TimeSliceUnion(self, other)
+
+    def is_active_now(self) -> bool:
+        tznow = pytz.timezone("UTC").localize(datetime.utcnow()).astimezone(self.timezone)
```
arikalon1, line 56: Is it a must to get the current time in UTC then convert to our tz?
Is getting the current time directly in the target tz give the same result? (like [here](https://stackoverflow.com/questions/1398674/display-the-time-in-a-different-time-zone#:~:text=145-,A%20simpler%20method%3A,-from%20datetime%20import))

arikalon1 (approved):
Excellent work!

Left few minor comments

src/robusta/core/sinks/timing.py
RobertSzefler: Indeed!

src/robusta/core/sinks/timing.py
RobertSzefler: Indeed they are the same. Sometimes you can miss the obvious solution :D

src/robusta/core/sinks/sink_base.py
RobertSzefler: That was a good idea. Simplified the code quite a bit.

arikalon1 (approved):
Excellent work!

PR #1221, opened 2023-12-22
fix SinkBase._build_time_slices_from_params

RobertSzefler:
Fixes an important bug that arose during the refactoring of the time slice code.


arikalon1 (approved):
LGTM

Issue #1176, opened 2023-11-17
Custom CA certificate cannot be set


bear-redhat:
**Describe the bug**
Once specify a custom CA certificate with
```
runner:
  certificate:
```

The runner container will crash with:

```Matplotlib created a temporary cache directory at /tmp/matplotlib-d2tsy8fk because the default path (/.config/matplotlib) is not a writable directory; it is highly recommended to set the MPLCONFIGDIR environment variable to a writable directory, in particular to speed up the import of Matplotlib and to better support multiprocessing.
setting up colored logging
2023-11-17 22:29:20.541 INFO logger initialized using INFO log level
Traceback (most recent call last):
File "/usr/local/lib/python3.9/runpy.py", line 197, in _run_module_as_main
return _run_code(code, main_globals, None,
File "/usr/local/lib/python3.9/runpy.py", line 87, in _run_code
exec(code, run_globals)
File "/app/src/robusta/runner/main.py", line 56, in <module>
main()
File "/app/src/robusta/runner/main.py", line 26, in main
if add_custom_certificate(ADDITIONAL_CERTIFICATE):
File "/app/src/robusta/runner/ssl_utils.py", line 10, in add_custom_certificate
with open(certifi.where(), "ab") as outfile:
PermissionError: [Errno 13] Permission denied: '/usr/local/lib/python3.9/site-packages/certifi/cacert.pem'
```

**To Reproduce**
1. Set `runner.certificate` in values

**Expected behavior**
`prometheus-k8s` connection successful.

Sheeproid:
Hi @bear-redhat , how have you seen this problem solved in other projects? Do you have a workaround?

bear-redhat:
I think we need to run the Pod as root in order to access that directory.

LeaveMyYard:
Should be closed with #1222

PR #1222, opened 2023-12-25
Fix issues with Openshift

LeaveMyYard:
- Added permissions to use Openshift's Alertmanager and Prometheus to the ClusterRole
- Added automatic Openshift token usage for Alertmanager and Prometheus
- Fixed no permissions to add custom certificate (NOTE: solved without root access as proposed by @bear-redhat in #1176)
- Removed `isSmallCluster` for OpenShift example from docs (nothing works with that flag, I suppose it was just an error)
- Fixed running Popeye in Openshift
- Fixed KRR passing auth token


arikalon1 (approved):
nice work!

PR #1240, opened 2024-01-08
Fixes a null ptr exception If the configuration was applied to the va…

ganeshrvel:
…lues file before enabling it in the UI


arikalon1 (approved):
LGTM

PR #1243, opened 2024-01-09
Fix Jira transition identification

paoloyx:
There is no opened issue, I've directly opened this PR. I can create an issue if required :)

The [Jira sink documentation](https://docs.robusta.dev/master/configuration/sinks/jira.html) says that one is allowed to pass following params

> * doneStatusName : [Optional - default: 'Done'] The name of the "Done" status in Jira. Will be used to identify "Done" tasks in Jira.
>  * reopenStatusName : [Optional - default: 'To Do'] The name of the "To Do" status in Jira. Will be used to identify "To Do" tasks in Jira.
> 

So the user expects for Robusta to compare provided params (if any) with Jira **status** name, while the code instead compares that params with a **transition** name, and _it's wrong._

So the PR fixes that, it makes Robusta compare a transition "TO" status' name with a target status name, which is provided by the user

### DIFF
--- a/src/robusta/integrations/jira/client.py
+++ b/src/robusta/integrations/jira/client.py
@@ -84,7 +84,7 @@ class JiraClient:
     def _get_transition_id(self, issue_id, status_name):
         transitions = self._get_transitions_for_issue(issue_id)
         for t in transitions:
-            if t.get("name", "").lower() == status_name.lower():
+            if self._get_nested_property(t, "to.name", "").lower() == status_name.lower():
                 return t.get("id", None)
         return None
 


paoloyx:
Here is an example of a Json that identifies a Jira transition

```
{
            "id": "171",
            "name": "Resolve TODO issue",
            "to": {
                "self": "https://my.atlassian.net/rest/api/3/status/10000",
                "description": "",
                "iconUrl": "https://my.atlassian.net/images/icons/statuses/closed.png",
                "name": "Done",
                "id": "10000",
                "statusCategory": {
                    "self": "https://my.atlassian.net/rest/api/3/statuscategory/3",
                    "id": 3,
                    "key": "done",
                    "colorName": "green",
                    "name": "Done"
                }
            },
            "hasScreen": False,
            "isGlobal": False,
            "isInitial": False,
            "isAvailable": True,
            "isConditional": True,
            "isLooped": False
        }
```

arikalon1 (approved):
Looks good
Thanks for the PR @paoloyx

PR #1244, opened 2024-01-10
Allow setting of annotations on the Service Accounts

sidick:
Allow setting annotations on the created service accounts to allow easier use of IRSA.

### DIFF
--- a/helm/robusta/templates/runner-service-account.yaml
+++ b/helm/robusta/templates/runner-service-account.yaml
@@ -263,9 +263,9 @@ kind: ServiceAccount
 metadata:
   name: {{ .Release.Name }}-runner-service-account
   namespace: {{ .Release.Namespace }}
-  {{- if .Values.runner.serviceAccount.annotations }}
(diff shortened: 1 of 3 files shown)


helm/robusta/templates/runner-service-account.yaml
```diff
@@ -263,6 +263,12 @@ kind: ServiceAccount
 metadata:
   name: {{ .Release.Name }}-runner-service-account
   namespace: {{ .Release.Namespace }}
+  {{- if .Values.runner.serviceAccount.annotations }}
```
arikalon1, line 266: can you move the annotations to the `runnerServiceAccount` helm value? (see line 272 below)

arikalon1 (approved):
Thanks for the PR @sidick 
Left 1 small comment

sidick:
> Thanks for the PR @sidick Left 1 small comment

Should be fixed now

arikalon1 (approved):
Thanks for the contribution @sidick 
Looks good

PR #1247, opened 2024-01-12
Update chatgpt.rst

aantn:
This way the page will show up if people search for "AI"


pavangudiwada (approved):
LGTM

pavangudiwada:
@aantn It takes time to get indexed? 
image

aantn:
Yep. We can probably trigger a re-index on docs build if we want, but its not setup right now.

PR #1250, opened 2024-01-15
Improve events block findings

ganeshrvel:


src/robusta/core/reporting/blocks.py
```diff
@@ -492,12 +494,19 @@ def __init__(self, data: PrometheusQueryResult, query: str,
         super().__init__(data=data, metadata=metadata, vertical_lines=vertical_lines, horizontal_lines=horizontal_lines,
                          y_axis_type=y_axis_type, graph_name=graph_name)
 
+        self.set_metrics_legends_labels(legends_labels=metrics_legends_labels)
```
arikalon1, line 497: I think we don't need a class method, and don't need to add that key if the legends are None
I would just do here:
```
if metrics_legends_labels:
  self.metadata["metrics_legends_labels"] =  json.dumps(metrics_legends_labels)
```

src/robusta/core/reporting/blocks.py
```diff
@@ -492,12 +494,19 @@ def __init__(self, data: PrometheusQueryResult, query: str,
...
+
     def dict(self, *args, **kwargs) -> Dict[str, Any]:
         obj_dict = super().dict()
         obj_dict["y_axis_type"] = str(self.y_axis_type) if self.y_axis_type else None
 
         return obj_dict
 
+    def set_metrics_legends_labels(self, legends_labels: List[str]):
```
arikalon1, line 505: I think we can remove this method

playbooks/robusta_playbooks/pod_enrichments.py
```diff
@@ -81,5 +82,5 @@ def pod_node_graph_enricher(pod_event: PodEvent, params: ResourceGraphEnricherPa
     if not node:
         logging.warning(f"Node {pod.spec.nodeName} not found for pod {pod.metadata.name}")
         return
-    graph_enrichment = create_node_graph_enrichment(params, node)
+    graph_enrichment = create_node_graph_enrichment(params, node, metrics_legends_labels=["pod", "node"])
```
arikalon1, line 85: This is a node graph, I ion't think the pod is relevant here
I also think it has an `instance` label, and not a `node`
Maybe for Node graphs, we should put `instance` and `node`
But you need to verify the UI knows to handle a missing label

playbooks/robusta_playbooks/node_enrichments.py
```diff
@@ -199,5 +200,5 @@ def node_graph_enricher(node_event: NodeEvent, params: ResourceGraphEnricherPara
     Get a graph of a specific resource for this node.
     """
     node = node_event.get_node()
-    graph_enrichment = create_node_graph_enrichment(params, node)
+    graph_enrichment = create_node_graph_enrichment(params, node, metrics_legends_labels=["node"])
```
arikalon1, line 203: I think we need `instance` and `node`

ganeshrvel: done

arikalon1 (approved):
LGTM

PR #1267, opened 2024-01-29
Add priorityClassName field to helm values and forwarder/runner templates

dzliang100:


helm/robusta/values.yaml
```diff
@@ -443,6 +443,7 @@ kubewatch:
     limits:
       cpu: ~
   additional_env_vars: []
+  priorityClassName: default
```
arikalon1, line 446: Can you change the default value to "" ?
So we don't change the existing behavior

helm/robusta/values.yaml
```diff
@@ -522,6 +523,7 @@ runner:
       cpu: ~
   additional_env_vars: []
   additional_env_froms: []
+  priorityClassName: default
```
arikalon1, line 526: Same here

arikalon1 (commented):
nice and clean!
left a small comment

arikalon1 (approved):
nice work!
Thanks for the PR @dzliang100

PR #1271, opened 2024-01-30
Added GKE cloud plugin and AWS CLI

pavangudiwada:
All vendor specific Kubernetes(GKE, EKS etc) code is removed from the main Kubernetes repo after v1.26. So GKE and EKS added their own ways to authenticate when you run Kubectl.

Blog by GKE: [Changes to kubectl authentication coming in GKE v1.26 ](https://cloud.google.com/blog/products/containers-kubernetes/kubectl-auth-changes-in-gke)

This PR adds the necessary tools, `gcloud-auth-plugin` and `aws cli` for authentication when using some robusta cli options like `robusta demo` and `robusta logs`. These commands use `kubectl` in the background to perform actions.


cli.Dockerfile
```diff
@@ -1,14 +1,34 @@
+# docker build -t robusta-cli -f cli.Dockerfile .
```
aantn, line 1: Please add a comment explaining this better. E.g. "Test test this locally, run..."

aantn (approved):

PR #1274, opened 2024-02-01
ServiceNow sink

RobertSzefler:
ServiceNow sink. Includes some refactoring of MailSink as well, because the two share quite a bit of common code.

Screenshot from 2024-02-05 17-56-50


arikalon1 (approved):
Nice work!

PR #1289, opened 2024-02-14
Be able to set memory limit on robusta forwarder

michMartineau:
Currently is not possible to set a memory limit for the forwarder. The limit is set with the request value.
I would like to be able to set it as the runner: 
if memory limit is not defined, the memory request is set.


arikalon1 (approved):
nice work!

arikalon1:
Thanks for your contribution @michMartineau

PR #1296, opened 2024-02-16
Fixed a data type error caused due to prometrix library

ganeshrvel:


src/robusta/core/playbooks/prometheus_enrichment_utils.py
```diff
@@ -63,7 +63,7 @@ def __prepare_promql_query(provided_labels: Dict[Any, Any], promql_query_templat
     return promql_query
 
 
-def custom_query_range(
+def safe_custom_query_range(
```
arikalon1, line 66: Can we remove this function?
it's confusing (has the same name of the one from prometrix), and is used only once in line 102
looks redundant

arikalon1 (approved):
nice work
left one commment

src/robusta/core/playbooks/prometheus_enrichment_utils.py
ganeshrvel: done

arikalon1 (approved):
nice work

PR #1298, opened 2024-02-20
"scope" matching for sinks

RobertSzefler:
Add block/allow functionality for sinks based on the new `scope` mechanism


arikalon1 (approved):
nice work

PR #1307, opened 2024-02-23
Removed RSA pair

ganeshrvel:


arikalon1 (approved):
nice work

arikalon1:
Hi @pavangudiwada 

We removed the RSA pair from the default values. (it's not in real use)
Can you please verify that the docs are also aligned?
For example: https://docs.robusta.dev/master/setup-robusta/configuration-secrets.html#pulling-values-from-kubernetes-secrets
Thanks!

PR #1309, opened 2024-02-25
Wrap running as sub-process with an env variable

arikalon1:
When running from the ide, the server does not exit cleanly


src/robusta/core/model/env_vars.py
```diff
@@ -106,3 +106,5 @@ def load_bool(env_var, default: bool):
 IS_OPENSHIFT = load_bool("IS_OPENSHIFT", False)
 
 ENABLE_GRAPH_BLOCK = load_bool("ENABLE_GRAPH_BLOCK", False)
+
+RUN_AS_SUB_PROCESS = load_bool("RUN_AS_SUB_PROCESS", True)
```
RobertSzefler, line 110: `RUN_AS_SUBPROCESS`?

RobertSzefler (approved):
+1, just one minor naming change suggestion

PR #1310, opened 2024-02-26
feat: add securityContext on pod- and containerLevel for krr jobs

yanehi:
By default, the krr jobs should run with non-root ids and least privileges.

These are set to pod and containerLevel.


LeaveMyYard (approved):

arikalon1:
thanks for the PR @yanehi !

arikalon1:
Hi @yanehi 

I'm going to revert this fix, since the new default behavior failed on gke, eks and openshift

Can you provide an alternative solution, that the default behavior is not changed?

Maybe adding `podSecurityContext` and `containerSecurityContext` which are default to None, and used only if the user provides overriding configuration?

PR #1313, opened 2024-02-28
Improved SSH host key verification

RobertSzefler:
Support for adding custom SSH host keys via the `CUSTOM_SSH_HOST_KEYS` env variable and some related security improvements.


src/robusta/integrations/git/git_repo.py
```diff
@@ -26,6 +27,20 @@ class GitRepoManager:
     manager_lock = threading.Lock()
     repo_map = defaultdict(None)
 
+    @staticmethod
+    def setup_host_keys(custom_host_keys: List[str]):
+        if not os.path.exists(SSH_ROOT_DIR):
+            os.mkdir(SSH_ROOT_DIR)
+        with open(f"{SSH_ROOT_DIR}/known_hosts", "w") as f:
```
RobertSzefler, line 34: Reviewers note: we are recreating the `known_hosts` file here from scratch. This is ran at the start of the runner and on each subsequent config change that might affect the list of known hosts. Is this correct? I think so, as there are no other places in the codebase where `known_hosts` are modified and as far as I understand, with a new deployment it is not even created in the first place.

src/robusta/runner/config_loader.py
```diff
@@ -105,6 +106,7 @@ def __load_playbooks_repos(
         playbooks_repos: Dict[str, PlaybookRepo],
     ):
         playbook_packages = []
+        GitRepoManager.setup_host_keys(CUSTOM_SSH_HOST_KEYS.split("\n"))
```
arikalon1, line 109: I think create a change of behavior I'm a little concerned with
 
Up until now, writing the known_hosts `/root` was happening only for user that defined a `git_key`
With this change it happens to all users

I know, that on some clusters we have limited access to disk, and might not have access to `/root/`
This will fail for them when the runner starts

Can we change it, to run once, if there actually `git_key` configured?

arikalon1 (commented):
nice work
left some comments

arikalon1 (approved):
nice work

PR #1320, opened 2024-03-04
Support extra customplaybooks

tip-dteller:
It proves to be difficult to manage everything via Generated_values.yaml file.
even if splitting away into a separate file. Its difficult to know which actions and triggers are in use by which department/team.

This PR allows to parse "customExtraPlaybooks" which is taking in a dictionary. Loops through it and adds into CustomPlaybooks field.


helm/robusta/values.yaml
```diff
@@ -147,6 +147,30 @@ priorityBuiltinPlaybooks:
...
+#         aggregation_key: "This is app-a - Requires your attention"
+#         severity: HIGH
+#         title: "Check app-a out"
+#         description: "@monitoring.monitoring this is for you"
+#   team-b-app-b:
+#   # Default Settings for all actions
+#   - triggers:
+#     - on_prometheus_alert:
+#         namespace_prefix: "app-b"
+customExtraPlaybooks: {}
```
arikalon1, line 171: Would it be better to rename it to `namedCustomPlaybooks`?
This way, the name hints the usage

arikalon1 (commented):
thanks for the PR @tip-dteller 
nice work

Left one small comment

helm/robusta/values.yaml
tip-dteller: renamed as suggested

arikalon1 (approved):

PR #1325, opened 2024-03-07
updated msteams instructions

Sheeproid:
replaced the old video with a gif and a swift explanation, also instructions fit the new MSTeams UI

see https://docs.robusta.dev/updated-msteams-docs/configuration/sinks/ms-teams.html


aantn (approved):
LGTM

PR #1328, opened 2024-03-08
Fix invalid adaptive card body for Webex sink

ppmathis:
This PR fixes an issue with the currently used adaptive card body for the Webex sink. I installed the latest Robusta version 0.10.30 in a test environment of mine and noticed an issue once I configured a Webex sink, specifically that the notification messages only show up properly in the web client for Cisco Webex. When attempting to view the messages in the desktop clients (Windows / macOS), the app only showed the placeholder text message (`.`) without any other data.

**Screenshot from Webex Web:**
image

**Screenshot from Webex Desktop (macOS):**
image

So I started analysing this issue by dumping the adaptive card data into a file and messing around with it, until I discovered that the sent payload is violating the official Adaptive Cards v1.2 spec as the property `wrap` is set as a string with a boolean inside, instead of an actual boolean. So the JSON body should contain `"wrap": true`, not `"wrap": "true"` like it does today.

While the Webex web client still properly displays such semi-broken cards, the desktop client straight up refuses without any useful information. Simply switching out the couple lines of code to pass `True` instead of `"true"` fully fixes the issue on both platforms.

### DIFF
--- a/src/robusta/integrations/webex/sender.py
+++ b/src/robusta/integrations/webex/sender.py
@@ -75,7 +75,7 @@ class WebexSender:
                     {
                         "type": "Column",
                         "width": "stretch",
-                        "items": [{"type": "TextBlock", "text": header, "wrap": "true"}],
+                        "items": [{"type": "TextBlock", "text": header, "wrap": True}],
                     }
                 )
             # seperating each row to add below headers of column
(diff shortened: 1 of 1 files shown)


arikalon1 (approved):
Nice find!
Thank you for the PR @ppmathis

PR #1333, opened 2024-03-11
Python debugging tools update and deprecation

Avi-Robusta:
updating debugger tools image
only running on python3 processes since debugger tools does not seem to work on older python versions 
Additionally python debugging is now deprecated and will be removed in the future until the python debug-toolkit project is improved to return results more consistently.
https://github.com/robusta-dev/debug-toolkit/pull/5


arikalon1 (approved):
nice work

PR #1334, opened 2024-03-11
Seperated Openshift from supported clusters page

pavangudiwada:


docs/setup-robusta/supported-clusters.rst
```diff
@@ -1,96 +1,22 @@
 Supported Clusters
 ################################
 
-We support all Kubernetes distributions other than Minikube.
+Robusta supports all Kubernetes distributions other than Minikube. We tested it on the following clusters
```
aantn, line 4: > We tested it on the following clusters

This seems a little awkward. It implies we tested Robusta just once in the past on these. Perhaps:

> Robusta supports all Kubernetes distributions other than Minikube, including:

docs/setup-robusta/supported-clusters.rst
```diff
@@ -1,96 +1,22 @@
...
+* KIND
+* RKE
+* :ref:`Openshift <Openshift>`
 
 Minikube
 ==========
 We don't recommend installing Robusta on Minikube, due to a Minikube bug. For more details, refer to `this GitHub issue <https://github.com/kubernetes/minikube/issues/14806>`_.
 
-We recommend testing with KIND instead.
+Instead, we recommend testing with KIND.
```
aantn, line 19: > Instead, we recommend testing with KIND.

Should be:

> Instead, we recommend testing Robusta with KIND.

(Otherwise "it" is ambigious and could refer to Minikube for example.)

docs/setup-robusta/openshift.rst
```diff
@@ -0,0 +1,68 @@
+.. _openshift-permissions:
+
+OpenShift
+========================================
+
+OpenShift is supported via the Helm chart.
```
aantn, line 6: What does this mean? Perhaps phrase as:

Robusta supports OpenShift. To use Robusta on OpenShift, you should enable the following Helm values:

...

And I'd put the warning below after the Helm values. Lead with the main thing, then add the caveat.

aantn (approved):
LGTM. Left minor comments. Please fix and then merge.

PR #1338, opened 2024-03-12
Openshift dc hikaru variant

RoiGlinik:


src/robusta/integrations/kubernetes/autogenerated/events.py
```diff
@@ -55,6 +55,7 @@
     RobustaDeployment,
     RobustaJob,
     RobustaPod,
+    DeploymentConfig,
```
arikalon1, line 58: This file is auto generated
Need to think how to include the DeploymentConfig, which is special because it's not a hikaru builtin resource

arikalon1 (commented):
nice work!
left one comment

arikalon1 (approved):
nice work

PR #1358, opened 2024-03-25
Scope filtering for Prometheus alerts

RobertSzefler:
`scope` filtering for Prometheus alerts.

Includes Sphinx-related dependency version bump to fix documentation build issues.


src/robusta/integrations/prometheus/trigger.py
```diff
@@ -11,10 +11,27 @@
...
+class PrometheusTriggerEventScopeMatcher(BaseScopeMatcher):
+    def __init__(self, data):
+        # We put labels data both on the top level and under the "labels" key,
+        # so that keys like alertname, severity etc. are accessible without
+        # labels: ... matching.
+        self.data = data.labels.copy()
+        self.data["annotations"] = data.annotations
+        self.data["labels"] = data.labels.copy()
+        # status is the only field of PrometheusAlert that we want to expose.
+        self.data["status"] = data.status
```
arikalon1, line 29: Not sure we want the `status` here, because we have the `status` filter on the trigger

docs/playbook-reference/triggers/prometheus.rst
```diff
@@ -29,7 +29,9 @@ The following triggers are available for Prometheus alerts:
             customPlaybooks:
             - triggers:
               - on_prometheus_alert:
-                  alert_name: HostHighCpuLoad
+                  scope:
```
arikalon1, line 32: I think this example is not correct. 
I think the scope, supports only `labels` and `annotations` (and maybe `status`)
So maybe the example should be:
```
            customPlaybooks:
            - triggers:
              - on_prometheus_alert:
                  alert_name: HostHighCpuLoad
                  scope:
                    include::
                    - labels:
                       - "deployment=nginx"
```

I think it's important to explain the scope only support `labels` and `annotations`

docs/playbook-reference/triggers/prometheus.rst
```diff
@@ -38,6 +40,10 @@ The following triggers are available for Prometheus alerts:
 
     .. pydantic-model:: robusta.integrations.prometheus.trigger.PrometheusAlertTrigger
 
+    The ``scope`` filtering mechanism works exactly as it does for sinks
+    (see :ref:`sink-scope-matching` for more information), but the elements that can
+    be matched are the ones available in Prometheus alert payload.
```
arikalon1, line 45: here I would say it's only alert `labels` and `annotations`

arikalon1 (commented):
nice work
left few comments

src/robusta/integrations/prometheus/trigger.py
RobertSzefler: refactored as discussed

docs/playbook-reference/triggers/prometheus.rst
RobertSzefler: changed wording

arikalon1 (approved):
nice work!

PR #1361, opened 2024-03-27
Feat/add zulip sink

oscgu:
This PR adds a Zulip sink, makes the `channel_override` functionality globally available and the local runner script a bit more *robust*


oscgu:
I'll fix the tests an re-open

PR #1363, opened 2024-03-27
Feat/add zulip sink

oscgu:
Changes in this PR:

- New Zulip sink
- `channel_override` functionality now globally available through the `ChannelTransformer` class + rewrote the tests which previously only targeted Slack
- Made local runner script more *robust*


src/robusta/core/sinks/slack/slack_sink_params.py
```diff
@@ -28,86 +10,6 @@ class SlackSinkParams(SinkBaseParams):
     channel_override: Optional[str] = None
     max_log_file_limit_kb: int = 1000
 
-    @validator("channel_override")
```
arikalon1, line 31: hi @oscgu 
why are we changing the Slack parameters?

oscgu: Hi @arikalon1,
I moved out the `channel_override` into `src/robusta/core/sinks/common/channel_transformer.py`so that it can be reused accross sinks like so:

```python
 overriden_channel = ChannelTransformer.template(
                sink_params.channel_override,
                sink_params.slack_channel,
                self.cluster_name,
                finding.subject.labels,
                finding.subject.annotations,
            )
```
(Here in the slack sink: https://github.com/robusta-dev/robusta/pull/1363/files#diff-05a4c5c486a82e25806bd8351699c08cf038425841d6d55e9fdbd3b6f099b776R380)

arikalon1: hey @oscgu 

I think all the parameters validators were deleted from the SlackSinkParams
the method `get_slack_channel` and some other methods

arikalon1: I understand the reason for moving the `get_slack_channel` logic
but why do we need to remove the validation?

oscgu: @arikalon1 the validation of the `channel_override` param has also been moved to the `ChannelTransformer` , see [here](https://github.com/robusta-dev/robusta/pull/1363/files/1bc313fe73c5e32af5dc57f79ac5f0fc96133de6#diff-9b26b848df0cca0c340d005d0ea7b912f92fa00b2911045341b54437c047f7abR78).
Should that rather be done in the sinks themselves, with the validator annotation?

oscgu: I guess it should happen in the sink or else it will get re-validated with each invocation of the template func, even though it never changes

arikalon1: yeah, I agree

I think the validation should happen only once, when creating the params object

This is also create after a configuration change, and an error at this phase cause the pod to crash, and the user should notice the configuration error

src/robusta/core/sinks/common/channel_transformer.py
```diff
@@ -0,0 +1,94 @@
...
+        annotations: Dict[str, str],
+    ) -> str:
+        if not channel_override:
+            return default_channel
+
+        channel = channel_override
+        if CLUSTER_PREF in channel:
+            # replace "cluster_name" or "$cluster_name" or ${cluster_name} with the value of the cluster name
+            channel = CLUSTER_PREF_PATTERN.sub(cluster_name, channel)
+
```
arikalon1, line 84: Since the below string manipulations might be expensive, performance wise, and are performed for each message, maybe it's worth leaving lines 85-92 under this `if`:
`if LABELS_PREF in channel or ANNOTATIONS_PREF in channel:`

We can skip it all, if the above is false

arikalon1 (commented):
Thanks for the PR @oscgu 

Really nice work!
Left 2 small comments, that might improve performance a llittle
WDYT?

src/robusta/core/sinks/common/channel_transformer.py
```diff
@@ -0,0 +1,94 @@
...
+        normalized_labels = cls.normalize_dict_keys(labels)
+        normalized_annotations = cls.normalize_dict_keys(annotations)
+
+        channel = cls.replace_token(BRACKETS_PATTERN, LABELS_PREF, channel, normalized_labels)
+        channel = cls.replace_token(BRACKETS_PATTERN, ANNOTATIONS_PREF, channel, normalized_annotations)
+
+        channel = cls.replace_token(LABEL_PREF_PATTERN, LABELS_PREF, channel, normalized_labels)
```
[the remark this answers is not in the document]
oscgu: Good catch!
I have combined your suggestions, because we don't need the normalized_annotations in the labels block and same thing for the other way around.
72486a12dbadd1d36d52582d36ac547e5131843a

src/robusta/core/sinks/common/channel_transformer.py
oscgu: 72486a12dbadd1d36d52582d36ac547e5131843a

arikalon1 (approved):
Thanks for the PR @oscgu

PR #1367, opened 2024-04-01
Added domains used by Robusta

pavangudiwada:


aantn (approved):
@pavangudiwada can you add a clarification that this is for the SaaS UI to function. (No need to communicate with these domains if you don't use the UI or if you use a self-hosted UI.)

Right now the language added would seem to contradict what is written above ("Robusta can run in airgapped environments") but it doesn't actually contradict.

PR #1369, opened 2024-04-07
added additional configs to prometheus test alert

Avi-Robusta:
changed "firing" to resolved since our alerts are test alerts were automatically resolved by default because there was an end time 
so I wanted to preserve the original behavior for our demo alerts


arikalon1 (approved):
LGTM

PR #1374, opened 2024-04-10
make krr_scan action available for mattermost sink

IdeoG:
https://robustacommunity.slack.com/archives/C02R0LVANKY/p1712750565337999

### DIFF
--- a/src/robusta/integrations/mattermost/sender.py
+++ b/src/robusta/integrations/mattermost/sender.py
@@ -11,6 +11,7 @@ from robusta.core.reporting.blocks import (
     KubernetesDiffBlock,
     ListBlock,
     MarkdownBlock,
+    ScanReportBlock,
     TableBlock,
 )
 from robusta.core.reporting.utils import add_pngs_for_all_svgs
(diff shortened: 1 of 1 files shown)


src/robusta/integrations/mattermost/sender.py
```diff
@@ -11,6 +11,7 @@
     KubernetesDiffBlock,
     ListBlock,
     MarkdownBlock,
+    ScanReportBlock,
     TableBlock,
 )
 from robusta.core.reporting.utils import add_pngs_for_all_svgs
```
arikalon1, line 17: I think there's a missing import here
`from robusta.core.reporting.consts import EnrichmentAnnotation`

Can you please add it?

src/robusta/integrations/mattermost/sender.py
```diff
@@ -144,7 +147,10 @@ def send_finding_to_mattermost(self, finding: Finding, platform_enabled: bool):
             blocks.append(MarkdownBlock(finding.description))
 
         for enrichment in finding.enrichments:
-            blocks.extend(enrichment.blocks)
+            if enrichment.annotations.get(EnrichmentAnnotation.SCAN, False):
```
arikalon1, line 150: I think here it should be:
```
            if enrichment.annotations.get(EnrichmentAnnotation.SCAN, False):
                enrichment.blocks = [Transformer.scanReportBlock_to_fileblock(b) for b in enrichment.blocks]
            
            blocks.extend(enrichment.blocks)
```

The `blocks` array is the collection sent to mattermost later on, so we want to add the converted file to it

arikalon1 (commented):
Thanks for the PR @IdeoG 
Left small comments

I don't have a mattermost account, so I did not test it

src/robusta/integrations/mattermost/sender.py
IdeoG: You're right. Got it!

arikalon1 (approved):
Looks good
Thanks for the contribution @IdeoG

PR #1377, opened 2024-04-11
Update change tracking tutorial

aantn:
@pavangudiwada I think we might need some more changes here too. Specifically:

1. the example for tracking deployment changes speaks about tracking only image changes, but actually it tracks much more changes
2. the example for ingress speaks about port or path changes but also tracks much more

Can you please add more changes to this PR so the tutoral is more accurate.


pavangudiwada (approved):
LGTM

PR #1378, opened 2024-04-12
Custom triggers: fix support for scope

RobertSzefler:
Adds missing support for the `scope` parameter for some custom triggers


arikalon1 (approved):
nice work

PR #1382, opened 2024-04-15
make use of all object data on attributes matcher

arikalon1:
add log of data if attribute not found, to help troubleshoot


RobertSzefler (approved):
LGTM

PR #1385, opened 2024-04-18
Be consistent with other locations

aantn:
Name it "ImagePullBackOff" not "image-pull-back-off" etc

### DIFF
--- a/playbooks/robusta_playbooks/pod_investigator_enricher.py
+++ b/playbooks/robusta_playbooks/pod_investigator_enricher.py
@@ -137,7 +137,7 @@ def get_pod_issue_explanation(event: KubernetesResourceEvent, issue: PodIssue, m
     elif issue == PodIssue.Pending:
         issue_text = "scheduling issue"
     elif issue in [PodIssue.CrashloopBackoff, PodIssue.Crashing]:
-        issue_text = "CrashLoopBackOff"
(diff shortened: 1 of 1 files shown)


playbooks/robusta_playbooks/pod_investigator_enricher.py
```diff
@@ -133,11 +133,11 @@ def get_pod_issue_explanation(event: KubernetesResourceEvent, issue: PodIssue, m
     resource = event.get_resource()
 
     if issue == PodIssue.ImagePullBackoff:
-        issue_text = "image-pull-backoff"
+        issue_text = "ImagePullBackoff"
     elif issue == PodIssue.Pending:
         issue_text = "scheduling issue"
     elif issue in [PodIssue.CrashloopBackoff, PodIssue.Crashing]:
-        issue_text = "crash-looping"
+        issue_text = "CrashLoopBackOff"
```
arikalon1, line 140: I think it should be `CrashloopBackoff` (with a lower case `o` on the `off`)
 It's important, because the matcher are case sensitive atm (maybe we need to change it)
 
 Would it be better to just use the value from the enum? `PodIssue.CrashloopBackoff.value` & `PodIssue.ImagePullBackoff.value`

arikalon1 (approved):
nice catch 
left a small comment

PR #1388, opened 2024-04-19
Add `nameOverride` and `fullnameOverride` in helm chart

kristeey:
Add possibility to overwrite naming in helm chart adding `nameOverride` and `fullnameOverride` helpers. 

These are commonly used helpers and helpful in scenarios where the robusta helm chart is used as a subchart. Without the opportunity to override the naming -> the robusta manifest will be enriched with the release name of the parent chart.

### DIFF
--- a/helm/robusta/templates/_helpers.tpl
+++ b/helm/robusta/templates/_helpers.tpl
@@ -14,12 +14,7 @@ If release name contains chart name it will be used as a full name.
 {{- if .Values.fullnameOverride }}
 {{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" }}
 {{- else }}
-{{- $name := default .Chart.Name .Values.nameOverride }}
-{{- if contains $name .Release.Name }}
 {{- .Release.Name | trunc 63 | trimSuffix "-" }}
-{{- else }}
-{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" }}
-{{- end }}
 {{- end }}
 {{- end }}
 
--- a/helm/robusta/templates/_helpers.tpl
+++ b/helm/robusta/templates/_helpers.tpl
@@ -7,7 +7,7 @@ If release name contains chart name it will be used as a full name.
 {{- if .Values.fullnameOverride }}
 {{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" }}
 {{- else }}
-{{- .Release.Name | trunc 63 | trimSuffix "-" }}
+{{- .Release.Name | trunc 63 }}
 {{- end }}
 {{- end }}
 
--- a/helm/robusta/templates/_helpers.tpl
+++ b/helm/robusta/templates/_helpers.tpl
@@ -1,3 +1,28 @@
+{{/*
+Expand the name of the chart.
+*/}}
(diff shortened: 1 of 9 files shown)


arikalon1 (commented):
thanks for the PR @kristeey 

I understand the need to override the release name for some users, but can you make this in a way that doesn't change the behavior to anyone that didn't configure the overrides?

In order words, if the user didn't configure the overrides, that chart would use the release name, like before

WDYT?

kristeey:
I belive this is already covered by my PR. Look at the helper template codelines: https://github.com/robusta-dev/robusta/blob/e31d6db8a2e3ee9a6ab1f413422411843044c64a/helm/robusta/templates/_helpers.tpl#L17-L21 

If `fullnameOverride` and/or `nameOverride` isn't specified the `robusta.fullname` template is just `.Release.name`

arikalon1:
Maybe I'm reading it wrong..
`{{- $name := default .Chart.Name .Values.nameOverride }}` - here name will be the Chart name
 
 then
` {{- if contains $name .Release.Name }} `
 and here we check if the Chart name is in the Release name?
 
 The chart name is probably fixed, and the release name can be anything, isn't it?

kristeey:
Yes. The `.Chart.Name` is static and determined by the `name` value in `Chart.yaml`.  It then checks if the name variable (consisting og the chart name + nameOverride value) is a substring of the release name.  If it is: `robusta.fullname` will be the release name. If it is not a substring `robusta.fullname` will be the release name + name variable. But if `nameOverride` is empty, `robusta.fullname` will still just be the release name. 

This is a common practice for helm charts, some examples are:

- [opentelemetry-operator](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/45400c8536c0f87d53731bd3e696b88eabb2f78f/charts/opentelemetry-operator/templates/_helpers.tpl#L8-L24)
- [external-dns](https://github.com/kubernetes-sigs/external-dns/blob/c506a20de6841636c1f6654c9436402cc461bb2c/charts/external-dns/templates/_helpers.tpl#L8-L24)
- [cert-manager](https://github.com/cert-manager/cert-manager/blob/76153b3bf2262f39a6a19aebceed63d53511968b/deploy/charts/cert-manager/templates/_helpers.tpl#L9-L24) 

It is easy to verify the output with this command 
```
helm install my-release-name ./robusta -f values.yaml --dry-run
```
playing around with the release name and the `fullnameOverride` and/or `nameOverride` values.

kristeey:
Well, 
There is however some cases where this will break as you point out. E.g. if someone has a completely different release name e.g. `some-release` which will result in `some-release-robusta`...

A possibility is to remove the `nameOverride` value, as fullnameOverride will be sufficient in my usecase.

arikalon1:
can you please simplify the logic, to something like:
```
{{- define "robusta.fullname" -}}
{{- if .Values.fullnameOverride }}
.... all the logic for the override
{{- else }}
{{- .Release.Name | trunc 63 }}
{{- end }}
{{- end }}
```

So if someone doesn't specify the `fullnameOverride` nothing changes

arikalon1 (approved):
nice work!

Thanks for the contribution @kristeey

PR #1390, opened 2024-04-22
namespace_labels matching for the sink scope mechanism

RobertSzefler:
Implements the matching of labels assigned to the namespace for the sink scope mechanism.

Includes some fixes in the k8s trigger autogeneration code.


scripts/generate_kubernetes_code.py
```diff
@@ -376,6 +378,13 @@ def autogenerate_triggers(f: TextIO):
     for resource in KUBERNETES_RESOURCES:
         f.write(f"# {resource} Triggers\n")
         for trigger_name, operation_type in sorted(TRIGGER_TYPES.items()):
+            if f"{resource}{get_trigger_class_name(trigger_name)}" in [
```
arikalon1, line 381: can you explain this?

src/robusta/integrations/kubernetes/api_client_utils.py
```diff
@@ -264,3 +268,21 @@ def parse_kubernetes_datetime_to_ms(k8s_datetime: str) -> float:
...
+# so we don't bother to pick what we exactly need (currently just labels - see below in
+# get_namespace_labels) and just cache all namespace data.
+@cached(cache=TTLCache(maxsize=1, ttl=NAMESPACE_DATA_TTL))
+def get_all_namespace_data():
+    logging.info("(re)loading all namespace data")
+    core_v1 = core_v1_api.CoreV1Api()
+    try:
+        return {ns.metadata.name: ns.metadata for ns in core_v1.list_namespace().items}
+    except ApiException:
+        logging.error("failed to list namespaces")
```
arikalon1, line 283: please log the stack trace

src/robusta/utils/scope.py
```diff
@@ -59,12 +54,12 @@ def scope_matches(self, scope: ScopeIncludeExcludeParamsT) -> bool:
...
         data = self.get_data()
         if attr_name not in data:
-            logging.warning(f'Scope match on non-existent attribute "{attr_name}" ({data=}')
+            logging.warning(f'Scope match on non-existent attribute "{attr_name}" ({data=})')
             return False
         attr_value = data[attr_name]
         for attr_matcher in attr_matchers:
-            if attr_name == "attributes":
-                return self.scope_match_attributes(attr_matcher, attr_value)
+            if hasattr(self, f"match_{attr_name}"):
```
arikalon1, line 61: please leave it more explicit

I think it's best to load the namespace labels only here, of there's actually a matcher that uses it
It will be used on a fraction of the users only

src/robusta/core/reporting/base.py
```diff
@@ -169,7 +173,12 @@ def matches(self, match_requirements: Dict[str, Union[str, List[str]]], scope_re
         # 1. "scope" check
         accept = True
         if scope_requirements is not None:
-            matcher = FilterableScopeMatcher(self.attribute_map)
+            data = self.attribute_map
```
arikalon1, line 176: I think it would be best to load and cache the namespaces, only if there's a matcher that uses it

arikalon1 (commented):
nice work
left few comments

src/robusta/integrations/kubernetes/api_client_utils.py
RobertSzefler: done

scripts/generate_kubernetes_code.py
RobertSzefler: As was discussed previously, only the event types listed here should have `change_filters` set to `DEFAULT_CHANGE_FILTERS`, the reason being change tracking for the other events makes no sense, right?

I think this change was previously made to the code under `src/robusta/integrations/kubernetes/autogenerated/`, and all the changes in `generate_kubernetes_code.py` in this PR are introduced just to make sure the autogenerated code after running the script is exactly the same as the code in the repo.

src/robusta/utils/scope.py
RobertSzefler: That way we would be accumulating abstract methods in the base class, each of which is implemented in some random derived class. Are you sure we want to go that way? The mechanism reintroduced here, `match_{whatever}` IMO elegantly implements the matching of attributes that is not explicitly implemented in the base class, basing on attribute name.

src/robusta/core/reporting/base.py
RobertSzefler: as discussed, always listing the namespaces does not incur an unacceptable time penalty, so we leave it

src/robusta/utils/scope.py
RobertSzefler: Okay, for reasons of clarity changed the `scope_match_*` methods/calls back to be explicit.

scripts/generate_kubernetes_code.py
RobertSzefler: As discussed, my reasoning here was actually wrong and all of these classes should have `change_filters` default set to `None`, which has been fixed.

arikalon1 (approved):
nice work

PR #1391, opened 2024-04-24
On busy clusters, krr pod might be deleted before the runner is getti…

arikalon1:
…ng the logs output

Added finalizer that prevents the pod from being deleted

### DIFF
--- a/src/robusta/integrations/kubernetes/custom_models.py
+++ b/src/robusta/integrations/kubernetes/custom_models.py
@@ -457,7 +457,7 @@ class RobustaJob(Job):
         """
         pods = self.get_pods()
         if len(pods) != 1:
-            raise Exception(f"got more pods than expected for job: {pods}")
+            raise Exception(f"got more pods than expected for job {self.metadata.name}: {pods}")
         return pods[0]
 
     def create_job_owned_secret(self, job_secret: JobSecret):
@@ -494,19 +494,27 @@ class RobustaJob(Job):
         ttl_seconds_after_finished: int = 120,
         delete_job_post_execution: bool = True,
         process_name: bool = True,
+        finalizers: Optional[
+            List[str]
+        ] = None,  # Finalizers are used to verify the pod is not deleted before getting the logs
     ) -> str:
+        pod_meta = ObjectMeta(annotations=custom_annotations)
+        if finalizers:
+            pod_meta.finalizers = finalizers
+        job_name = to_kubernetes_name(name) if process_name else name
         job = RobustaJob(
             metadata=ObjectMeta(
                 namespace=INSTALLATION_NAMESPACE,
-                name=to_kubernetes_name(name) if process_name else name,
+                name=job_name,
                 annotations=custom_annotations,
             ),
             spec=JobSpec(
                 backoffLimit=0,
-                template=PodTemplateSpec(spec=spec, metadata=ObjectMeta(annotations=custom_annotations)),
+                template=PodTemplateSpec(spec=spec, metadata=pod_meta),
                 ttlSecondsAfterFinished=ttl_seconds_after_finished,
             ),
         )
+        pod = None
         try:
             job = job.createNamespacedJob(job.metadata.namespace).obj
             job = hikaru.from_dict(job.to_dict(), cls=RobustaJob)  # temporary workaround for hikaru bug #15
@@ -517,6 +525,16 @@ class RobustaJob(Job):
             pod = job.get_single_pod()
             return pod.get_logs() or ""
         finally:
+            if pod and finalizers:
+                try:  # must use patch, since the pod revision changed at this point
+                    client.CoreV1Api().patch_namespaced_pod(
+                        name=pod.metadata.name,
+                        namespace=pod.metadata.namespace,
+                        body=[{"op": "remove", "path": "/metadata/finalizers"}],
+                    )
+                except Exception:
+                    logging.exception(f"Failed to clear pod finalizers for {job_name}")
+
             if delete_job_post_execution:
                 job.deleteNamespacedJob(
                     job.metadata.name,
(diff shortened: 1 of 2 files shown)


src/robusta/integrations/kubernetes/custom_models.py
```diff
@@ -457,7 +457,7 @@ def get_single_pod(self) -> RobustaPod:
         """
         pods = self.get_pods()
         if len(pods) != 1:
-            raise Exception(f"got more pods than expected for job: {pods}")
+            raise Exception(f"got more pods than expected for job {self.metadata.name}: {pods}")
```
aantn, line 460: got more/less

src/robusta/integrations/kubernetes/custom_models.py
```diff
@@ -517,6 +525,21 @@ def run_simple_job_spec(
...
             return pod.get_logs() or ""
         finally:
+            if pod and finalizers:
+                try:  # must use patch, since the pod revision changed at this point
+                    pod_finalizers = pod.metadata.finalizers
+                    for finalizer in finalizers:
+                        if finalizer in pod_finalizers:
+                            pod_finalizers.remove(finalizer)
+
+                    client.CoreV1Api().patch_namespaced_pod(
```
aantn, line 535: There is a race condition here if another controller put finalizers and removes them while this runs.

Is it possibleto do a PATCH w/ `op: remove` and explicitly remove only the finalizers that we set w/o impacting others?

aantn (approved):
Left some comments

src/robusta/integrations/kubernetes/custom_models.py
arikalon1: I don't know if you can do that in a patch
do you?

aantn:
Yes, the syntax is a little obscure and poorly documented, but you can do
something like:

```
patch = {
            "metadata": {
"$deleteFromPrimitiveList/finalizers": ["finalizer.example.com/foo"]
            }
        }
```

Here is a full test case:

```
from kubernetes import client, config
from kubernetes.client.rest import ApiException

def main():
    config.load_kube_config()
    api_instance = client.CoreV1Api()

    pod_manifest = {
        "apiVersion": "v1",
        "kind": "Pod",
        "metadata": {
            "name": "example-pod",
            "finalizers": ["finalizer1.example.com/bar", "
finalizer2.example.com/foo"]
        },
        "spec": {
            "containers": [{
                "name": "nginx",
                "image": "nginx:1.19.0"
            }]
        }
    }

    try:
        api_instance.create_namespaced_pod(namespace="default",
body=pod_manifest)
        print("Pod created with two finalizers.")
    except ApiException as e:
        print(f"Exception when creating a pod: {e}\n")

    patch = {
            "metadata": {
"$deleteFromPrimitiveList/finalizers": ["finalizer2.example.com/foo"]
            }
        }
    try:
        x=api_instance.patch_namespaced_pod(name="example-pod",
namespace="default", body=patch)
        print("One finalizer removed from pod.")
        print(x)
    except ApiException as e:
        print(f"Exception when patching the pod to remove a finalizer:
{e}\n")

if __name__ == '__main__':
    main()
```

src/robusta/integrations/kubernetes/custom_models.py
aantn: Yes, the syntax is a little obscure and poorly documented, but you can do
something like:

```
patch = {
     "metadata": {
         "$deleteFromPrimitiveList/finalizers": ["finalizer.example.com/foo"]
     }
}
```

Here is a full test case:

```
from kubernetes import client, config
from kubernetes.client.rest import ApiException

def main():
    config.load_kube_config()
    api_instance = client.CoreV1Api()

    pod_manifest = {
        "apiVersion": "v1",
        "kind": "Pod",
        "metadata": {
            "name": "example-pod",
            "finalizers": ["finalizer1.example.com/bar", "
finalizer2.example.com/foo"]
        },
        "spec": {
            "containers": [{
                "name": "nginx",
                "image": "nginx:1.19.0"
            }]
        }
    }

    try:
        api_instance.create_namespaced_pod(namespace="default",
body=pod_manifest)
        print("Pod created with two finalizers.")
    except ApiException as e:
        print(f"Exception when creating a pod: {e}\n")

    patch = {
            "metadata": {
                 "$deleteFromPrimitiveList/finalizers": ["finalizer2.example.com/foo"]
            }
        }
    try:
        x=api_instance.patch_namespaced_pod(name="example-pod", namespace="default", body=patch)
        print("One finalizer removed from pod.")
        print(x)
    except ApiException as e:
        print(f"Exception when patching the pod to remove a finalizer:
{e}\n")

if __name__ == '__main__':
    main()
```

arikalon1: are you sure this is a part of the API and not something that will break in future versions?

aantn: It is documented [here](https://github.com/kubernetes/community/blob/2f8415788e5a50dec90383408185b535b0e55b65/contributors/devel/sig-api-machinery/strategic-merge-patch.md?plain=1#L177) and is the idiomatic way (as far as I can tell) to remove finalizers. Here is an [official code sampe](https://github.com/kubernetes-client/java/blob/fd530a9126e0ffe80d918ccd7b49b00aa4dde382/examples/examples-release-20/src/main/java/io/kubernetes/client/examples/PatchExample.java#L47) and [here is code from the mysql operator using it](https://github.com/mysql/mysql-operator/blob/4a80d27486c36a1ba1262d79cddbb99be21e52ba/mysqloperator/controller/innodbcluster/cluster_api.py#L2330). It is also widely used internally in the Kubernetes codebase itself.

So I think it is safe to use and better than risking the race condition.

arikalon1: I don't think there's any real chance of race condition
but it's it's officially supported, it's a better way to do it

PR #1394, opened 2024-04-29
Fix secrets interpolation for KRR

LeaveMyYard:
The syntax for kubernetes is `$(ENV_NAME)` and not `$ENV_NAME`


arikalon1 (approved):
very nice

PR #1396, opened 2024-05-02
Deduplicate alert manager alerts by caching their hashes

RobertSzefler:


src/robusta/runner/web.py
```diff
@@ -57,11 +69,40 @@ def handle_alert_event():
         alert_manager_event = AlertManagerEvent(**req_json)
         for alert in alert_manager_event.alerts:
             alert = Web._relabel_alert(alert)
-            Web.alerts_queue.add_task(Web.event_handler.handle_trigger, PrometheusTriggerEvent(alert=alert))
+            alert_hash = Web.get_compound_hash([
+                alert.fingerprint.encode('ascii'),
+                alert.status.encode('utf-8'),
+                str(alert.startsAt.timestamp()).encode('ascii'),
```
arikalon1, line 75: please also add the `endsAt`
I don't think you can get 2 alerts with everything the same, and only endsAt is different, but I don't see a downside in adding it

src/robusta/runner/web.py
```diff
@@ -57,11 +69,40 @@ def handle_alert_event():
...
+                    Web.processed_alerts_cache[alert_hash] = True
+            Web.alerts_queue.add_task(
+                Web.event_handler.handle_trigger, PrometheusTriggerEvent(alert=alert)
+            )
 
         Web.event_handler.get_telemetry().last_alert_at = str(datetime.now())
         return jsonify(success=True)
 
+    @staticmethod
+    def processed_alerts_cache_flusher():
```
arikalon1, line 90: I think we don't need this.
I think the TTL Cache handles expiration

src/robusta/runner/web.py
```diff
@@ -26,13 +32,19 @@ class Web:
...
+    processed_alerts_cache_lock = threading.Lock()
+
     @staticmethod
     def init(event_handler: PlaybooksEventHandler, loader: ConfigLoader):
         Web.metrics = QueueMetrics()
         Web.api_server_queue = TaskQueue(name="api_server_queue", num_workers=NUM_EVENT_THREADS, metrics=Web.metrics)
         Web.alerts_queue = TaskQueue(name="alerts_queue", num_workers=NUM_EVENT_THREADS, metrics=Web.metrics)
         Web.event_handler = event_handler
         Web.loader = loader
+        threading.Thread(
```
arikalon1, line 45: I think we don't need this

arikalon1 (commented):
nice work
left few comments

src/robusta/runner/web.py
RobertSzefler: Added

src/robusta/runner/web.py
RobertSzefler: As discussed, this is implemented to handle some worst case scenarios when the built-in expiration mechanism in `TTLCache` would be run and take a very long time. To guard against this, we run `expire` periodically to distribute the workload.

RobertSzefler: Okay, upon further inspection it looks like the "bulk expire", if it happens, in practice won't take so long to warrant having this mechanism implemented. Removed.

PR #1403, opened 2024-05-06
Updated zulip docs and minor fixes

pavangudiwada:
* Added Zulip docs to index
* Fixed Json formatting warnings - We have comments, so changed it to YAML
* yml won't work so changed it to yaml to remove warning.


arikalon1 (approved):
LGTM

PR #1404, opened 2024-05-06
Fix/zulip block formatting

oscgu:
The last arg in __to_zulip_block() should be the format, if anything, or else you end up with unreadable messages


arikalon1 (approved):
Thanks @oscgu , looks good

Unrelated to this fix, but why are you using `yield`?

oscgu:
> Thanks @oscgu , looks good
> 
> Unrelated to this fix, but why are you using `yield`?

IMO it makes it easier to return 1..N strings from the func, without the need for some sort of temporary list.
The performance difference should be negligible too.
Should I have implemented it differently?

PR #1413, opened 2024-05-10
Better supabase expiry handling

RobertSzefler:
Upgrade supabase client, enable automatic token refresh, remove old code that was supposed to handle it, some related cleanups.


arikalon1 (approved):
nice work

PR #1414, opened 2024-05-12
Handle the time the job is starting, and doesn't have any condition y…

arikalon1:
…et, or any active/failed/succeeded pods


aantn (approved):
Lgtm

PR #1418, opened 2024-05-15
Updated KRR OOMkill docs

pavangudiwada:
image


aantn:
LGTM. Can you also add a link at the section above where people might look for it? (See bolded text. Don't actually bold it.)

> If you are having performance issues, specifically with Prometheus using a lot of memory, reduce ``max_workers`` to reduce memory usage. KRR uses 3 workers by default. **If the KRR Job is getting OOMKilled adjust its memory requirements (link).**

aantn (approved):

arikalon1 (approved):
LGTM

PR #1419, opened 2024-05-15
Evicted trigger

Avi-Robusta:
This is only for __standard__ evictions due to node pressure
(when you run kubectl get pods and you see the status as evicted)

This is __not__ for:
1. Pod evictions due to node taint
for example :
```
kubectl taint nodes NODE_NAME key1=value1:NoExecute
```
2. Pod evections due to directly calling the eviction api
for example
```
kubectl proxy --port=8080  &
POD_TO_EVICT="POD_NAME"
curl -k -v -XPOST  -H "Accept: application/json" -H "Content-Type: application/json" "http://localhost:8080/api/v1/namespaces/default/pods/${POD_TO_EVICT}/eviction" --data '{"kind":"Eviction","apiVersion":"policy/v1beta1","metadata": {"namespace": "default","name": "'"${POD_TO_EVICT}"'"}}'
```

Both have different status's and fields in the pod yaml so they are not in this trigger, they can be added in the future.
--------
Here is an example playbook
```
customPlaybooks:
- triggers:
  - on_pod_evicted: {}
  actions:
  - create_finding: #
      title: "Pod $name in namespace $namespace was Evicted"
      aggregation_key: "PodEvictedTriggered"
```
Screen Shot 2024-05-15 at 17 14 58


Avi-Robusta:
Add Docs!!!

arikalon1 (approved):
LGTM

I would just add the `scope` to the docs of the trigger parameters

Issue #1422, opened 2024-05-16
make template_enricher support Prometheus Alert annotations


zmx:
**Is your feature request related to a problem?**
We have some third-party PrometheusRules having runbook URL in annotations

**Describe the solution you'd like**
Make template_enricher could annotate with PrometheusRules annotations

**Describe alternatives you've considered**
Static/Custom template_enrich configs

**Additional context**
There are previous discussion/request in https://robustacommunity.slack.com/archives/C02R0LVANKY/p1708600339808769

https://github.com/robusta-dev/robusta/blob/67c77644516af5a19cab7ccb8fce54f6197002ab/playbooks/robusta_playbooks/alerts_integration.py#L331

```
 if isinstance(event, PrometheusKubernetesAlert):
        labels.update(event.alert.labels)
        labels.update(event.alert.annotations)
```

zmx:
P.S. mention_enricher already support event.alert.annotations

PR #1423, opened 2024-05-16
make template_enricher support prometheus alert annotations (#1422)

zmx:
Current template_enricher only support prometheus alert.labels, make it support alert.annotations.
The same feature already exists in mention_enricher


arikalon1 (approved):
nice work!
Thanks for your contribution @zmx

PR #1428, opened 2024-05-17
Document how to run on specific nodes w/ embedded prometheus stack

aantn:
Based on https://artifacthub.io/packages/helm/prometheus-community/kube-prometheus-stack?modal=values&path=prometheusOperator.admissionWebhooks.patch.affinity and https://github.com/prometheus-community/helm-charts/issues/1853

### DIFF
--- a/docs/setup-robusta/node-selector.rst
+++ b/docs/setup-robusta/node-selector.rst
@@ -21,10 +21,37 @@ Add the following to your Helm values:
       nodeSelector:
         kubernetes.io/os: linux
 
+    # if using Robusta's embedded kube-prometheus-stack, you can configure the Prometheus Operator's components to run on a specific node too
+    kube-prometheus-stack:
+      alertmanager:
+        alertmanagerSpec:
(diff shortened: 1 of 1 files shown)


docs/setup-robusta/node-selector.rst
```diff
@@ -21,11 +21,38 @@ Add the following to your Helm values:
       nodeSelector:
         kubernetes.io/os: linux
 
+    # if using Robusta's embedded kube-prometheus-stack, you can configure the Prometheus Operator's components to run on a specific node too
+    kube-prometheus-stack:
```
pavangudiwada, line 25: I suppose we only need one of these, no?

pavangudiwada (approved):

docs/setup-robusta/node-selector.rst
aantn: @pavangudiwada I think they're all needed. Each subvalue here controls a different component in the Prometheus stack. You need to add the selector to each one.

Issue #504, opened 2022-08-30
Integration using Webhook provided by Slack


SharmaSid00:
**Is your feature request related to a problem?**
As we integrate robusta in our slack, it ask permission for full slack workspace, however it is not possible to provide full slack workspace in a company.
**Describe the solution you'd like**
Enable addition of webhook url provided by slack itself - unique for each channel

aantn:
Hi, can you elaborate on the restrictions you have adding the Slack app?

We’ve tried to make the Slack permissions super minimal. E.g. we don’t ask for read access to messages in your workspace. Is there a specific permission there which is problematic?

That said, we’re definitely open to a PR adding support for this. You can see an [example PR](https://github.com/robusta-dev/robusta/pull/445/files) adding a sink here, we have a [guide on adding new sinks](https://docs.robusta.dev/master/developer-guide/platform/sinks-guide.html), and you can find all the source code for the [existing Slack sink here](https://github.com/robusta-dev/robusta/tree/master/src/robusta/core/sinks/slack) and the similar [Mattermost sink here](https://github.com/robusta-dev/robusta/blob/master/src/robusta/core/sinks/mattermost/mattermost_sink.py).

Lastly, on our commercial plans there is an option to create your own Slack app which is entirely self-hosted by you. See the self-hosting docs.

PR #1430, opened 2024-05-19
add a work around for sending messages to slack via webhook

romga:
suppose to solve issue:
https://github.com/robusta-dev/robusta/issues/504


arikalon1 (approved):
looks good
thanks for the contribution @romga

Issue #1168, opened 2023-11-14
Logs enricher when use prometheus does not show the graph for node utilisation


antikilahdjs:
**Describe the bug**

Hey, I am using the Robusta some days ago with Prometheus plus Alertmanager but the enricher for prometheus like memory or cpu is empty when fired to teams as alert. Only the graph for memory usage appears in the graph but the node utilisation is not showing.

The image below will demonstrate it 

![image](https://github.com/robusta-dev/robusta/assets/33559260/7e86808f-3e34-4c81-ba88-303ac5f1620a)

**To Reproduce**
1 - Install using the official helm charts
3 - Configure Prometheus and  Alertmanager
4 - Configure the SINK to use Teams
5 - In my global env has been configured like below 
```yaml
globalConfig:
  grafana_url: ""
  grafana_api_key: ""
  grafana_dashboard_uid: ""
  alertmanager_url: "http://alertmanager-operated.thanos:9093"
  prometheus_url: "http://thanos-query-frontend.thanos:9090"
  signing_key: ""
  account_id: 695c3053-0e56-xxxxxxxxxxxxxxxxxxxxxx
  custom_annotations: []
```

5 - My trigger and action is:
```yaml
- triggers:
  - on_pod_oom_killed:
      rate_limit: 3600
  actions:
  - pod_oom_killer_enricher: {}
  - logs_enricher: {}
  - pod_node_graph_enricher:
      resource_type: Memory
      display_limits: true
  - oomkilled_container_graph_enricher:
      resource_type: Memory
      display_limits: true
  stop: true
```

**Expected behavior**

The graph woks for both sides, node utilisation and pod utilisation

**Screenshots**
It was added above

**Desktop (please complete the following information):**
 - OS: RedHat 8.5 and Ubunut 20.04LTS
 - Browser: Chrome
 - Version: 119

**Additional context**

saireddyb:
Yes I to receive the node graph empty.

wrbbz:
Same here on Robusta 0.12.0 without UI integration

arikalon1:
@Bobses @wrbbz do you see any exception in the robusta-runner pod logs ?

aantn:
Hi all, I believe this is because robusta is using the recording rule `instance:node_memory_utilisation:ratio` which isn't present in your environment.

[...]

PR #1432, opened 2024-05-21
Usage of records definition instead of records itself

wrbbz:
Uses metrics definition instead of metrics record.

As was shown in #1168, those records could be absent on some Prometheus instances


aantn (approved):
LGTM thank you!

PR #1433, opened 2024-05-21
add promtheus labels to slack message

romga:


arikalon1 (approved):
LGTM

wrbbz:
But why?

romga:
> But why?

can you please elaborate?

wrbbz:
What for do we need prom labels in slack message? As I think, it is a little bit excessive information. Am I wrong?

romga:
@wrbbz 
i am using the labels in order to decide which alert sent to which channel :)

wrbbz:
Why not use kubernetes labels?

Nevertheless, they will be shown in aggregation key

On the other hand, I would suggest more dynamic message creation. If there are some labels - we add labels in the message, if none - no mention at all (instead of `None`). It could be misleading. I would have decided, that there is an issue with some data absence if I see `None`

PR #1438, opened 2024-05-22
added custom service account

Avi-Robusta:
How to configure from generated_values.yaml
```
kubewatch:
  customServiceAccount: test-service-account
runner:
  customServiceAccount: test-service-account
```
Verified pods have the same origonal service account when this is not set


helm/robusta/values.yaml
```diff
@@ -551,6 +551,7 @@ kubewatch:
       readOnlyRootFilesystem: false
       runAsUser: 1000
     pod: {}
+  customServiceAccount: "" # to override the kubewatch service account
```
arikalon1, line 554: will it make more sense to put in in the `serviceAccount` object?
```
seviceAccount:
  name: ""
```

helm/robusta/values.yaml
```diff
@@ -587,6 +588,7 @@ runner:
   sentry_dsn: https://53b627690db14de7b02095407596fa16@o1120648.ingest.sentry.io/6156573
   sendAdditionalTelemetry: false
   certificate: "" # base64 encoded
+  customServiceAccount: "" # to override the runner service account
```
arikalon1, line 591: would it be better to add it in the `runnerServiceAccount` object?

arikalon1 (approved):
looks good

left a minor comment

PR #1450, opened 2024-06-10
Remove robusta cli

Sheeproid:
As the CLI code moved to it's own project, removing the remainder of the code which is no longer useful


arikalon1 (approved):
LGTM

Issue #1314, opened 2024-02-28
Unable to deploy robusta helm chart with ArgoCD


mckernanin:
**Describe the bug**
Original slack conversation: https://robustacommunity.slack.com/archives/C02R0LVANKY/p1707937278191989

I am unable to deploy robusta when using a yaml templated application for ArgoCD. As suggested on slack, I have tried a number of different ways to refer to my secret values, none of them have worked. Example yaml application for Argo to deploy the helm chart:

```yaml
apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  name: robusta
  namespace: argocd
  finalizers:
  - resources-finalizer.argocd.argoproj.io
spec:
  destination:
    server: https://kubernetes.default.svc
    namespace: robusta
  project: default
  source:
    chart: robusta
    repoURL: https://robusta-charts.storage.googleapis.com
    targetRevision: 0.10.29
    helm:
      values: |
        globalConfig:
          signing_key: "{{ '{{' }} env.SIGNING_KEY {{ '}}' }}"
          account_id: "{{ '{{' }} env.ACCOUNT_ID {{ '}}' }}"
        sinksConfig:
        - robusta_sink:
            name: robusta_ui_sink
            token: "{{ '{{' }} env.ROBUSTA_UI_TOKEN {{ '}}' }}"
        - discord_sink:
          name: mckernans_discord_sink
          url: "{{ '{{' }} env.DISCORD_SINK_WEBHOOK {{ '}}' }}"
        enablePlatformPlaybooks: true
        runner:
          sendAdditionalTelemetry: true
        rsa:
          private: "{{ '{{' }} env.RSA_PRIVATE {{ '}}' }}"
          public: "{{ '{{' }} env.RSA_PUBLIC {{ '}}' }}"
        runner:
          additional_env_vars:
            - name: SIGNING_KEY
              valueFrom:
                secretKeyRef:
                  name: environment
                  key: SIGNING_KEY
            - name: ACCOUNT_ID
              valueFrom:
                secretKeyRef:
                  name: environment
                  key: ACCOUNT_ID
            - name: ROBUSTA_UI_TOKEN
              valueFrom:
                secretKeyRef:
                  name: environment
                  key: ROBUSTA_UI_TOKEN
            - name: DISCORD_SINK_WEBHOOK
              valueFrom:
                secretKeyRef:
                  name: environment
                  key: DISCORD_SINK_WEBHOOK
            - name: RSA_PRIVATE
[...]

PR #1451, opened 2024-06-10
Update docs on deploying with argocd

aantn:
See https://github.com/robusta-dev/robusta/issues/1314#issuecomment-2158639324


arikalon1 (approved):
LGTM

PR #1457, opened 2024-06-14
Fix markdown block quote truncation

RobertSzefler:


pyproject.toml
```diff
@@ -71,6 +71,8 @@ botocore = "1.31.72"
 boto3 = "1.28.72"
 requests = "^2.32.3"
 certifi = "^2023.7.22"
+regex = "2024.5.15"
```
RobertSzefler, line 74: This because stdlib `re` does not support reverse search (`regex.REVERSE`) up to Python 3.12

pyproject.toml
```diff
@@ -71,6 +71,8 @@ botocore = "1.31.72"
...
+more_itertools = { version = "^10.3", python = "<3.12" }
```
RobertSzefler, line 75: This because `batched` appeared in stdlib `itertools` only in Python 3.12

src/robusta/core/sinks/transformer.py
```diff
@@ -58,6 +58,10 @@ def apply_length_limit(msg: str, max_length: int, truncator: Optional[str] = Non
         truncator = truncator or "..."
         return msg[: max_length - len(truncator)] + truncator
 
+    @staticmethod
+    def apply_length_limit_to_markdown(msg: str, max_length: int, truncator: str = "...") -> str:
+        return trim_markdown(msg, max_length, truncator)
```
Avi-Robusta, line 63: i would put this in a try catch and if this fails try to run apply_length_limit

Avi-Robusta (approved):
LGTM one comment

src/robusta/core/sinks/transformer.py
RobertSzefler: done

PR #1459, opened 2024-06-16
Streamline integration docs

aantn:


pavangudiwada (approved):
"Data Sources" makes a lot of sense
image

PR #1460, opened 2024-06-17
Add alertmanager troubleshooting docs

aantn:
# Motivation
Today, when the AlertManager isn't configued (or it is configured wrong), the UI sends users to https://docs.robusta.dev/master/configuration/alertmanager-integration/index.html

That page doesn't contain troubleshooting instructions, nor is it specific to alertmanager (it also describes the Prometheus integration)

This PR adds a new page that we can send UI users to when alertmanager isn't setup. The new page contains more useful information for UI users.


pavangudiwada (approved):
Very helpful! :clap:

PR #1466, opened 2024-06-19
MAIN-1682 Upgrade branch

Avi-Robusta:


Dockerfile
```diff
@@ -1,33 +1,69 @@
...
+    && apt-get install -y --no-install-recommends git ssh curl libcairo2 \
+    && rm -rf /var/lib/apt/lists/*
+
+# Patching CVE-2024-32002
+RUN git config --global core.symlinks false
 
+# Run the application
 # -u disables stdout buffering https://stackoverflow.com/questions/107705/disable-output-buffering
-CMD [ "python3", "-u", "-m", "robusta.runner.main"]
+CMD [ "python3", "-u", "/app/src/robusta/runner/main.py"]
```
RobertSzefler, line 69: Why do we insist on running with `-u`?

RobertSzefler (approved):
LGTM but I suggest getting a second, independent review. Left one minor comment.

Dockerfile
```diff
@@ -1,33 +1,69 @@
-# see https://pythonspeed.com/articles/alpine-docker-python/ for the reason we don't use alpine
-FROM python:3.9-slim
+# Build stage
+FROM python:3.11-slim as builder
+ENV PATH="/root/.local/bin/:$PATH"
+
 RUN apt-get update \
     && dpkg --add-architecture arm64 \
-    && apt-get install -y --no-install-recommends git ssh socat wget curl libcairo2 python3-dev libffi-dev \
+    && apt-get install -y --no-install-recommends curl gcc \
```
arikalon1, line 7: I think we need `&& apt-get purge -y --auto-remove gcc \`

Why did you remove it?

Dockerfile
```diff
@@ -1,33 +1,69 @@
...
+
+# Install project dependencies
 COPY pyproject.toml poetry.lock /app/
-RUN /root/.local/bin/poetry install --no-root --no-dev --extras "all"
+RUN poetry install --no-root --without dev --extras "all"
+
+# Copy project source code
 COPY src/ /app/src
-RUN /root/.local/bin/poetry install --no-dev --extras "all"
+RUN poetry install --without dev --extras "all"
```
arikalon1, line 36: why are we doing `poetry install` twice?

Dockerfile
```diff
@@ -1,33 +1,69 @@
...
+COPY --from=builder /etc/robusta/playbooks/defaults /etc/robusta/playbooks/defaults
+# Copy virtual environment and application files from the build stage
+COPY --from=builder /app /app
+
+# Install necessary packages for the runtime environment
+RUN apt-get update \
+    && dpkg --add-architecture arm64 \
+    && pip3 install --no-cache-dir --upgrade pip \
+    && apt-get install -y --no-install-recommends git ssh curl libcairo2 \
+    && rm -rf /var/lib/apt/lists/*
```
arikalon1, line 62: I think we need `&& apt-get purge -y` here as well

pyproject.toml
```diff
@@ -18,7 +18,7 @@ include_trailing_comma = true
 
 
 [tool.poetry.dependencies]
-python = "^3.8, <3.11"
+python = "^3.8, <3.12"
```
arikalon1, line 21: Should this really be `3.12` or `3.11`?

arikalon1 (approved):
nice work

I left some comments

Merging it since there are no blockers here, butt probably worth fixing

Dockerfile
Avi-Robusta: since we are doing a two stage build, we dont need to uninstall anything in the first stage

Dockerfile
Avi-Robusta: this comment was in the original dockerfile
`# we install the project requirements and install the app in separate stages to optimize docker layer caching
`
I think ill add it back in

pyproject.toml
Avi-Robusta: this dockerfile is 3.11, so you mean <= 3.11?
I think this causes issues with subversions like 3.11.1 thats why they do <next python version
I had an issue with 3.12 where i did a specific subversion and they released a new subversion breaking the project

Dockerfile
Avi-Robusta: ill check this

PR #1471, opened 2024-06-24
Holmes chart dependency

arikalon1:


src/robusta/core/model/base_params.py
```diff
@@ -70,6 +71,45 @@ def post_initialization(self):
...
+        if v and not v.startswith("http"):  # if the user configured url without http(s)
+            v = f"http://{v}"
+            logging.info(f"Adding protocol to holmes_url: {v}")
+        return v
+
+
+class AIInvestigateParams(HolmesParams):
+    """
+    :var resource: The resource related to this investigation. A resource has a `name` and `kind`, and may have `namespace` and `node`
+    :var investigation_type: The type of investigation: Issue/Service/Cluster/Custom
```
Avi-Robusta, line 97: this comment is a bit confusing since I see the code only sets it to `issue` or `analyze_problems` not Service/Cluster/Custom
Correct me if im wrong?

src/robusta/core/model/base_params.py
```diff
@@ -70,6 +71,45 @@ def post_initialization(self):
...
+    :var runbooks: List of human readable recommended runbooks that holmes can use for the investigation.
```
Avi-Robusta, line 98: we support custom runbooks from the action? cool

src/robusta/core/playbooks/internal/ai_integration.py
```diff
@@ -0,0 +1,70 @@
...
+    if not holmes_url:
+        logging.error("Holmes url not found")
+        return
+
+    investigation__title = build_investigation_title(params.investigation_type, params)
+    try:
+        holmes_req = HolmesRequest(
+            source=params.context.get("source", "unknown source"),
+            title=f"{investigation__title}",
+            subject=params.resource.dict() if params.resource else None,
```
Avi-Robusta, line 35: i think subject might not be able to be none via pydantic so this line might need to be
`subject=params.resource.dict() if params.resource else {}`

src/robusta/core/playbooks/internal/ai_integration.py
```diff
@@ -0,0 +1,70 @@
...
+            context=params.context if params.context else None,
```
Avi-Robusta, line 36: context also cant be None so replace with {}

src/robusta/core/playbooks/internal/ai_integration.py
```diff
@@ -0,0 +1,70 @@
...
+            include_tool_calls=True,
+            include_tool_call_results=True,
+        )
+        result = requests.post(f"{holmes_url}/api/investigate", data=holmes_req.json())
```
Avi-Robusta, line 40: will this run with no subject and/or context?

src/robusta/integrations/slack/sender.py
```diff
@@ -341,6 +370,93 @@ def __create_links(self, finding: Finding):
...
+
+        ai_enrichments = [
+            enrichment for enrichment in finding.enrichments if enrichment.enrichment_type == EnrichmentType.ai_analysis
+        ]
+
+        if not ai_enrichments:
+            logging.warning(f"No matching ai enrichments found for id: {finding.id} - {title}")
+            return
+
+        ai_analysis_blocks = [block for block in ai_enrichments[0].blocks if isinstance(block, HolmesResultsBlock)]
```
Avi-Robusta, line 414: just to clarify, right now only slack and the UI can get HolmesResultsBlock?

src/robusta/core/playbooks/internal/ai_integration.py
```diff
@@ -0,0 +1,70 @@
...
+            failure=False,
+        )
+        finding.add_enrichment(
+            [HolmesResultsBlock(holmes_result=holmes_result)], enrichment_type=EnrichmentType.ai_analysis
+        )
+
+        event.add_finding(finding)
+
+    except Exception:
+        logging.exception(f"Failed to get holmes analysis for {investigation__title} {params.context}")
```
Avi-Robusta, line 70: is this enough debugging information for you or do you need subject too?

src/robusta/core/reporting/callbacks.py
```diff
@@ -27,7 +27,7 @@ def create_for_func(
 
         if not signing_key:
             raise Exception("Cannot create callback request with no signing key. Configure signing_key in globalConfig")
-        action_params = {} if choice.action_params is None else choice.action_params.dict()
+        action_params = {} if choice.action_params is None else choice.action_params.dict(exclude_defaults=True)
```
Avi-Robusta, line 30: will this break any other callback functionality?

src/robusta/integrations/slack/sender.py
```diff
@@ -341,6 +370,93 @@ def __create_links(self, finding: Finding):
...
+            text=text,
+            blocks=[
+                {
+                    "type": "section",
+                    "text": {"type": "mrkdwn", "text": text},
+                }
+            ],
+        )
+
+    def send_holmes_analysis(
```
Avi-Robusta, line 395: I am not in love with this flow but I get it.

src/robusta/integrations/slack/sender.py
```diff
@@ -351,6 +467,19 @@ def send_finding_to_slack(
...
+            sink_params.slack_channel,
+            self.cluster_name,
+            finding.subject.labels,
+            finding.subject.annotations,
+        )
+
+        if finding.finding_type == FindingType.AI_ANALYSIS:
+            # holmes analysis message needs special handling
+            self.send_holmes_analysis(finding, slack_channel, platform_enabled, thread_ts)
+            return ""  # [arik] Looks like the return value here is not used, needs to be removed
```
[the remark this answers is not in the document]
arikalon1: this is just a reminder, didn't want to include it in this change

src/robusta/integrations/slack/sender.py
arikalon1: yeah, it's a different way of handling

src/robusta/core/reporting/callbacks.py
arikalon1: no, I checked 
This was actually a bug in the way we handled callback params
Since we weren't using callbacks much, it was un noticed

src/robusta/core/playbooks/internal/ai_integration.py
arikalon1: fixing

src/robusta/core/playbooks/internal/ai_integration.py
arikalon1: yes, I think you can ask a question on the cluster, and it would work

src/robusta/core/playbooks/internal/ai_integration.py
arikalon1: yeah, added the subject as well

src/robusta/core/model/base_params.py
arikalon1: We didn't implement it yet, but it's coming very soon, and it's going to be super cool!

src/robusta/core/model/base_params.py
arikalon1: good catch, fixed

PR #1473, opened 2024-06-25
Support for aligning Slack message grouping intervals

RobertSzefler:


src/robusta/core/sinks/sink_base_params.py
```diff
@@ -91,12 +91,21 @@ def validate_exactly_one_defined(cls, values: Dict):
...
     interval: int = 15*60  # in seconds
+    aligned: bool = False
     notification_mode: Optional[NotificationModeParams]
 
     @root_validator
-    def validate_notification_mode(cls, values: Dict):
-        if values is None:
-            return {"summary": SummaryNotificationModeParams()}
+    def validate_interval_alignment(cls, values: Dict):
+        if values["aligned"]:
```
Avi-Robusta, line 99: what if values is none? wont this throw? i would handle that case first

src/robusta/core/sinks/sink_base_params.py
```diff
@@ -91,12 +91,21 @@ def validate_exactly_one_defined(cls, values: Dict):
...
+            if values["interval"] < 24 * 3600:
+                if (24 * 3600) % values["interval"]:
```
Avi-Robusta, line 101: not sure i understand, we allow the interval to be seconds but we dont support seconds, shouln't we change it to minutes or have it round up to the nearest minute instead of failing ?

src/robusta/core/sinks/sink_base.py
```diff
@@ -33,20 +34,63 @@ def register_notification(self, interval: int, threshold: int) -> bool:
...
 
-    def register_notification(self, summary_key: KeyT, resolved: bool, interval: int):
-        now_ts = time.time()
+    def register_notification(self, summary_key: KeyT, resolved: bool, interval: int, aligned: bool):
+        now_dt = datetime.now()
+        now_ts = int(now_dt.timestamp())
         idx = 1 if resolved else 0
-        if now_ts - self.start_ts > interval or not self.summary_table:
-            # Expired or the first summary ever for this group_key, reset the data
+        if not self.end_ts or now_ts > self.end_ts:
```
Avi-Robusta, line 46: what if not self.summary_table?

src/robusta/core/sinks/sink_base_params.py
RobertSzefler: It's marked as required, so only possible values should be `True` and `False`.

src/robusta/core/sinks/sink_base.py
RobertSzefler: If `end_ts` is not set it means that the whole object is not initialized, and so `summary_table` is also not set. I basically simplified the condition here.

src/robusta/core/sinks/sink_base_params.py
RobertSzefler: There will be no backwards compatibility problems, as prior users don't use the interval aligning mechanism (it's not been released yet).

Also, we support aligning on thresholds shorter than a minutes here without problem. For example, if the user specs `interval` as 15 seconds, the intervals will be automatically aligned at `x:00`, `x:15`, `x:30`, `x:45` (for consecutive minutes `x`).

Avi-Robusta: got it

PR #1475, opened 2024-06-29
MAIN 1815 - Handle playbooks from external tar/tgz files etc

RobertSzefler:


src/robusta/runner/config_loader.py
```diff
@@ -116,32 +120,81 @@ def __load_playbooks_repos(
...
-
-                    if not os.path.exists(local_path):  # in case the repo url was defined before it was actually loaded
-                        logging.error(f"Playbooks local path {local_path} does not exist. Skipping")
-                        continue
-
-                    # Adding to pip the playbooks repo from local_path
-                    subprocess.check_call([sys.executable, "-m", "pip", "install", "--no-build-isolation", local_path])
-                    playbook_package = self.__get_package_name(local_path=local_path)
+                    remove_pkg_path = None
+                    try:
```
Avi-Robusta, line 124: where is the except for this? (it could be something i dont see in git?
alsi there is a try a few lines up, why do we need another try?

src/robusta/runner/config_loader.py
```diff
@@ -116,32 +120,81 @@ def __load_playbooks_repos(
...
+                        check_pkg_path = True
+                        if (
```
Avi-Robusta, line 126: I would just add a comment above, for url of zipped playbooks

src/robusta/runner/config_loader.py
```diff
@@ -116,32 +120,81 @@ def __load_playbooks_repos(
...
+                            if playbooks_repo.url.startswith("http://"):
+                                logging.warning(
+                                    f"Downloading a playbook package from non-https source f{playbooks_repo.url}"
+                                )
+                            temp_dir = tempfile.mkdtemp()
+                            # after installing the downloaded package, remove it along with its directory
+                            remove_pkg_path = temp_dir
+                            pkg_name = urlparse(playbooks_repo.url).path.rsplit('/', 1)[-1]
+                            filename = os.path.join(temp_dir, pkg_name)
+                            self.download_file(playbooks_repo.url, filename, playbooks_repo.http_headers)
```
Avi-Robusta, line 147: where do you unzip the file?

Avi-Robusta: did you try running this?

src/robusta/runner/config_loader.py
```diff
@@ -116,32 +120,81 @@ def __load_playbooks_repos(
...
+                            pkg_path = f"file://{filename}"
+                            check_pkg_path = False
+                        elif playbooks_repo.url.startswith(GIT_SSH_PREFIX) or playbooks_repo.url.startswith(GIT_HTTPS_PREFIX):
```
Avi-Robusta, line 150: do these (below) also need the specified http_headers?

docs/playbook-reference/defining-playbooks/external-playbook-repositories.rst
```diff
@@ -109,6 +116,20 @@ The ``key`` parameter must contain a ``base64`` encoded deployment key with ``re
...
+
+For external Python packages, just specify an URL starting with http(s), and ending with
+either .tar, .tar.gz, .tgz, .tar.bz2, or .tbz2.
+
+.. code-block:: yaml
+
+    playbookRepos:
+      web_playbooks:
+        url: "https://my-domain.com/bla/web-playbooks.tgz"
+        http_headers: # optional, may be used for auth
```
Avi-Robusta, line 130: i think we should either allow this to be used for all the urls or specify here that its only for zip and use it in the validator

RobertSzefler: It's explicitly requested (see https://robusta.youtrack.cloud/issue/MAIN-1815/load-actions-repo-from-https-get-url) that this only applies to external Python packages (tgz etc) loaded via the mechanism implemented here.

src/robusta/runner/config_loader.py
RobertSzefler: There is no `except`. There is a `finally` in line 174 below.

docs/playbook-reference/defining-playbooks/external-playbook-repositories.rst
RobertSzefler: I made it more clear in the docs.

src/robusta/runner/config_loader.py
RobertSzefler: Yes, I tested it. The file is unzipped into a temporary directory created in line 142.

src/robusta/runner/config_loader.py
RobertSzefler: Nope, it does not apply to them.

src/robusta/runner/config_loader.py
RobertSzefler: Added a comment

RoiGlinik:
rework to extract tgz and install

docs/playbook-reference/defining-playbooks/external-playbook-repositories.rst
```diff
@@ -109,6 +116,27 @@ The ``key`` parameter must contain a ``base64`` encoded deployment key with ``re
...
+either .tar.gz or .tgz.
+
+.. code-block:: yaml
+
+    playbookRepos:
+      web_playbooks:
+        url: "https://my-domain.com/bla/web-playbooks.tgz"
+        http_headers: # optional, may be used for auth
+          Authorization: Bearer XXXYYY
+        # pip_install: true      # optional: load this playbook's dependencies (default True)
```
arikalon1, line 132: I think this might be a little confusing for users.
The `pip_install` and `build_isolation` are related to all forms of playbooks loading, and not only the `tgz`

So I think it should be removed from here, and documented on a separate section, explaining that this should be used when installing playbooks on air gapped envs, or similar.
WDYT?

arikalon1 (approved):
nice work

left one comment regarding the docs

docs/playbook-reference/defining-playbooks/external-playbook-repositories.rst
RoiGlinik: created a different section for it

Issue #1083, opened 2023-09-19
dynamic channel on the teams sink


gioppoluca:
**Is your feature request related to a problem?**
The need is to choose the teams where to send info dynamically without having to change anything in robusta.
Teams are associated to projects that are managed outside Robusta: the ability to read labels from Alerts (through alert relabeling) will enable Robusta to send to a sink it does not know and is not hard coded in its configurartion

**Describe the solution you'd like**
Solution similar to slack where sink info is red from labels/annotations

**Describe alternatives you've considered**
Defining CR for sink that are placed in the project NS and Robusta able to watch for these CR and configure itself to use those new CR

**Additional context**
Talked about this on Slack with @arikalon1

gioppoluca:
to be more specific the need is to override the existing configuration of the teams configuration:
```
sinksConfig:
- ms_teams_sink:
    name: dynamic_ms_teams_sink
    webhook_url: teams channel incoming webhook  # see video below
```

I believe that at the moment Robusta require that we define a MSTeams like the `dynamic_ms_teams_sink`
The need is to override the `webhook_url` with the value taken from the label/annotation (probably annotations should be better).
In this way we can have an existing sink configuration that robusta needs to wire its stuff but we would dynamically send to different destinations every time.

RoiGlinik:
@gioppoluca This will be available in the next version

PR #1476, opened 2024-06-29
Added support for webhook_url dynamic overriding using annotations (issue 1083)

itisallgood:
Introduced the `webhook_override` attribute for ms_teams_sink, enabling the use of annotations from YAML files to dynamically override the `webhook_url`.

 [Issue #1083](https://github.com/robusta-dev/robusta/issues/1083).

```
sinksConfig:
- ms_teams_sink:
    name: main_ms_teams_sink
    webhook_url: teams-incoming-webhook 
    webhook_override: "annotations.msteams"

```
If no such annotation exists, the default webhook will be used. If the annotation is found but its value
does not contain a valid URL, the system will search for an environmental variable with the name of the value in the ``additional_env_vars`` section of your ``generated_values.yaml`` file.

**Reason for using annotations instead of labels:**

Annotations are used instead of labels because URLs can't be passed to labels (see comment for `MsTeamsWebhookUrlTransformer` in `src/robusta/core/sinks/msteams/msteams_webhook_tranformer.py` )


src/robusta/core/sinks/msteams/msteams_webhook_tranformer.py
```diff
@@ -0,0 +1,65 @@
...
+            if not regex.match(ANNOTATIONS_COMPOSITE_PATTERN, v):
+                err_msg = f"webhook_override must be '{ANNOTATIONS_PREF}foo' or contain patterns like: '${ANNOTATIONS_PREF}foo'"
+                raise ValueError(err_msg)
+        return v
+
+    @classmethod
+    def validate_url_or_get_env(cls, webhook_url: str, default_webhook_url: str) -> str:
+        if URL_PATTERN.match(webhook_url):
+            return webhook_url
+        env_value = os.getenv(webhook_url)
```
RoiGlinik, line 42: Can you add info logs here, in case the url matching is failed?

Also is there any reference to this url pattern ?

itisallgood: @RoiGlinik There is no reference to this url pattern as I composed it from different resources, but I added pytests with different valid and invalid urls.

RoiGlinik (approved):

RoiGlinik (approved):

PR #1480, opened 2024-07-01
More doc improvements

aantn:


docs/images/ai-investigation.png
aantn, line 1: @pavangudiwada I am not sure this is the best image here. It seems out of place with the other images and I can't put my finger on why. Maybe it is best to just show the AI Analysis part without the window underneath it?

pavangudiwada: Not that different, but wdyt?

image

aantn: Lgtm

pavangudiwada (approved):

PR #1484, opened 2024-07-02
MAIN-1825 playbook bugfix holmes

Avi-Robusta:
params.context and params.resource is optional so the manual trigger was broken

now it works:

` robusta playbooks trigger ask_holmes investigation_type=analyze_problems ask="What issues do i have in my cluster?"
`
Screen Shot 2024-07-02 at 10 11 23


arikalon1 (approved):
LGTM

PR #1485, opened 2024-07-02
bug when loading custom certificate too late - some http clients were…

arikalon1:
… initialized before it was loaded, caused ssl errors


RobertSzefler (approved):
+1

PR #1494, opened 2024-07-09
Slack sink: possibility to disable investigate links/buttons (MAIN-1857)

RobertSzefler:
Tests performed: setting `investigate_link` to `true`/`false` (and unset) for both grouping and normal Slack message sending and observing the results.


arikalon1 (approved):
nice work

PR #1496, opened 2024-07-10
[MAIN-1742] clean up krr finalizers

Avi-Robusta:


playbooks/robusta_playbooks/robusta_utils.py
```diff
@@ -0,0 +1,47 @@
...
+def timestamp_in_last_hours_back(timestamp: datetime, hours_back: int) -> bool:
+    # can't subtract offset-naive and offset-aware datetimes
+    timestamp_native = timestamp.replace(tzinfo=None)
+    now_native = datetime.now() .replace(tzinfo=None)
+    return now_native - timestamp_native < timedelta(hours=hours_back)
+
+
+@action
+def cleanup_robusta_pods(event: ExecutionBaseEvent, params: RobustaUtilsParams):
+    logging.info("running test_pod_orm")
```
arikalon1, line 26: don't think this is required

playbooks/robusta_playbooks/robusta_utils.py
```diff
@@ -0,0 +1,47 @@
...
+    v1 = client.CoreV1Api()
+    pod_list = v1.list_namespaced_pod(INSTALLATION_NAMESPACE)
+    finalizers_to_remove = ["robusta.dev/krr-job-output"]
+
+    # remove krr finalizers
+    for pod in pod_list.items:
+        if 'krr-job' not in pod.metadata.name:
+            continue
+
+        if pod.status.phase == 'running':
```
arikalon1, line 36: Is this working?
I thought the phase is `Running` and not `running`
In any case, worth comparing `lower(pod.status.phase)`

What is the phase of a pod that is stuck in `Terminating`?

playbooks/robusta_playbooks/robusta_utils.py
```diff
@@ -0,0 +1,47 @@
...
+            continue
+
+        creation_ts = getattr(pod.metadata, "creation_timestamp", None)
+        if not creation_ts or timestamp_in_last_hours_back(creation_ts, params.hours_back):
+            continue
+        body = {"metadata": {"$deleteFromPrimitiveList/finalizers": finalizers_to_remove}}
+        client.CoreV1Api().patch_namespaced_pod(
```
arikalon1, line 43: I would add an info log here, in the case we're actually cleaning finalizers

helm/robusta/values.yaml
```diff
@@ -554,6 +554,17 @@ platformPlaybooks:
   sinks:
     - "robusta_ui_sink"
 
+- name: "RobustaMaintenance"
+  triggers:
+  - on_schedule:
+      cron_schedule_repeat:
+        cron_expression: "0 * * * *" # every hour
```
arikalon1, line 561: I think we can run less than every hour
I think every few hours is probably enough

arikalon1 (approved):
nice work
few minor comments

playbooks/robusta_playbooks/robusta_utils.py
Avi-Robusta: Succeeded, Failed or Unknown
https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase

helm/robusta/values.yaml
Avi-Robusta: okay, every 3 hours

PR #1498, opened 2024-07-15
fix: add missing _get_sink_type method to zulip sink

oscgu:
Without the method the runner fails on startup:
```
sinks_config -> 0 -> zulip_sink
  Can't instantiate abstract class ZulipSinkParams with abstract method _get_sink_type (type=type_error)
```


arikalon1 (approved):
nice work
thanks for the contribution

PR #1501, opened 2024-07-18
Added prometheus_get_label_names action for the UI 

itisallgood:
Added prometheus_get_label_names action for the UI


playbooks/robusta_playbooks/prometheus_enrichments.py
```diff
@@ -99,6 +99,33 @@ def get_prometheus_series(prometheus_params: PrometheusGetSeriesParams) -> dict:
         raise e
 
 
+class PrometheusGetLabelNames(PrometheusParams):
+    """
+    :var match: List of Prometheus series selectors.
```
arikalon1, line 104: I assume the UI will always call it without params
Can you please add in the comments, what is the behavior when leaving each of the params empty.

arikalon1 (approved):
nice work

Left one minor comment

playbooks/robusta_playbooks/prometheus_enrichments.py
itisallgood: @arikalon1 Added comments

PR #1502, opened 2024-07-18
[MAIN-1863] - avg utilization of mem/cpu 

Avi-Robusta:


src/robusta/core/sinks/robusta/prometheus_discovery_utils.py
```diff
@@ -39,6 +38,30 @@ def __init__(self, discovery_period_sec: int, global_config: dict):
     def get_status(self) -> PrometheusHealthStatus:
         return self.status
 
+    def get_cluster_avg_cpu(self) -> Optional[float]:
+        cpu_query = os.getenv("OVERRIDE_CLUSTER_CPU_AVG_QUERY",
+                              f'avg_over_time(sum(irate(container_cpu_usage_seconds_total{{}}[5m]))[1h:])')
```
arikalon1, line 43: I'm not sure this is the correct query
The goal, is to show the avg **node** cpu utilization in the cluster, during the last hour (this metric is for `container` which has a different meaning I think)

Also, why are we using both `[5m]` and `[1h:]`?

src/robusta/core/sinks/robusta/prometheus_discovery_utils.py
```diff
@@ -39,6 +38,30 @@ def __init__(self, discovery_period_sec: int, global_config: dict):
...
+        return self._get_query_prometheus_value(query=cpu_query)
+
+    def get_cluster_avg_memory(self) -> Optional[float]:
+        memory_query = os.getenv("OVERRIDE_CLUSTER_MEM_AVG_QUERY",
+                                 f'avg_over_time(sum(container_memory_usage_bytes{{}})[1h:])')
```
arikalon1, line 48: This query also uses container and not node
I also think it returns the total memory used, and not memory utilization

src/robusta/core/sinks/robusta/robusta_sink.py
```diff
@@ -526,7 +525,9 @@ def __update_cluster_status(self):
             )
 
             self.dal.publish_cluster_status(cluster_status)
-            self.dal.publish_cluster_nodes(cluster_stats.nodes, self.__pods_running_count)
+            avg_cpu = self.__prometheus_discovery_util.get_cluster_avg_cpu()
```
arikalon1, line 528: I would put this in `try/except`, so if we have an error querying prometheus, we still publish the cluster nodes

src/robusta/core/sinks/robusta/dal/supabase_dal.py
```diff
@@ -556,12 +556,14 @@ def publish_namespaces(self, namespaces: List[NamespaceInfo]):
...
             raise
 
-    def publish_cluster_nodes(self, node_count: int, pod_count: int):
+    def publish_cluster_nodes(self, node_count: int, pod_count: int, avg_cpu: float, avg_mem: float):
         data = {
             "_account_id": self.account_id,
             "_cluster_id": self.cluster,
             "_node_count": node_count,
-            "_pod_count": pod_count,
+            "_cpu_utilization": avg_cpu,
```
arikalon1, line 564: I think both should be float, utilization percentage

arikalon1 (commented):
nice work
left some comments

src/robusta/core/sinks/robusta/robusta_sink.py
arikalon1, line 528: Will the db function work with `None`  as `cpu_utilization` or `memory_utilization`?

Avi-Robusta: yea

arikalon1 (approved):
nice work

PR #1503, opened 2024-07-18
[MAIN-1885] bumped java-toolkit

Avi-Robusta:
there are CVE's that have been removed in the python:3.12-slim base image, 
I rebuilt the docker file with the newer base image and it removed the vunerabilities


arikalon1 (approved):
LGTM

PR #1512, opened 2024-07-29
[MAIN-1909] Severity improvements

Avi-Robusta:
example test

```
globalConfig:
  custom_severity_map:
    funky: high
```
    
`   robusta playbooks trigger prometheus_alert alert_name="TestAlert"  severity=funky --namespace monitoring `

verified it updates the Severity map also on 
`robusta playbooks edit-config`


src/robusta/integrations/prometheus/models.py
```diff
@@ -20,15 +20,32 @@
...
     "warning": FindingSeverity.LOW,
+    "low": FindingSeverity.LOW,
     "info": FindingSeverity.INFO,
+    "debug": FindingSeverity.DEBUG,
 }
 
 
+class SeverityMapper:
+    @staticmethod
+    def get(value: str, default: FindingSeverity = FindingSeverity.INFO) -> FindingSeverity:
```
arikalon1, line 39: Can this be simplified?

Instead of using the GlobalConfig object, can we just add the custom mapping to the SEVERITY_MAP once when we reload config?

arikalon1 (commented):
nice work, left one comment

arikalon1 (approved):
LGTM

PR #1514, opened 2024-08-01
[MAIN-1938] added optional annotations on default playbook

Avi-Robusta:
Some users wanted alert annotations in all their prometheus alerts, so it is added by enabling 
```
globalConfig:
  show_alert_annotations: true
```
Screen Shot 2024-08-01 at 10 18 30
Screen Shot 2024-08-01 at 10 18 47

also fixed aggregation key in create_finding so it parametizes it for this specific use case ( see pictures above for example)
```
customPlaybooks:
- triggers:
  - on_prometheus_alert:
      status: "all"
  actions:
  - create_finding:
      description: 'Eng-Dev-EKS - $annotations.description'
      aggregation_key: '$labels.alertname'
```

### DIFF
--- a/playbooks/robusta_playbooks/alerts_integration.py
+++ b/playbooks/robusta_playbooks/alerts_integration.py
@@ -58,12 +58,12 @@ class SeverityParams(ActionParams):
 
 class DefaultEnricherParams(ActionParams):
     """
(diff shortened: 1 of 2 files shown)


arikalon1 (approved):
nice work

PR #1519, opened 2024-08-06
Main 1907 support holmes ask

RoiGlinik:


src/robusta/core/model/base_params.py
```diff
@@ -77,6 +77,7 @@ class ResourceInfo(BaseModel):
     kind: Optional[str]
     node: Optional[str]
     container: Optional[str]
+    cluster: Optional[str]
```
arikalon1, line 80: Since we already have the cluster, I think it's better to get it from the runner
It's less bug prone

src/robusta/core/model/base_params.py
```diff
@@ -110,6 +111,30 @@ class AIInvestigateParams(HolmesParams):
...
+
+    ask: Optional[str]
+    resource: Optional[ResourceInfo] = ResourceInfo()
+    alert_history: bool = True
+    alert_history_since_hours: float = 24
+    stored_instrucitons: bool = True
+    instructions: List[str] = []
+    include_tool_calls: bool = True
+    include_tool_call_results: bool = True
+    prompt_template: str = "builtin://generic_ask.jinja2"
```
arikalon1, line 135: Why do we need the prompt override option?
If we do need it, maybe it's better to have it None, and take the default from Holmes?

src/robusta/core/playbooks/internal/ai_integration.py
```diff
@@ -67,5 +68,53 @@ def ask_holmes(event: ExecutionBaseEvent, params: AIInvestigateParams):
...
+        else:
+            raise ActionException(ErrorCodes.HOLMES_UNEXPECTED_ERROR, "An unexpected error occured.")
+
+
+@action
+def holmes_workload_health(event: ExecutionBaseEvent, params: HolmesWorkloadHealthParams):
+    holmes_url = HolmesDiscovery.find_holmes_url(params.holmes_url)
+    if not holmes_url:
+        logging.error("Holmes url not found")
+        return
```
arikalon1, line 86: should we notify the UI somehow what happened?

arikalon1 (approved):
nice work

left few minor comments

src/robusta/core/model/base_params.py
RoiGlinik: ill remove it not sure its that necessary

arikalon1 (approved):
LGTM

PR #1545, opened 2024-09-09
removed deprecated matches config and added simulate alert link

Sheeproid:
Added a tip to simulate alert in sinks configuration and removed deprecated instructions to make the page easier to read

Screen Shot 2024-09-09 at 15 47 54

Screen Shot 2024-09-09 at 15 48 15


arikalon1 (approved):
nice work

PR #1546, opened 2024-09-09
Add service and service monitor for forwarder

itisallgood:
Related to https://github.com/robusta-dev/kubewatch/pull/86


helm/robusta/values.yaml
```diff
@@ -634,6 +634,11 @@ kubewatch:
       runAsUser: 1000
     pod: {}
   customServiceAccount: "" # to override the kubewatch service account
+  service:
+    port: 2112
+  serviceMonitor:
+    path: /metrics
+    interval: 15s
```
arikalon1, line 641: any reason to change the default?
I think we don't need an overriding value here

helm/robusta/values.yaml
```diff
@@ -586,7 +586,7 @@ image:
 
 # parameters for the robusta forwarder deployment
 kubewatch:
-  image: ~  # image can be used to override image.registry/imageName
+  image: ~ # image can be used to override image.registry/imageName
   imageName: kubewatch:v2.7.0
```
arikalon1, line 590: Don't we need a new Kubewatch image?

helm/robusta/templates/forwarder.yaml
```diff
@@ -89,4 +89,56 @@ spec:
...
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: {{ include "robusta.fullname" . }}-forwarder
+spec:
+  selector:
+    app: {{ include "robusta.fullname" . }}-forwarder
+  ports:
+    - name: forwarder-metrics
+      protocol: TCP
+      port: {{ default 2112 .Values.kubewatch.service.port }}
```
arikalon1, line 106: This is the port of the service, I think it's ok to have it as 80 (I don't even think we need a parameter for it)

helm/robusta/templates/forwarder.yaml
```diff
@@ -89,4 +89,56 @@ spec:
...
+      targetPort: {{ default 2112 .Values.kubewatch.service.port }}
```
arikalon1, line 107: This is the pod port
Why do we need to have the default both here, and in the default values file?
Can it just be in the default values file?

arikalon1 (approved):
nice work
left few small comments

helm/robusta/templates/forwarder.yaml
itisallgood: Left only default values

helm/robusta/templates/forwarder.yaml
itisallgood: Left only default values

helm/robusta/values.yaml
itisallgood: We need to change it after new kubewatch release

helm/robusta/values.yaml
itisallgood: Removed it

PR #1550, opened 2024-09-11
Added docs related to new AI user onboarding

pavangudiwada:


docs/configuration/ai-analysis.rst
```diff
@@ -124,3 +124,58 @@ To use AWS Bedrock:
...
+
+.. image:: /images/AI_Analysis_demo.png
+    :width: 1000px
+
+5. Click the "Root Cause" tab on the top. This gives you the result of an investigation done by HolmesGPT based on the alert.
+
+.. image:: /images/AI_Analysis_demo2.png
+    :width: 1000px
+
+Additionally your alerts on Slack will have an "Ask Holmes" button. Clicking it will give you the same results in the Slack channel itself.
```
aantn, line 181: There are actually some limitations here, as via `Ask Holmes` we don't send Holmes all the alert metadata (e.g. labels) because of Slack limitations on how much data you can transfer when a Slack button is clicked.

So maybe phrase it as

> Clicking it will give you results in the Slack channel itself. Note that due to technical limitations with Slack-buttons, alerts analyzed from Slack will be sent to the AI without alert-labels. For the most accurate results, it is best to use the UI.

aantn (approved):
LGTM. Left one comment.

PR #1552, opened 2024-09-12
Fixed some warnings and a redirect

pavangudiwada:
Fixes things like this 
image


aantn (approved):
Nice work. Approved with pleasure!

PR #1553, opened 2024-09-12
Fixes minor issue with generated templates having double document separator

RoryDoherty:
Moves the document separator into the if condition to avoid scenarios with multiple document separators with nothing in between. Some yaml tools cannot handle an empty doc e.g. checkov


arikalon1 (approved):
nice catch!
Thanks for fixing it @RoryDoherty

PR #1558, opened 2024-09-16
Created a new tab for notifications and routing

pavangudiwada:


aantn (approved):
Looks great. I have only one comment: The first page you see when clicking on "Notifications & Routing" seems a bit too "sudden". You land deep in the details of an advanced feature that is only supported for Slack, before you ever see the big picture or an overview.

Could we add a more general first page there? Maybe something titled "Notifications 101" or "Introduction to Notifications Routing". I think we can move [this page](https://docs.robusta.dev/notifications-tab/configuration/configuring-sinks.html) and more or less re-use it for that. (Feel free to adjust the first paragraph and title a little to make it fit better as the index page, if appropriate.)

PR #1564, opened 2024-09-20
Add ability to set Assignee and Epic on Jira sink 

daanbosch:
Added the ability to set the assignee and Epic (link) in the JIra sink config


arikalon1 (approved):
nice work!

Thanks for the contribution @daanbosch

PR #1578, opened 2024-10-04
Bug fix: We saved a copy of global config in PrometheusDiscoveryUtils

arikalon1:
This caused prometheus configuraiton not to be effected by hot reload. The runner had to be restarted for the configuration to take effect


Avi-Robusta (approved):
LGTM

PR #1596, opened 2024-10-15
Deprecating python tools

Avi-Robusta:
The python toolset image has additional cves due to gdb importing libpython3.11 
we are fully deprecating the python tools. 
They still can be used by adding the following to your generated_values.yaml

```
runner:
  additional_env_vars:
  - name: PYTHON_DEBUGGER_IMAGE
    value: debug-toolkit:v6.0
```


arikalon1 (approved):
LGTM

PR #1599, opened 2024-10-17
fix: ensure sink names are unique

nherment:
This ensures sink names are unique. For example the following sink configuration would throw a validation error because the sink name `my_sink1` is used more than once.

```yaml
sinks_config:
  - slack_sink:
      name: my_sink1
      slack_channel: my-alert-channel
      api_key: "foobar"
      scope:
        include:
          - labels:
              - "severity=high"
              - "severity=critical"

  - slack_sink:
      name: my_sink1
      slack_channel: my-notification-channel
      api_key: "foobar"
      scope:
        include:
          - labels:
              - "severity=warning"

```

### DIFF
--- a/src/robusta/core/model/runner_config.py
+++ b/src/robusta/core/model/runner_config.py
@@ -84,7 +84,6 @@ class RunnerConfig(BaseModel):
 
     @root_validator()
     def ensure_unique_sink_name(cls, val: Dict) -> Dict:
-        print(f"validating {val}")
         if val.get('sinks_config'):
             value_set = set()
             sinksConfig = cast(List[SinkConfigBase], val.get('sinks_config'))
--- a/src/robusta/core/model/runner_config.py
+++ b/src/robusta/core/model/runner_config.py
@@ -91,8 +91,8 @@ class RunnerConfig(BaseModel):
                 sink_name = sink_config.get_name()
                 if sink_name in value_set:
                     raise ValueError(f"Sink name \"{sink_name}\" is already defined. Sink names must be unique.")
-                else:
-                    value_set.add(sink_name)
+
+                value_set.add(sink_name)
         return val
 
     @validator("playbook_repos")
--- a/src/robusta/core/model/runner_config.py
+++ b/src/robusta/core/model/runner_config.py
@@ -1,5 +1,5 @@
 import base64
-from typing import Dict, List, Optional, Union
+from typing import Dict, List, Optional, Union, cast
 
 from pydantic import BaseModel, SecretStr, root_validator, validator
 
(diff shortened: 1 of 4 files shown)


src/robusta/core/model/runner_config.py
```diff
@@ -81,6 +82,20 @@ class RunnerConfig(BaseModel):
     active_playbooks: Optional[List[PlaybookDefinition]] = []
     alert_relabel: Optional[List[AlertRelabel]] = []
 
+    @root_validator()
+    def ensure_unique_sink_name(cls, val: Dict) -> Dict:
+        print(f"validating {val}")
```
itisallgood, line 87: @nherment I think you forgot to remove print

src/robusta/core/model/runner_config.py
```diff
@@ -81,6 +82,20 @@ class RunnerConfig(BaseModel):
...
+        if val.get('sinks_config'):
+            value_set = set()
+            sinksConfig = cast(List[SinkConfigBase], val.get('sinks_config'))
+            for sink_config in sinksConfig:
+                sink_name = sink_config.get_name()
+                if sink_name in value_set:
+                    raise ValueError(f"Sink name \"{sink_name}\" is already defined. Sink names must be unique.")
+                else:
```
itisallgood, line 94: You don't need `else` here because after raising a ValueError the loop stops

nherment: Thank you

src/robusta/core/model/runner_config.py
nherment: Thank you. :facepalm:

tests/test_config_validation.py
```diff
@@ -0,0 +1,24 @@
...
+        with open(FIXTURE_VALID_CONFIG_FILE_PATH) as file:
+            yaml_content = yaml.safe_load(file)
+            config = RunnerConfig(**yaml_content)
+
+    def test_duplicate_sink_names(self):
+        with open(FIXTURE_DUPL_SINKS_NAME_FILE_PATH) as file:
+            yaml_content = yaml.safe_load(file)
+            with pytest.raises(ValidationError) as validation_error:
+                config = RunnerConfig(**yaml_content)
+            assert "sink name".casefold() in str(validation_error).casefold()
```
Avi-Robusta, line 23: I would just check the whole validation string once instead of checking 'sink name' and 'unique'
f"Sink name \"my_sink1\" is already defined. Sink names must be unique."

Avi-Robusta (approved):
LGTM, Good job

src/robusta/core/model/runner_config.py
```diff
@@ -29,6 +29,7 @@
 from robusta.model.alert_relabel_config import AlertRelabel
 from robusta.model.playbook_definition import PlaybookDefinition
 from robusta.utils.base64_utils import is_base64_encoded
+from src.robusta.core.sinks.sink_config import SinkConfigBase
```
arikalon1, line 32: please change the import to `module` style, i.e.:
`robusta.code.sinks...`

Using the `src` directory causes issues when running it directly from the ide

tests/test_config_validation.py
nherment: Done

src/robusta/core/model/runner_config.py
nherment: Done, thank you for spotting this.

PR #1603, opened 2024-10-22
MAIN-2345 Support new conversation functionality

itisallgood:


src/robusta/core/model/base_params.py
```diff
@@ -171,6 +171,33 @@ class ConversationType(str, Enum):
...
 
 
+class HolmesIssueChatParamsContext(BaseModel):
+    """
+    :var investigation_result: HolmesInvestigationResult object that contains investigation saved to Evidence table by frontend for the issue.
+    :var issue_type: aggregation key of the issue
+    :var robusta_issue_id: id of the issue
+    """
+
+    investigation_result: HolmesInvestigationResult
```
arikalon1, line 181: This class is duplicated with `HolmesConversationIssueContext`

Why do we need both?
Can we use the same class, or make one of them extend the other?

src/robusta/core/playbooks/internal/ai_integration.py
```diff
@@ -243,3 +249,100 @@ def delayed_health_check(event: KubernetesAnyChangeEvent, action_params: Delayed
...
     )
+
+
+@action
+def holmes_issue_chat(event: ExecutionBaseEvent, params: HolmesIssueChatParams):
+    holmes_url = HolmesDiscovery.find_holmes_url(params.holmes_url)
+    if not holmes_url:
+        raise ActionException(ErrorCodes.HOLMES_DISCOVERY_FAILED, "Robusta couldn't connect to the Holmes client.")
+
+    conversation_title = build_conversation_title(params)
```
arikalon1, line 260: type mismatch

src/robusta/core/playbooks/internal/ai_integration.py
```diff
@@ -243,3 +249,100 @@ def delayed_health_check(event: KubernetesAnyChangeEvent, action_params: Delayed
...
+    account_id = event.get_context().account_id
+
+    try:
+        holmes_req = HolmesChatRequest(ask=params.ask, conversation_history=params.conversation_history)
+        result = requests.post(f"{holmes_url}/api/chat", data=holmes_req.json())
+        result.raise_for_status()
+        holmes_result = HolmesChatResult(**json.loads(result.text))
+
+        finding = Finding(
+            title=f"AI Ask Chat for {cluster_name} cluster {account_id} account_id",
```
arikalon1, line 324: I don't think we want to add the account id, it's just a uuid
I think the cluster is probably redundant as well

src/robusta/core/reporting/holmes.py
```diff
@@ -55,3 +70,23 @@ def __init__(
...
+class HolmesChatResult(BaseModel):
+    analysis: Optional[str] = None
+    tool_calls: Optional[List[ToolCallResult]] = None
+    conversation_history: Optional[List[dict]] = None
+
+
+class HolmesChatResultsBlock(BaseBlock):
+    holmes_result: Optional[HolmesChatResult]
+
+    def __init__(
```
arikalon1, line 84: why do we need the init function here? Aren't we getting it for free from Pydantic?

src/robusta/core/sinks/robusta/dal/model_conversion.py
```diff
@@ -86,6 +86,33 @@ def get_empty_file_object(block: EmptyFileBlock):
...
+        structured_data.append(
+            {
+                "type": "markdown",
+                "metadata": {"type": "ai_investigation_result", "createdAt": datetime_to_db_str(datetime.now())},
+                "data": Transformer.to_github_markdown(block.holmes_result.analysis),
+            }
+        )
+        ModelConversion.append_to_structured_data_tool_calls(block.holmes_result.tool_calls, structured_data)
+
+        conversation_history_block = FileBlock(
```
arikalon1, line 109: Why is this a file block?
Wouldn't a json block be better here?

arikalon1 (approved):
nice work

left a few comments

src/robusta/core/sinks/robusta/dal/model_conversion.py
itisallgood: It makes sense to me to use FileBlock as conversation_history can become pretty huge, and a few lines upper we use FileBlock to compress tool's output (just as a reminder, conversation_history includes tools and all messages)

PR #1604, opened 2024-10-22
fix: improve OOM handling for discovery process

nherment:


src/robusta/core/discovery/discovery.py
```diff
@@ -543,9 +546,17 @@ def discover_resources() -> DiscoveryResults:
         try:
             future = Discovery.executor.submit(Discovery.discovery_process)
             return future.result(timeout=DISCOVERY_PROCESS_TIMEOUT_SEC)
-        except Exception as e:
+        except BrokenProcessPool as bpp:
```
arikalon1, line 549: Feels like a little code duplication

Would a single catch clause, with a `if isinstance(e, BrokenProcessPool): ....` would be cleaner

arikalon1 (approved):
nice work

left one comment

src/robusta/core/discovery/discovery.py
nherment: I don't feel strongly either way so made the change you suggest.

PR #1605, opened 2024-10-23
fix: issue with OOM not being reported due to being captured downstream

nherment:
The code in [this PR](https://github.com/robusta-dev/robusta/pull/1604) is incorrect. The discovery process captures the error and simply restart. This means the `__discover_resources` function call should not rely on an exception being raised by `Discovery.discover_resources()`  to check the oom flag

## Front end payload example:

image


arikalon1 (approved):
LGTM

PR #1617, opened 2024-11-05
remove deprecated python and java actions from the docs

aantn:


docs/setup-robusta/alertsui.rst
```diff
@@ -1,7 +1,9 @@
-Alerts UI
-######################
+Managed Prometheus Alerts
```
aantn, line 1: @arikalon1 sorry, I meant to include this as a separate PR but if it's OK lets just review and merge it here?

arikalon1 (approved):
LGTM

PR #1621, opened 2024-11-05
fix bugs related to openshift rbad

arikalon1:
Fix bug in discovery
Add permissions to read openshift users and groups


Avi-Robusta (approved):
LGTM

PR #1622, opened 2024-11-06
update docs for AM and Prom integration

aantn:


docs/configuration/alertmanager-integration/_pull_integration.rst
```diff
@@ -3,25 +3,27 @@ Configure Metric Querying
 
 Metrics querying lets Robusta pull metrics and create silences.
 
-To configure it, add the following to ``generated_values.yaml`` and :ref:`update Robusta <Simple Upgrade>`.
+To configure it, add the ``prometheus_url`` and ``alertmanager_url`` to your Helm values and :ref:`update Robusta <Simple Upgrade>`.
```
arikalon1, line 6: I would add a note here about auto detection. i.e.:
```
If Robusta fails to auto-detect the Prometheus and Alertmanager urls, and you see related connection errors in the logs,  configure the ``prometheus_url`` and ``alertmanager_url`` to your Helm values and :ref:`update Robusta <Simple Upgrade>`
```

arikalon1 (approved):
looks good
added one comment

docs/configuration/alertmanager-integration/_pull_integration.rst
aantn: @arikalon1 I think it's best to avoid it here because this section is included in multiple pages, not just for robusta in-cluster. So it is also for external prometheuses etc when auto-detect is not relevant.

We could fix by changing the docs and not reusing snippets, but for now I suggest we just merge this as is.

Is that OK?

arikalon1: auto-detect is mostly relevant for 3rd party prometheus and alert manager

arikalon1: we try to auto detect all known prometheus distributions

aantn: ok, updated

PR #1636, opened 2024-11-18
Fix oomkill graphs for slack

itisallgood:
image


src/robusta/core/playbooks/prometheus_enrichment_utils.py
```diff
@@ -308,6 +326,14 @@ def create_chart_from_prometheus_query(
 
         chart.range = (0, max_y_value_with_padding)
 
+        # Calculate delta corresponding to 6 pixels
+        delta = (chart.range[1] - chart.range[0]) * 6 / chart.config.height
```
nherment, line 330: I'd suggest adding a good explanation for these 6 pixels otherwise it'll be very hard to understand some time from now.

nherment (approved):

Issue #1635, opened 2024-11-18
Failed to publish finding to sink ops_genie_sink


micheleliabe:
I'm trying to configure an Opsgenie sink to receive notifications, but I'm getting the following error in robusta-runner

```
2024-11-14 20:14:25.875 ERROR    Failed to publish finding to sink ops_genie_sink
Traceback (most recent call last):
  File "/app/src/robusta/core/playbooks/playbooks_event_handler_impl.py", line 317, in __handle_findings
    sink.write_finding(finding_copy, self.registry.get_sinks().platform_enabled)
  File "/app/src/robusta/core/sinks/opsgenie/opsgenie_sink.py", line 80, in write_finding
    self.__open_alert(finding, platform_enabled)
  File "/app/src/robusta/core/sinks/opsgenie/opsgenie_sink.py", line 58, in __open_alert
    details = self.__to_details(finding)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/src/robusta/core/sinks/opsgenie/opsgenie_sink.py", line 108, in __to_details
    if self.conf.extra_details_labels:
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'Configuration' object has no attribute 'extra_details_labels'
```

**Here’s my Helm configuration:**
- opsgenie_sink:
      name: ops_genie_sink
      api_key: API_KEY
      teams:
      - "exemple"

PR #1639, opened 2024-11-20
fix: issue extra details on opsgenie sink

sk4mi:
Fix for https://github.com/robusta-dev/robusta/issues/1635 

- Wrong assignment of `extra_details_labels` property
- condition was not valid from `extra_details_labels` paramater type


arikalon1 (approved):
nice work!

Thanks for the fix @sk4mi

PR #1640, opened 2024-11-24
[MAIN-2448] - Slack labels to ai

Avi-Robusta:


src/robusta/core/playbooks/internal/ai_integration.py
```diff
@@ -161,8 +161,9 @@ def holmes_workload_health(event: ExecutionBaseEvent, params: HolmesWorkloadHeal
             raise ActionException(ErrorCodes.HOLMES_UNEXPECTED_ERROR, "An unexpected error occured.")
 
 
-def build_conversation_title(params: HolmesConversationParams) -> str:
-    return f"{params.resource}, {params.ask} for issue {params.context.robusta_issue_id}"
+def build_conversation_title(params: HolmesConversationParams):
+    label_string = f"with alert labels {params.labels}" if params.labels else ""
```
arikalon1, line 165: why did you remove the returned type annotation?

Also, I think the title with label is a little weird
It'll be:
`.... with labels with alert labels ...`
Is that on purpose?

arikalon1 (dismissed):
nice work

left one small comment

src/robusta/integrations/slack/sender.py
```diff
@@ -308,6 +309,28 @@ def __send_blocks_to_slack(
                 f"error sending message to slack\ne={e}\ntext={message}\nchannel={channel}\nblocks={*output_blocks,}\nattachment_blocks={*attachment_blocks,}"
             )
 
+
+    def __limit_labels_size(self, labels: dict, max_size: int = 1000) -> dict:
+        # slack can only send 2k tokens in a callback so the labels are limited in size
+
+        low_priority_labels = ["job", "prometheus", "severity", "service"]
+        limited_labels = copy.deepcopy(labels)
+        current_length = len(str(limited_labels))
```
moshemorad, line 318: worth to check the size before the copy etc... no need to copy if the size is under the threshold. No?

src/robusta/integrations/slack/sender.py
```diff
@@ -308,6 +309,28 @@ def __send_blocks_to_slack(
...
+        for key in low_priority_labels:
+            if current_length <= max_size:
+                break
+            if key in limited_labels:
+                del limited_labels[key]
+                current_length = len(str(limited_labels))
+
+        while current_length > max_size and limited_labels:
+            limited_labels.pop(next(iter(limited_labels)))
+            current_length = len(str(limited_labels))
```
moshemorad, line 330: Do we really want to convert the dict to str on every iteration? We can't just assume what would be the length of the string after the pop since we know the key & value length?

moshemorad (changes requested):

moshemorad (approved):

PR #1645, opened 2024-11-27
feat: Implemented Incident.io Sink

igorhrcek:
Here is implementation of Incident.io Sink which uses Alert Events V2 to send events to Incidenti.io

### DIFF
--- a/src/robusta/core/sinks/incidentio/incidentio_sink.py
+++ b/src/robusta/core/sinks/incidentio/incidentio_sink.py
@@ -50,13 +50,6 @@ class IncidentioSink(SinkBase):
                     "href": finding.get_investigate_uri(self.account_id, self.cluster_name),
                 }
             )
-        else:
-            links.append(
-                {
-                    "text": "🔎 Enable Robusta UI to investigate",
-                    "href": "https://bit.ly/robusta-ui-incidentio",
-                }
-            )
 
         # Collect metadata
         metadata["resource"] = finding.subject.name
--- a/src/robusta/core/sinks/incidentio/incidentio_sink.py
+++ b/src/robusta/core/sinks/incidentio/incidentio_sink.py
@@ -1,10 +1,9 @@
 import logging
 from typing import Optional, Dict, List, Any
-from urllib.parse import urljoin
 from robusta.core.sinks.incidentio.incidentio_client import IncidentIoClient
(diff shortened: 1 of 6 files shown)


src/robusta/core/sinks/incidentio/incidentio_sink.py
```diff
@@ -0,0 +1,115 @@
...
+    TableBlock,
+    KubernetesDiffBlock,
+)
+from robusta.core.sinks.sink_base import SinkBase
+from robusta.core.sinks.sink_config import SinkConfigBase
+from robusta.core.sinks.incidentio.incidentio_sink_params import IncidentioSinkParams
+
+
+class IncidentioSink(SinkBase):
+    def __init__(self, sink_config: SinkConfigBase, registry):
```
arikalon1, line 25: I think the `sink_config` type here should be the concrete type, `IncidentioSinkConfigWrapper`
The base class doesn't have the attribute `incidentio_sink`

src/robusta/core/sinks/incidentio/incidentio_sink.py
```diff
@@ -0,0 +1,115 @@
...
+                {
+                    "text": "🔎 Investigate in Robusta",
+                    "href": finding.get_investigate_uri(self.account_id, self.cluster_name),
+                }
+            )
+        else:
+            links.append(
+                {
+                    "text": "🔎 Enable Robusta UI to investigate",
+                    "href": "https://bit.ly/robusta-ui-incidentio",
```
arikalon1, line 57: Do we really want this else clause? 
I checked, and this link `https://bit.ly/robusta-ui-incidentio` doesn't have anything

arikalon1 (commented):
thanks for the PR @igorhrcek 

Left 2 minor comments

src/robusta/core/sinks/incidentio/incidentio_sink.py
igorhrcek: Fair point, I removed it.

src/robusta/core/sinks/incidentio/incidentio_sink.py
igorhrcek: Yeah, you are right, I've updated my implementation.

arikalon1 (approved):
nice work

Thanks for the contribution @igorhrcek !

PR #1646, opened 2024-11-28
helm: configurable Service annotations

pznamensky:
Hi guys,
In our infrastructure, we need to add custom annotations to k8s services in order to configure load balancers. Currently we have to patch Service resource after installing Robusta which is not very convenient.
As it seems to be a common case, I think it would be worthwhile to add the corresponding helm values into the chart.

### DIFF
--- a/helm/robusta/values.yaml
+++ b/helm/robusta/values.yaml
@@ -35,6 +35,11 @@ globalConfig:
   custom_annotations: []
   custom_severity_map: {}
 
+# k8s service config
+service:
+  # service annotations
+  annotations: {}
+
 # see https://docs.robusta.dev/master/user-guide/configuration/additional-settings.html#relabel-prometheus-alerts
 alertRelabel: []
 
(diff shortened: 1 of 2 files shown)


helm/robusta/values.yaml
```diff
@@ -35,6 +35,11 @@ globalConfig:
   custom_annotations: []
   custom_severity_map: {}
 
+# k8s service config
+service:
```
arikalon1, line 39: Do you think we can move this into the `runner` section?
Since these annotations are for the `runner` service, I think it might belong there

arikalon1 (commented):
Thanks for the PR @pznamensky 
Nice work

Left a small comment in the review

helm/robusta/values.yaml
pznamensky: `runner` section indeed seems to be a better place for that.
Thanks for taking a look!

PR #1655, opened 2024-12-08
bug fix: include scope with multiple label values, was evaluating onl…

arikalon1:
…y the first one

In addition, if there was an error in sending to a sink, the "stop" wasn't respected


Avi-Robusta (approved):
LGTM

aantn (approved):

PR #1665, opened 2024-12-21
MAIN-2389 Stream actions

RoiGlinik:


src/robusta/core/playbooks/internal/ai_integration.py
```diff
@@ -62,35 +62,43 @@ def ask_holmes(event: ExecutionBaseEvent, params: AIInvestigateParams):
...
-                container=params.resource.container if params.resource else "",
-            ),
-            finding_type=FindingType.AI_ANALYSIS,
-            failure=False,
-        )
-        finding.add_enrichment(
-            [HolmesResultsBlock(holmes_result=holmes_result)], enrichment_type=EnrichmentType.ai_analysis
-        )
+        if params.stream:
+            with requests.post(f"{holmes_url}/api/stream/investigate", data=holmes_req.json(), stream=True) as resp:
```
moshemorad, line 67: Just to make sure:
A. The whole info we adding to the Finding & Enrichment in the none stream implementation isn't used by the UI?
B. We added couple of weeks ago logic that the UI save the RCA results history. Is that still working us expected?

moshemorad (approved):

PR #1677, opened 2025-01-05
[MAIN-2842] added kubectl action in runner

Avi-Robusta:
Screen Shot 2025-01-05 at 14 43 59
example config

```
customPlaybooks:
- actions:
  - create_finding:
      aggregation_key: "DeploymentScaledDown"
      title: "Pod $name on namespace $namespace crashed 3 times"
  - kubectl_command:
      command: kubectl scale --replicas=0 deployment/$(kubectl get pod $name -n $namespace -o jsonpath='{.metadata.ownerReferences[?(@.kind=="ReplicaSet")].name}' | sed 's/-[a-f0-9]\{9,10\}$//') -n $namespace
  name: DeploymentScaledDown
  triggers:
  - on_pod_crash_loop:
      restart_count: 3
      
runner:
  customClusterRoleRules: 
  - apiGroups: ["apps"]
    resources: ["deployments/scale"]
    verbs: ["get", "list", "watch", "update", "patch"]
```


playbooks/robusta_playbooks/kubectl_enrichments.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+class KubectlParams(PodRunningParams):
+    """
+    :var kubectl_command: The full kubectl command to run, formatted as a shell command string.
+    :var description: A description of the command ran.
+    :var timeout: The maximum time (in seconds) to wait for the kubectl command to complete. Default is 3600 seconds.
+    """
+
+    command: str = None  # type: ignore
```
moshemorad, line 27: Why type ignore? Just need to wrap str with Optional no?

playbooks/robusta_playbooks/kubectl_enrichments.py
```diff
@@ -0,0 +1,77 @@
...
+@action
+def kubectl_command(event: ExecutionBaseEvent, params: KubectlParams):
+    """
+    Runs a custom kubectl command inside a Kubernetes pod using a Job.
+    """
+
+    subject = event.get_subject()
+    formatted_kubectl_command = format_event_templated_string(subject, params.command)
+    
+    logging.debug(f"kubectl_command running '{formatted_kubectl_command}'")
```
moshemorad, line 41: Please change it to info, I think that this log is something that we alway see since it might have effect for users.

playbooks/robusta_playbooks/kubectl_enrichments.py
```diff
@@ -0,0 +1,77 @@
...
+                imagePullPolicy="Always",
+                command=["/bin/sh", "-c"],
+                args=[formatted_kubectl_command]
+            )
+        ],
+        restartPolicy="Never",
+    )
+
+    try:
+        kubectl_response = RobustaJob.run_simple_job_spec(
```
moshemorad, line 58: Did we consider using Ephemeral containers? https://kubernetes.io/docs/concepts/workloads/pods/ephemeral-containers/

playbooks/robusta_playbooks/kubectl_enrichments.py
```diff
@@ -0,0 +1,77 @@
...
+            spec,
+            f"debug-kubectl-{str(uuid.uuid4())}",
```
moshemorad, line 60: debug? maybe change to something like robusta-kubectl-command-uuid? Make it more clear. WDYT?

playbooks/robusta_playbooks/kubectl_enrichments.py
```diff
@@ -0,0 +1,77 @@
...
+            params.timeout,
+            custom_annotations=params.custom_annotations,
+            ttl_seconds_after_finished=43200,  # 12 hours
```
moshemorad, line 63: Can we add all those args to `KubectlParams`? I think it is good idea letting the user decide if he want to override those.

moshemorad (changes requested):

playbooks/robusta_playbooks/kubectl_enrichments.py
Avi-Robusta: we dont want to always have this incase there is sensitive information there

playbooks/robusta_playbooks/kubectl_enrichments.py
Avi-Robusta: No reason to implement a whole new system for this when we have this working, maybe if we specifically need this in the future we will

playbooks/robusta_playbooks/kubectl_enrichments.py
Avi-Robusta: its not optional, whats a kubectl_command without the actual kubectl command to run?

playbooks/robusta_playbooks/kubectl_enrichments.py
moshemorad: In that case is it fine in debug? maybe we can instead log the description?

playbooks/robusta_playbooks/kubectl_enrichments.py
moshemorad: A. I think that it is something that possible for users to want to change, don't remove the job after it run or change ttl.
B. It is small effort no? if no then ignore if yes let's just do it.
C. From my point of view it is very confusing we are passing ttl but we anyway delete the job when it is finished.

playbooks/robusta_playbooks/kubectl_enrichments.py
Avi-Robusta: debug should be fine

playbooks/robusta_playbooks/kubectl_enrichments.py
Avi-Robusta: opened ticket

moshemorad (approved):

PR #1683, opened 2025-01-08
add a playbook that sends k8s resource manifests, as json, to a speci…

arikalon1:
…fied url.


playbooks/robusta_playbooks/babysitter.py
```diff
@@ -79,3 +82,27 @@ def resource_babysitter(event: KubernetesAnyChangeEvent, config: BabysitterConfi
...
+    """
+
+    url: str
+    headers: Optional[Dict[str, str]] = None
+
+
+@action
+def json_change_tracker(event: KubernetesAnyChangeEvent, params: UrlParam):
+    """
+    post change json to the specified url
```
Avi-Robusta, line 99: I would change this since it shows up in the docs
```
Posts Kubernetes resource changes as JSON to a specified URL.
```

playbooks/robusta_playbooks/babysitter.py
```diff
@@ -79,3 +82,27 @@ def resource_babysitter(event: KubernetesAnyChangeEvent, config: BabysitterConfi
...
+    this action doesn't create a finding
+    """
+    try:
+        event_dict = event.obj.to_dict()
```
Avi-Robusta, line 103: is this enough?
do we want event.obj and event.old_obj to see the changes? or only the new one?
event.filtered_diffs, event.obj_filtered, event.old_obj_filtered?

If they only want the json of the new object and not the changes wouldn't it make more sense to use the webhook sink and then make a get_resource_json action like what we have for get_resource_yaml?

playbooks/robusta_playbooks/babysitter.py
arikalon1: good point, fixing

playbooks/robusta_playbooks/babysitter.py
arikalon1: yes, in this case they just want to resource json, without the finding details
so this simple solution works best here

PR #1687, opened 2025-01-13
Fix Opsgenie extra details labels iteration

sk4mi:


arikalon1 (approved):
nice work, thanks for the fix @sk4mi

PR #1705, opened 2025-01-23
[MAIN-2678] Fixing missing slack sink workload

Avi-Robusta:
if the workload does not exist, like the pod or deployment is deleted it falls back to using the prometheus labels for sinc matching

### DIFF
--- a/src/robusta/core/discovery/top_service_resolver.py
+++ b/src/robusta/core/discovery/top_service_resolver.py
@@ -1,7 +1,7 @@
 import threading
 import time
 from collections import defaultdict
-from typing import Dict, List, Optional
+from typing import Any, Dict, List, Optional
 
 from pydantic.main import BaseModel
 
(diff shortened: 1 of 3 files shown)


src/robusta/core/sinks/slack/slack_sink.py
```diff
@@ -42,7 +42,7 @@ def handle_notification_grouping(self, finding: Finding, platform_enabled: bool)
             investigate_uri = self.get_timeline_uri(self.account_id, self.cluster_name)
             finding_data = finding.attribute_map
             # The top level entity name (the owner of the pod etc)
-            finding_data["workload"] = finding.service.name if finding.service else None
+            finding_data["workload"] = finding.service.name if finding.service else finding.backup_workload_name
```
arikalon1, line 45: why are we doing it only for Slack?
it's probably useful on other places as well

src/robusta/core/discovery/top_service_resolver.py
```diff
@@ -67,6 +67,19 @@ def guess_cached_resource(cls, name: str, namespace: str) -> Optional[TopLevelRe
                 return cached_resource
         return None
 
+    @classmethod
+    def guess_workload_from_labels(cls, labels: Dict[Any, Any] = None) -> Optional[TopLevelResource]:
```
arikalon1, line 71: I think this should be part of the `guess_cached_resource` method
I would just add this logic, in case the existing logic didn't work.
This way, it will also be available to all other places using the returned value

arikalon1 (commented):
nice work
left a small comment

src/robusta/core/sinks/slack/slack_sink.py
Avi-Robusta: i saw that `workload` only exists for slack, we can do this for the robusta sink if you think too

src/robusta/core/discovery/top_service_resolver.py
Avi-Robusta: on it

PR #1709, opened 2025-01-30
explain value of Robusta-UI in the Robsuta-UI sink and help users start

Sheeproid:
See https://docs.robusta.dev/docs-ui-sink-demo/configuration/sinks/RobustaUI.html


aantn (approved):
Looks good!

PR #1715, opened 2025-02-06
Added ActionException `This feature is not available in free accounts` for Holmes

itisallgood:


arikalon1 (commented):
@itisallgood this is not required
We don't need any code in the runner for this

please close this PR

PR #1726, opened 2025-02-12
Redirced to platform only if enabled

moshemorad:


docs/configuration/sinks/ms-teams.rst
```diff
@@ -77,3 +77,20 @@ Example:
             name: main_ms_teams_sink
             webhook_url: teams-incoming-webhook  # see instructions below
             webhook_override: ${annotations.kubernetes.io/service-name}
+
+Redirect to Platform
+-------------------------------------------------------------------
+
+Some buttons in ms-teams are redirecting to Robusta UI by default, this could be change to redirect to
```
aantn, line 84: Suggested

> By default, MS Teams notifications include buttons to view more information in the Robusta SaaS platform. If you don't use Robusta SaaS you can modify these links to point at Prometheus instead. To do so, set ``prefer_redirect_to_platform: false``.

docs/configuration/sinks/slack.rst
```diff
@@ -95,6 +95,25 @@ Example:
          slack_channel: my-fallback-channel
          channel_override: "$cluster_name-alerts-$labels.env-${annotations.kubernetes.io/service-name}"
 
+Redirect to Platform
+-------------------------------------------------------------------
+
+Some button in slack are redirecting to Robusta UI by default, this could be change to redirect to
```
aantn, line 101: Same comment as above (replacing teams with slack)

aantn (approved):

PR #1755, opened 2025-03-18
Removed MEDIUM severity from runner

itisallgood:


docs/configuration/exporting/exporting-data.rst
```diff
@@ -333,7 +333,7 @@ Each alert in the ``alerts`` list must follow the specific schema, which include
      - The source of the alert.
      - Yes
    * - ``priority``
-     - string (one of: ``critical``, ``high``, ``medium``, ``error``, ``warning``, ``info``, ``low``, ``debug``)
+     - string (one of: ``critical``, ``high``, ``error``, ``warning``, ``info``, ``low``, ``debug``)
```
moshemorad, line 336: Do we really also want to remove it from here? This is used to load alerts from other systems and they might have this level. Shouldn't we only convert it in our side? Could it break already using customers?

arikalon1 (approved):
nice work

PR #1763, opened 2025-03-24
Improve instructions on sanitizing logs

aantn:
To assume less familiarity with robusta internals (actions etc)


pavangudiwada (approved):
LGTM

PR #1771, opened 2025-03-28
ROB-1005: toolset prometheus/metrics now returns data by default

nherment:
Document that the prometheus/metrics toolset now returns data to Holmes by default.
Also split the optional config from the basic config so that users do not copy default values when setting up prometheus.


nherment:
associated with https://github.com/robusta-dev/holmesgpt/pull/356

arikalon1 (approved):

PR #1787, opened 2025-04-18
Added changes for tools result output for holmes

itisallgood:


src/robusta/core/sinks/robusta/dal/model_conversion.py
```diff
@@ -89,7 +89,13 @@ def get_empty_file_object(block: EmptyFileBlock):
...
     def append_to_structured_data_tool_calls(tool_calls: List[ToolCallResult], structured_data) -> None:
         for tool_call in tool_calls:
-            file_block = FileBlock(f"{tool_call.description}.txt", tool_call.result.encode())
+            
+            if isinstance(tool_call.result, str):
+                result_content = tool_call.result.encode()
+            elif isinstance(tool_call.result, (dict, list)):
+                result_content = json.dumps(tool_call.result, indent=2).encode()
+               
+            file_block = FileBlock(f"{tool_call.description}.txt", result_content)
```
nherment, line 98: Is this the right thing to do when we now have structured output which has metedata as well as the file content?
It seems to me we would be better off with a ToolBlock or something like that. It will make it both cleaner and easier for the clients (UI & relay) in differentiating tool calls that are raw output  (i.e. FileBlock) from the new structured result.

nherment (changes requested):
I'm not sure about these changes. I think it's better to have a dedicated structure for the new tool results

src/robusta/core/sinks/robusta/dal/model_conversion.py
nherment: Sounds good. This will keep the coralogix tool call URL broken for one of our clients though. We could equally do this on Holmes which would make it possible to continue releasing holmes alpha versions without a new robusta version

PR #1795, opened 2025-04-27
[ROB-1179] - slack file upload bugfix

Avi-Robusta:
We improved the Slack file upload process to be more resilient:

1. If the filesystem is not writable, we automatically generate the file in memory using a `SpooledTemporaryFile`.
2. If uploading a file fails, it no longer blocks the creation of the finding or the upload of additional files.

In the image, you can see four test cases executed by the Slack file uploader:

1. Successful upload under normal conditions.
2. Upload with a simulated `NamedTemporaryFile` failure.
3. (and 4) Uploading a finding where both `NamedTemporaryFile` and `SpooledTemporaryFile` fail.

Screen Shot 2025-05-04 at 14 15 31


src/robusta/integrations/slack/sender.py
```diff
@@ -213,16 +219,37 @@ def __to_slack(self, block: BaseBlock, sink_name: str) -> List[SlackBlock]:
...
 
-    def __upload_file_to_slack(self, block: FileBlock, max_log_file_limit_kb: int) -> str:
+    def __upload_file_to_slack(self, block: FileBlock, max_log_file_limit_kb: int) -> Optional[str]:
+        """Upload a file to Slack and return a permalink to it."""
         truncated_content = block.truncate_content(max_file_size_bytes=max_log_file_limit_kb * 1000)
+        filename = block.filename
 
-        """Upload a file to slack and return a link to it"""
-        with tempfile.NamedTemporaryFile() as f:
+        def _upload_temp_file(file_reference) -> Optional[str]:
```
moshemorad, line 227: Why we need this as inner function? any good reason for that? If not please move it outside of the function.

src/robusta/integrations/slack/sender.py
```diff
@@ -213,16 +219,37 @@ def __to_slack(self, block: BaseBlock, sink_name: str) -> List[SlackBlock]:
...
+            result = self.slack_client.files_upload_v2(
+                title=filename,
+                file_uploads=[{"file": file_reference, "filename": filename, "title": filename}],
+            )
             return result["file"]["permalink"]
 
+        try:
+            with tempfile.NamedTemporaryFile() as f:
+                logging.debug("Trying NamedTemporaryFile for Slack upload")
+                return _upload_temp_file(f.name)
```
moshemorad, line 242: How this works? You are passing the name which is string. am i missing something?

src/robusta/integrations/slack/sender.py
Avi-Robusta: removing duplicate calls of the same 5 lines for each

src/robusta/integrations/slack/sender.py
```diff
@@ -213,16 +219,37 @@ def __to_slack(self, block: BaseBlock, sink_name: str) -> List[SlackBlock]:
...
+            f.seek(0)
+
+            result = self.slack_client.files_upload_v2(
+                title=filename,
+                file_uploads=[{"file": file_reference, "filename": filename, "title": filename}],
+            )
             return result["file"]["permalink"]
 
+        try:
+            with tempfile.NamedTemporaryFile() as f:
```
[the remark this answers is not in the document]
Avi-Robusta: spooledfile  is a memory file, relying on creating files just in memory is asking for OOMs imo so it should be a last resort.

src/robusta/integrations/slack/sender.py
Avi-Robusta: string of the location in the filesystem - full filepath

src/robusta/integrations/slack/sender.py
moshemorad: But why it is inner function? why not have it as function on the class?

src/robusta/integrations/slack/sender.py
moshemorad: so it uses f as inner function arg instead of passing it to the inner function? It just harder to understand what is going on here and why we have done it that way. Please move the inner function our and pass the file as arg.

src/robusta/integrations/slack/sender.py
```diff
@@ -235,7 +262,8 @@ def prepare_slack_text(self, message: str, max_log_file_limit_kb: int, files: Li
                 if len(file_block.contents) == 0:
                     continue
                 permalink = self.__upload_file_to_slack(file_block, max_log_file_limit_kb=max_log_file_limit_kb)
-                uploaded_files.append(f"* <{permalink} | {file_block.filename}>")
+                if permalink:
```
moshemorad, line 265: Don't we want to show message in the slack that file is missing due to error?

tests/utils/slack_utils.py
```diff
@@ -26,6 +28,12 @@ def _create_or_join_channel(client: WebClient, channel_name: str) -> str:
         for channel in client.conversations_list()["channels"]:
             if channel["name"] == channel_name:
                 # TODO: join the channel if necessary
```
moshemorad, line 30: Implemented? if so remove the TODO?

tests/test_slack.py
```diff
@@ -1,4 +1,10 @@
-from robusta.api import Finding, MarkdownBlock, SlackSender, TableBlock
+import base64
```
moshemorad, line 1: Remove unused.

src/robusta/integrations/slack/sender.py
```diff
@@ -213,16 +219,37 @@ def __to_slack(self, block: BaseBlock, sink_name: str) -> List[SlackBlock]:
...
             return result["file"]["permalink"]
 
+        try:
+            with tempfile.NamedTemporaryFile() as f:
+                logging.debug("Trying NamedTemporaryFile for Slack upload")
+                return _upload_temp_file(f.name)
+        except Exception as e:
+            logging.debug(f"NamedTemporaryFile failed: {e}")
+        try:
+            with tempfile.SpooledTemporaryFile(max_size=max_log_file_limit_kb * 1000) as f:
```
moshemorad, line 246: Are we sure this max_size will always fit? Littlebit risky. Why we don't keep it without limit?

logic in other places can change and we will miss it here so eventrually it will try to write to disk again.

tests/test_slack.py
```diff
@@ -48,3 +54,87 @@ def test_long_table_columns(slack_channel: SlackChannel):
...
+    slack_params = SlackSinkParams(name="test_slack", slack_channel=slack_channel.channel_name, api_key="")
+
+    # verify NamedTemporaryFile sending works
+    # verify SpooledTemporaryFile sending works
+    with patch("tempfile.NamedTemporaryFile", side_effect=FileNotFoundError("No usable temporary directory found")):
+        slack_sender.send_finding_to_slack(finding, slack_params, False)
+
+    # Verify that the message contains the finding title but not the file content
+    latest_message = slack_channel.get_latest_message()
+    assert "Test Text File Upload" in latest_message
```
moshemorad, line 78: Please add assert that the file content indeed sent in both caess.

moshemorad (changes requested):

src/robusta/integrations/slack/sender.py
Avi-Robusta: again, a limitless file size could cause oom to robusta which is worse than not sending a finding, do you want to set the max size to something large but manageable like 50Mb?

src/robusta/integrations/slack/sender.py
Avi-Robusta: what do you have in mind?

src/robusta/integrations/slack/sender.py
Avi-Robusta: got it

tests/test_slack.py
Avi-Robusta: The Slack message looks like this:
`Test Text File Upload * <https://robustaco.slack.com/files/U020QD7TF0X/F08PGU349ST/test.txt | test.txt>`

The file content itself is not included in the message — only a link to the uploaded file.

From what I can find, the Slack API does not natively support downloading file contents directly. We would need to implement that functionality ourselves by fetching the file via its private URL, using the correct authorization headers.

Since this would require additional implementation work (and updating our Slack tokens with the required file access permissions), I would recommend creating a separate task for it if we want to start verifying file contents automatically in Slack tests.

That said, I have manually verified that the correct file content was uploaded, so I wouldn't block this PR from getting merged based on this point.

src/robusta/integrations/slack/sender.py
Avi-Robusta: In my opinion, having the error logged is sufficient.
If someone looks for the file and doesn't see it, they'll naturally check the runner logs, find the error, and be able to fix the issue.

Adding an error message directly into the Slack message would make it look messy and could give the impression that the system is unstable.

src/robusta/integrations/slack/sender.py
moshemorad: So just lets decide on putting somthing like 20mb so other files which are not log files will also have some buffer before try to write to disk again.

tests/test_slack.py
```diff
@@ -48,3 +55,83 @@ def test_long_table_columns(slack_channel: SlackChannel):
...
+    # Mock NamedTemporaryFile to raise an exception
+    with patch("tempfile.NamedTemporaryFile", side_effect=FileNotFoundError("No usable temporary directory found")):
+        with patch("tempfile.SpooledTemporaryFile", side_effect=FileNotFoundError("Cant create spooled file")):
+            slack_sender.send_finding_to_slack(finding, slack_params, False)
+
+        # Verify that the message contains the finding title but not the file content
+        latest_message = slack_channel.get_latest_message()
+        assert TEST_FINDING_TITLE in latest_message
+        assert TEST_FILE_NAME not in latest_message  # File should not be included
+        assert TEST_FILE_CONTENT not in latest_message  # File content should not be included
```
moshemorad, line 123: If not doing anything please remove.

moshemorad (changes requested):

src/robusta/integrations/slack/sender.py
Avi-Robusta: added message according to thread

moshemorad (approved):

Issue #1781, opened 2025-04-10
Support Mimir/Alertmanager/Loki with multiple tenants


samui-aug:
**Is your feature request related to a problem?**
Our current centralized mimir/loki cluster setup is as follow
1. Centralized Mimir
2. Centralized Loki
3. Centralized Mimir Alertmanager (support multiple tenants)

Then individual workload cluster will have assigned tenants that they remote write to using specific `X-Scope-OrgID`

Currently, there's no way to specify `X-Scope-OrgID` to robusta-runners to be able to query correct Mimir tenants so with the particular setup we can't query prometheus and we cannot send simulated/silence alert to correct alertmanager tenant as well

**Describe the solution you'd like**
A way to add custom header through robusta-runner via `globalConfig` or Environment variable for any source that support `X-Scope-OrgID` such as
1. Prometheus (Mimir)
2. Alertmanager
3. Loki

**Describe alternatives you've considered**
There's no workaround, if we can't specify tenant, robusta will not get any data from Mimir

arikalon1:
thanks for the update @samui-aug !

PR #1797, opened 2025-05-05
ROB-1126: headers support for centralized prometheus

nherment:


nherment:
Partially fixes https://github.com/robusta-dev/robusta/issues/1781

moshemorad (approved):

PR #1798, opened 2025-05-05
ROB-1033 custom owners

RoiGlinik:


src/robusta/core/model/env_vars.py
```diff
@@ -135,3 +135,6 @@ def load_bool(env_var, default: bool):
 
 # simple calculated values (not direct environment vars)
 SENTRY_ENABLED = SEND_ADDITIONAL_TELEMETRY and SENTRY_DSN
+
+# enable custom CRDs supported by robusta "["StrimziPodSet", "Cluster"]"
```
arikalon1, line 139: Should this be the same values: ExecutionContext, StrimziPodSet, CNPGCluster ?

src/robusta/integrations/kubernetes/custom_crds.py
```diff
@@ -0,0 +1,130 @@
...
+    ready_pods_path: ClassVar[str] = ""
+
+    metadata: ObjectMeta
+    spec: dict
+    status: dict
+    apiVersion: str
+    kind: str
+
+    @classmethod
+    def readNamespaced(cls, name: str, namespace: str):
```
arikalon1, line 25: This assumes all CRDs are namespaced resource - but I'm not sure that's the case
I don't think we should change it though

src/robusta/integrations/kubernetes/custom_crds.py
```diff
@@ -0,0 +1,130 @@
...
+            {"items": [cls(**crd) for crd in crd_res.get("items", [])]},
+        )()
+
+        return crd_list
+
+
+# https://github.com/orgs/strimzi/discussions/9140 // restarts
+# https://github.com/strimzi/strimzi-kafka-operator/blob/main/helm-charts/helm3/strimzi-kafka-operator/crds/042-Crd-strimzipodset.yaml
+@dataclass
+class StrimziPodSet(CRDBase, HikaruDocumentBase, HikaruCRDDocumentMixin):
```
arikalon1, line 79: Looks like the distance to create something completely configurable is very very small!

arikalon1 (approved):
Very nice work

I think you're 1 hour work from making it completely generic 
Left some minor comments - feel free to merge

PR #1799, opened 2025-05-05
Slack sink replies to a thread with holmes analysis

itisallgood:
## Summary by CodeRabbit

- **New Features**
  - Added support for passing Slack thread and channel context through actions, enabling more precise message threading and context in Slack notifications.
- **Documentation**
  - Updated documentation links in configuration comments to point to new locations in the official documentation.
- **Style**
  - Minor whitespace and formatting improvements for code readability.


moshemorad:
@coderabbitai please review

moshemorad:
@coderabbitai full review

src/robusta/integrations/slack/sender.py
```diff
@@ -568,25 +568,45 @@ def send_holmes_analysis(
...
             logging.exception(f"error sending message to slack. {title}")
+    
+    def _resolve_slack_thread(
+        self,
+        finding: Finding,
+        sink_params: SlackSinkParams,
+        thread_ts: Optional[str] = None,
+    ) -> tuple[str, Optional[str]]:
+        
+        channel = ChannelTransformer.template(
```
moshemorad, line 579: A. Any reason to call this if channel override was provided?
B. Not sure i fully understand how this will work if ChannelTransformer.template will work correctly if labels & annotations aren't pass in that flow.

src/robusta/core/playbooks/internal/ai_integration.py
```diff
@@ -112,7 +112,13 @@ def ask_holmes(event: ExecutionBaseEvent, params: AIInvestigateParams):
             finding.add_enrichment(
                 [HolmesResultsBlock(holmes_result=holmes_result)], enrichment_type=EnrichmentType.ai_analysis
             )
-
+            runner_context = getattr(params, "robusta_context", None)
```
moshemorad, line 115: Looks like we are ignoring the sink param in this flow. It might cause issues with sending the finding to other sinks as well. I think we need to support the sink param in this flow as well.

helm/robusta/values.yaml
moshemorad, line 1: Add tests

src/robusta/core/model/base_params.py
```diff
@@ -83,6 +83,10 @@ class ResourceInfo(BaseModel):
 class HolmesParams(ActionParams):
     holmes_url: Optional[str]
     model: Optional[str]
+    # Additional internal context that helps runner to send investigation to appropriate sinks
+    # for now it is used only for passing thread_ts to slack sink internally;
+    robusta_context: Optional[Dict[str, Any]] = None
```
moshemorad, line 88: Thinking here loud, we already have SlackCallbackParams which being used in other flows. I think it will be good idea to add relevant info there and reuse in relevant flows.
WDYT?

moshemorad (changes requested):

PR #1809, opened 2025-05-08
ROB-1268: Alert limitation

arikalon1:


src/robusta/integrations/slack/sender.py
```diff
@@ -411,6 +413,11 @@ def __create_finding_header(
         self, finding: Finding, status: FindingStatus, platform_enabled: bool, include_investigate_link: bool
     ) -> MarkdownBlock:
         title = finding.title.removeprefix("[RESOLVED] ")
+
+        match = MENTION_PATTERN.search(title)
+        mention = f" {match.group(0)}" if match else ""
+        title = MENTION_PATTERN.sub("", title).strip()
```
moshemorad, line 419: Shouldn't this only be done if there is match?

src/robusta/integrations/slack/sender.py
```diff
@@ -411,6 +413,11 @@ def __create_finding_header(
         self, finding: Finding, status: FindingStatus, platform_enabled: bool, include_investigate_link: bool
     ) -> MarkdownBlock:
         title = finding.title.removeprefix("[RESOLVED] ")
+
+        match = MENTION_PATTERN.search(title)
+        mention = f" {match.group(0)}" if match else ""
```
moshemorad, line 418: What happen if there are more than 1 mention?

src/robusta/integrations/slack/sender.py
```diff
@@ -426,7 +433,7 @@ def __create_finding_header(
             title = f"<{finding.get_investigate_uri(self.account_id, self.cluster_name)}|*{title}*>"
         return MarkdownBlock(
             f"""{status_name} {sev.to_emoji()} *{sev.name.capitalize()}*
-{title}"""
+{title}{mention}"""
```
moshemorad, line 436: Can you please add test for this use case? Thanks

moshemorad (dismissed):

src/robusta/integrations/slack/sender.py
arikalon1: fixed

src/robusta/integrations/slack/sender.py
arikalon1: fixed

src/robusta/integrations/slack/sender.py
arikalon1: fixed

tests/slack/test_slack_mentions.py
```diff
@@ -0,0 +1,20 @@
...
+def test_title_with_single_mention():
+    title = "Title with single mention <@U024BE7LH>"
+    new_title, mention = SlackSender.extract_mentions(title)
+    assert new_title == "Title with single mention"
+    assert mention == " <@U024BE7LH>"
+
+def test_title_with_two_mentions():
+    title = "Title with <@U024BE7LG> two mentions <@U024BE7LH>"
+    new_title, mention = SlackSender.extract_mentions(title)
+    assert new_title == "Title with  two mentions"
```
moshemorad, line 19: Are we fine with the double space in that case? Just double checking that you noticed it and ok with that.

moshemorad (dismissed):

tests/slack/test_slack_mentions.py
arikalon1: I think it's fine

moshemorad (approved):

PR #1815, opened 2025-05-13
Switch robusta installation docs to pipx

aantn:
The current pip instructions don't work on my machine. I think users will be better off trying pipx by default. 
I've also improved the explanation on why all this is necessary


pavangudiwada (approved):
LGTM

PR #1818, opened 2025-05-14
Readme integrations to docs update

pavangudiwada:
Part of updating the Readme, will add more changes in other PR's. 

Make it easy for users to understand which of their tools we support, and also get them to relevant docs faster. 

Note: Installation options is commented for now (waiting for some suggestions about it, but that need not delay this PR)

CleanShot 2025-05-14 at 12 27 03


aantn:
Excellent, I love this. My only comment is that it's too far down on the page and might be missed by users.
Can we move it up higher?

aantn (dismissed):

aantn (approved):

PR #1830, opened 2025-05-29
ROB-1323 cluster status when crashing

RoiGlinik:
sort sinks config before creating them. robusta sink to the end.

only the second commit is the change

### DIFF
--- a/src/robusta/model/config.py
+++ b/src/robusta/model/config.py
@@ -68,8 +68,12 @@ class SinksRegistry:
 
         new_sinks: Dict[str, SinkBase] = dict()
 
-        # Reload sinks, order does matter and should be loaded & added to the dict by config order.
-        for sink_config in new_sinks_config:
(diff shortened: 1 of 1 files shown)


src/robusta/model/config.py
```diff
@@ -45,16 +56,24 @@ def construct_new_sinks(
     ) -> Dict[str, SinkBase]:
         new_sink_names = [sink_config.get_name() for sink_config in new_sinks_config]
         # remove deleted sinks
-        deleted_sink_names = [sink_name for sink_name in existing_sinks.keys() if sink_name not in new_sink_names]
+        deleted_sink_names = [
```
arikalon1, line 59: @RoiGlinik I don't think you can change the order of the sinks
the order of the sinks matters

At least 2 places that I can think of:
1. using `stop` on a sink - to the finding not continue to others
2. Taking a value from one sink, and adding it to others (for example, taking the alert id from OpsGenie, and putting it in a Slack sink.

Please try to think of a different solution to solve this

PR #1835, opened 2025-06-04
ROB-1323 cluster status init crash take 2

RoiGlinik:
## Summary by CodeRabbit

- **Bug Fixes**
	- Improved reliability by ensuring cluster discovery and status updates occur only after configuration is fully initialized. This helps prevent issues related to premature initialization during startup.


src/robusta/runner/config_loader.py
```diff
@@ -67,6 +67,7 @@ def __init__(
         self.watcher = FileSystemWatcher(self.root_playbook_path, self.__reload_playbook_packages)
         self.conf_watcher = FileSystemWatcher(self.config_file_path, self.__reload_playbook_packages)
         self.__reload_playbook_packages("initialization")
+        self.registry.get_event_emitter().emit_event("config_init")
```
arikalon1: should this be done inside reload?
Is there a case where you chance something in the robusta sink, and a new RobustaSink class is created?
I think it's worth checking what happen if you upgrade the chart, with and without doing a change in the robusta sink configuration 

If you indeed move it inside the reload, I would check that I initialize discovery, only if it wasn't initialized before

arikalon1 (commented):
nice work, left one comment

PR #1855, opened 2025-06-26
[ROB-1589] Krr push_scans - requires a new krr version

Avi-Robusta:
### Summary

This PR makes two key changes to the KRR job:

1. **Removes finalizers** – to simplify cleanup and ensure smoother resource deletion.
2. **actionl to Robusta Runner** – after each scan, KRR will call the Robusta Runner's `process_scan` action with the scan result, regardless of whether the scan succeeded or failed.

### Notes

* This change **requires the next KRR release** before it can be merged.
* Behavior is **identical on success and failure of the krr job/pod**:

  * **Success**: Logs and UI remain unchanged.
  * **Failure**: Logs and UI also appear the same.
  * Attached is an example from Robusta logs on a failure case for reference.

Screenshot 2025-06-26 at 12 40 09

### Additional Benefits

* **Resilient scan publishing**: If Robusta is restarted during a scan, it will still be able to publish the scan result afterward.
* **No more job finalizers**: This allows KRR scan jobs to be cleaned up normally without hanging on deletion.
* **Larger scan support**: Scan size is no longer constrained by the log output size, enabling support for larger results.


playbooks/robusta_playbooks/krr.py
```diff
@@ -296,6 +298,105 @@ def _generate_prometheus_secrets(prom_config: PrometheusConfig) -> List[KRRSecre
...
+            "namespace": INSTALLATION_NAMESPACE,
+        },
+    }
+    try:
+        krr_scan = KRRResponse(**params.result)
+    except Exception as e:
+        if isinstance(e, json.JSONDecodeError):
+            logging.exception("*KRR scan job failed. Expecting json result.*")
+        elif isinstance(e, TypeError):
+            logging.error(f"*KRR scan job failed.\n Error from KRR pod:\n {params.result}.*")
```
moshemorad: Do we really want to print params.result? this could be huge bulk of text.

playbooks/robusta_playbooks/krr.py
```diff
@@ -364,7 +467,6 @@ def krr_scan(event: ExecutionBaseEvent, params: KRRParams):
     )
 
     start_time = datetime.now()
-    logs = None
     job_name = f"krr-job-{scan_id}"
```
moshemorad: Please extract and reuse also above

playbooks/robusta_playbooks/krr.py
```diff
@@ -37,6 +38,7 @@
 KRR_MEMORY_LIMIT: str = os.getenv("KRR_MEMORY_LIMIT", "2Gi")
```
moshemorad, line 40: Putting here the comment just to remember to update the version before we merge it.

playbooks/robusta_playbooks/krr.py
```diff
@@ -296,6 +298,105 @@ def _generate_prometheus_secrets(prom_config: PrometheusConfig) -> List[KRRSecre
...
+        event.emit_event(
+            "scan_updated",
+            scan_id=params.scan_id,
+            metadata=metadata,
+            state=ScanState.FAILED,
+            type=ScanType.KRR,
+            start_time=params.start_time,
+        )
+        return
+    else:
```
moshemorad, line 316: No need for the else in that case since you anyway return in the except clause.

Also consider to break down to smaller function just to make it more clearer and easier to maintin. we can discuss it online if you prefer.

playbooks/robusta_playbooks/krr.py
```diff
@@ -296,6 +298,105 @@ def _generate_prometheus_secrets(prom_config: PrometheusConfig) -> List[KRRSecre
...
+    result: Any
+    scan_id: str
+    start_time: datetime
+    
+@action
+def process_scan(event: ExecutionBaseEvent, params: ProcessScanParams):
+    if params.scan_type.lower() != "krr":
+        logging.warning(f"Processing scans not supported for type: {params.scan_type}")
+        return
+    metadata: Dict[str, Any] = {
```
moshemorad: Please add logs in the flow so we will be able to understand better what is going on when debuging.

moshemorad (changes requested):

playbooks/robusta_playbooks/krr.py
```diff
@@ -311,7 +412,9 @@ def krr_scan(event: ExecutionBaseEvent, params: KRRParams):
         subject = event.get_subject()
         args_sanitized = format_event_templated_string(subject, params.args_sanitized)
 
-    python_command = f"python krr.py {params.strategy} {args_sanitized} {additional_flags} "
+    publish_scan_args = f"--publish_scan_url={KRR_PUBLISH_URL} --scan_id={scan_id} --start_time=\"{datetime.now()}\""
```
moshemorad: Last thing. Do we really want to pass the start_time and not just actually take it from the krr.job start time?

For example the job could be pending for couple of minutes before actually start WDYT?

playbooks/robusta_playbooks/krr.py
Avi-Robusta: here it will be either the error message or stack trace if its not a krr result output format.
It shouldnt be a bulk of text

playbooks/robusta_playbooks/krr.py
Avi-Robusta: I think we want the time robusta ran the action not when the scan actually ran, the user doesn't care when the job started, in the UI they consider it started when they click the button

playbooks/robusta_playbooks/krr.py
```diff
@@ -27,16 +28,18 @@
...
 from robusta.integrations.openshift import IS_OPENSHIFT
 from robusta.integrations.prometheus.utils import generate_prometheus_config
 from robusta.utils.parsing import format_event_templated_string
 
-IMAGE: str = os.getenv("KRR_IMAGE_OVERRIDE", f"{IMAGE_REGISTRY}/krr:v1.23.0")
+IMAGE: str = os.getenv("KRR_IMAGE_OVERRIDE", f"{IMAGE_REGISTRY}/krr:v1.24.0")
 KRR_MEMORY_LIMIT: str = os.getenv("KRR_MEMORY_LIMIT", "2Gi")
 KRR_MEMORY_REQUEST: str = os.getenv("KRR_MEMORY_REQUEST", "2Gi")
 KRR_STRATEGY: str = os.getenv("KRR_STRATEGY", "simple")
+KRR_PUSH_SCAN: str = load_bool("KRR_PUSH_SCAN", True)
```
moshemorad, line 43: Thinking here loud, @Avi-Robusta @arikalon1 What do you think we should enabled it by default or for the problematic customers ask to enable it and make sure it works?

moshemorad: https://robustaco.slack.com/archives/C047B4M8Y7K/p1751881847815149

playbooks/robusta_playbooks/krr.py
```diff
@@ -296,57 +300,207 @@ def _generate_prometheus_secrets(prom_config: PrometheusConfig) -> List[KRRSecre
...
+                    for resource in krr_scan.resources
+                ],
+            )
+            for scan in krr_scan.scans
+        ],
+        config=config_json,
+    )
+
+@action
+def fail_krr_scan(event: JobEvent, params: FailedScanParams):
```
arikalon1: I think we never use the params with this action, right?
Is it only for testing?

Might be better to remove it, and simplify the code a little

playbooks/robusta_playbooks/krr.py
```diff
@@ -296,57 +300,207 @@ def _generate_prometheus_secrets(prom_config: PrometheusConfig) -> List[KRRSecre
...
+
+    start_time = params.start_time
+    if not start_time:
+        start_time = job.metadata.annotations.get("start-time")
+
+    if not start_time or not scan_id:
+        logging.error(f"cannot run fail_krr_scan, scan_id or start_time are missing: {event}")
+        return
+
+    metadata = {"job": {"name": f"krr-job-{params.scan_id}", "namespace": INSTALLATION_NAMESPACE}}
```
arikalon1: we're building the metadata in multiple places, might be better to extract to a function

playbooks/robusta_playbooks/krr.py
```diff
@@ -296,57 +300,207 @@ def _generate_prometheus_secrets(prom_config: PrometheusConfig) -> List[KRRSecre
...
+    _emit_failed_scan_event(event, scan_id, start_time, metadata, "Krr Job Failed", None, None)
+
+@action
+def process_scan(event: ExecutionBaseEvent, params: ProcessScanParams):
+    if params.scan_type.lower() != "krr":
+        logging.warning(f"Processing scans not supported for type: {params.scan_type}")
+        return
+
+    logging.info(f"Received process_scan request")
```
arikalon1: I would add the scan_id

playbooks/robusta_playbooks/krr.py
```diff
@@ -296,57 +300,207 @@ def _generate_prometheus_secrets(prom_config: PrometheusConfig) -> List[KRRSecre
...
+    logging.debug(f"process scan contents {params}")
+
+    metadata = {"job": {"name": f"krr-job-{params.scan_id}", "namespace": INSTALLATION_NAMESPACE}}
+
+    try:
```
arikalon1: the code from here until the end looks very similar krr_scan action
Can it be reused?

arikalon1 (dismissed):
nice work

Left a few small comments

arikalon1 (approved):
nice work

PR #1860, opened 2025-07-02
feat(aws ses support): 

stvnksslr:
the library that abstracts email already supports ses technically this is just exposing that support more directly.


arikalon1 (approved):
Thanks for the contribution @stvnksslr !

That's a great enhancement.

arikalon1:
@stvnksslr ,
Can you please sign the CLA with both committing users, so we can merge it?

stvnksslr:
Of course apologies forgot to switch from my work profile for one of the commits. I'll fix it.

stvnksslr:
its been signed not sure when the status updates in the workflows check

arikalon1:
@stvnksslr I'm not sure why the CLA status is incorrect 
Do you mind opening another PR with the same changes, committed with this user?
I think this will be the easiest

PR #1869, opened 2025-07-17
[ROB-1605]  Custom Namespaced resource monitoring

Avi-Robusta:
We can add something like this
```
sinksConfig:
  - robusta_sink:
      name: robusta_ui_sink
...
      namespaceMonitoredResources:
        # Workload controllers
        - apiGroup: "apps"
          apiVersion: "v1"
          kind: "Deployments"
        # Networking
        - apiGroup: "networking.k8s.io"
          apiVersion: "v1"
          kind: "Ingresses"
        # Pod-related
        - apiGroup: ""
          apiVersion: "v1"
          kind: "Pods"
        - apiGroup: "policy"
          apiVersion: "v1"
          kind: "PodDisruptionBudgets"
        - apiGroup: "autoscaling"
          apiVersion: "v2"
          kind: "HorizontalPodAutoscalers"
```
and it will scan (every hour by default) and add to a new column the amount in the namespaces table
in a format like this:
```
{
  "resources": [
    {
      "kind": "StatefulSets",
      "count": 1,
      "apiGroup": "apps",
      "apiVersion": "v1"
    },
    {
      "kind": "Services",
      "count": 1,
      "apiGroup": "",
      "apiVersion": "v1"
    },
    {
      "kind": "Endpoints",
      "count": 1,
      "apiGroup": "",
      "apiVersion": "v1"
    },
    {
      "kind": "ConfigMaps",
      "count": 4,
      "apiGroup": "",
      "apiVersion": "v1"
    },
    {
      "kind": "Secrets",
      "count": 1,
      "apiGroup": "",
      "apiVersion": "v1"
    },
    {
      "kind": "ServiceAccounts",
      "count": 2,
      "apiGroup": "",
      "apiVersion": "v1"
    }
  ]
}
```

### DIFF
--- a/src/robusta/core/sinks/robusta/robusta_sink.py
+++ b/src/robusta/core/sinks/robusta/robusta_sink.py
@@ -378,7 +378,9 @@ class RobustaSink(SinkBase, EventHandler):
         updated_namespaces: List[NamespaceInfo] = []
 
         for namespace_name, namespace in discovered_namespaces.items():
-            namespace.metadata = self.__namespaces_cache.get(namespace_name).metadata
+            cached_namespace = self.__namespaces_cache.get(namespace_name)
+            if cached_namespace:
+                namespace.metadata = cached_namespace.metadata
             updated_namespaces.append(namespace)
         
         return updated_namespaces


src/robusta/core/sinks/robusta/robusta_sink.py
```diff
@@ -556,6 +615,16 @@ def __discovery_watchdog(self):
             time.sleep(DISCOVERY_WATCHDOG_CHECK_SEC)
         logging.warning("Watchdog finished")
 
+    def __discover_namespace_resources(self):
```
arikalon1: I think this is not in use, and should be removed
Also there's no `__activ` 
Looks like some some leftover

src/robusta/core/sinks/robusta/robusta_sink.py
```diff
@@ -135,6 +140,10 @@ def _on_config_reload(self) -> None:
             self._thread = threading.Thread(target=self.__discover_cluster, daemon=True)
             self._thread.start()
 
+        if not hasattr(self, '_custom_resource_thread') and self.namespace_monitored_resources:
```
arikalon1: I think running it in a single thread, introduces some race conditions and synchronisation issues (both in memory and when writing to DB)
I think we should just do it as part of discovery to avoid it.
I think it's fine, if discovery will be a little slower, once an hour - I don't expect this to add much latency

src/robusta/core/sinks/robusta/robusta_sink.py
```diff
@@ -324,6 +333,56 @@ def __send_helm_release_events(self, release_data: List[HelmRelease]):
...
+                                apiGroup=resource.apiGroup,
+                                count=count
+                            ))
+
+                    except ResourceAccessForbiddenError as e:
+                        logging.warning(f"Skipping resource {resource.kind} due to insufficient permissions: {e}")
+                    except Exception as e:
+                        logging.exception(f"Unexpected error counting resource {resource.kind}: {e}")
+
+                namespaces_to_publish = []
```
arikalon1: I think the namespace publishing should be a part of the discovery, and not in parallel on a separate thread

src/robusta/core/sinks/robusta/robusta_sink.py
```diff
@@ -324,6 +333,56 @@ def __send_helm_release_events(self, release_data: List[HelmRelease]):
...
+                        logging.warning(f"Namespace '{namespace_name}' not found in cache when assigning metadata.")
+
+                if namespaces_to_publish:
+                    self.dal.publish_namespaces(namespaces_to_publish)
+                    logging.info(f"Published metadata for {len(namespaces_to_publish)} namespaces")
+
+            except Exception as e:
+                logging.exception(f"Discovery process failed: {e}")
+
+            logging.info(f"Discovery completed, sleeping for {NAMESPACE_DISCOVERY_SECONDS} seconds")
```
arikalon1: would probably be better to add it as a parameter on the sync, with a default value than env var

src/robusta/core/discovery/discovery.py
```diff
@@ -87,7 +88,29 @@ class Config:
...
+    "services": "list_service_for_all_namespaces",
+    "secrets": "list_secret_for_all_namespaces",
+    "persistentvolumeclaims": "list_persistent_volume_claim_for_all_namespaces",
+    "serviceaccounts": "list_service_account_for_all_namespaces",
+    "replicationcontrollers": "list_replication_controller_for_all_namespaces",
+    "limitranges": "list_limit_range_for_all_namespaces",
+    "resourcequotas": "list_resource_quota_for_all_namespaces",
+    "events": "list_event_for_all_namespaces",
+    "podtemplates": "list_pod_template_for_all_namespaces",
+    # Cluster scoped
```
arikalon1: Do we really need those?
Because we don't store anything that isn't namespace scoped

arikalon1 (approved):
nice work!

My only comment, is that we should do it part of discovery, and not in a separate parallel thread.

Please also add docs

PR #1871, opened 2025-07-20
Docs for sending grafana alerts to the timeline

arikalon1:


docs/configuration/alertmanager-integration/grafana-alert-manager.rst
```diff
@@ -1,25 +1,98 @@
...
+
+Save your new ``Notification Policy``
+
+
+That's it!
+
+You can now see your Grafana alerts in the Robusta Timeline, and use AI to analyze it.
+
+
+Send Alerts to Robusta for enrichments
```
aantn, line 93: I don't think it's clear how this is different from the above.

docs/configuration/alertmanager-integration/grafana-alert-manager.rst
```diff
@@ -1,25 +1,98 @@
...
-Grafana can send alerts to Robusta for automatic enrichment and visualization.
+Grafana can send alerts to the Robusta timeline for visualization and AI investigation.
 
 .. image:: /images/grafana-docs-robusta-ui.png
   :width: 600
   :align: center
 
 
-This guide only covers integrating alerts from Grafana Alerting with Robusta, not configuring Robusta to query metrics from the relevant Grafana data source.
+This guide only covers integrating alerts from Grafana Alerting with Robusta.
```
aantn: After the edits, this sentence has lost all its meaning. (It only covers X as opposed to ???)

I suggest either removing it or replacing with 

This guide only covers sending alerts from Grafana Alerting to the Robusta timeline. If you'd like Robusta to also query metrics from Grafana, refer to general :ref:`metrics-integration docs for Prometheus <Integrating with Prometheus>`.

docs/configuration/alertmanager-integration/grafana-alert-manager.rst
```diff
@@ -1,25 +1,98 @@
...
 
-After completing this tutorial, we recommend configuring a metrics integration according to the :ref:`standard instructions for your metrics backend <Integrating with Prometheus>`
-
-Prerequisite
+Kubernetes Alerts
```
aantn: I would put this section beneath Send Alerts to Robusta's timeline. It only makes sense after you've covered the basics.

docs/configuration/alertmanager-integration/grafana-alert-manager.rst
aantn: Could we in theory deprecate the below endpoint (https://api.robusta.dev/integrations/generic/alertmanager) and add a flag 'route_to_cluster' which a user would add to the above integration (https://api.robusta.dev/integrations/alerts/grafana?account_id=YOUR_ACCOUNT_ID) if they were interested in this.

We still need to explain when and how to do this, which I'm not sure is easy to explain

docs/configuration/alertmanager-integration/grafana-alert-manager.rst
arikalon1: fixed

docs/configuration/alertmanager-integration/grafana-alert-manager.rst
arikalon1: fixed

docs/configuration/alertmanager-integration/grafana-alert-manager.rst
arikalon1: This would be a longer task. Opening a ticket for it

aantn (approved):

PR #1873, opened 2025-07-21
feat(ses support):

stvnksslr:
the library that abstracts email already supports ses technically this is just exposing that support more directly.

re-commit pf #1860


arikalon1 (approved):
Thank you for this great PR @stvnksslr 

Nice work!

PR #1877, opened 2025-07-24
Add prometheus discovery selector for standard k8s labels

pavangudiwada:
Add's Prometheus auto discovery for `prometheus-community/prometheus` Helm chart 

CleanShot 2025-07-24 at 11 30 53@2x

CleanShot 2025-07-24 at 11 46 13@2x


arikalon1 (approved):
nice work!

PR #1881, opened 2025-07-28
feat: runner binding address configurable using env variable

donch:
Hi, 

On AWS EKS, you have the ability to create Kubernetes cluster with IPV6 stack only. 
This behavior prevent robusta runner to be reachable by the forwarder since it listen to an IPV4 address.

This PR allow the user to change the bind address for the runner. The default behavior of the runner is unchanged.

### DIFF
--- a/src/robusta/core/model/env_vars.py
+++ b/src/robusta/core/model/env_vars.py
@@ -88,7 +88,7 @@ SERVICE_CACHE_TTL_SEC = int(os.environ.get("SERVICE_CACHE_TTL_SEC", 900))
 SERVICE_CACHE_MAX_SIZE = int(os.environ.get("SERVICE_CACHE_MAX_SIZE", 1000))
 
 PORT = int(os.environ.get("PORT", 5000))  # PORT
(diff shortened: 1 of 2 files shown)


src/robusta/core/model/env_vars.py
```diff
@@ -88,6 +88,7 @@ def load_bool(env_var, default: bool):
 SERVICE_CACHE_MAX_SIZE = int(os.environ.get("SERVICE_CACHE_MAX_SIZE", 1000))
 
 PORT = int(os.environ.get("PORT", 5000))  # PORT
+RUNNER_BIND_ADDR = os.environ.get("RUNNER_BIND_ADDR", "127.0.0.1")  # Listen address for runner
```
moshemorad: Can you please change the default to 0.0.0.0

moshemorad (changes requested):

src/robusta/core/model/env_vars.py
donch: Hi, it's done 👍

PR #1882, opened 2025-07-28
Bugfix: add sink param to krr args

Avi-Robusta:
without it when publishing a scan it would send to all sinks, not just the sinks requested


arikalon1 (dismissed):
nice work

arikalon1 (approved):
nice work

PR #1907, opened 2025-08-26
[ROB-1923] amp irsa support

Avi-Robusta:


arikalon1 (dismissed):
nice work

I think we need a new krr image for this to work, right?

moshemorad (approved):

PR #1915, opened 2025-09-08
Allow ignoring sinks init errors

arikalon1:
Fix finding dates timezone issues

### DIFF
--- a/docs/configuration/holmesgpt/getting-started.rst
+++ b/docs/configuration/holmesgpt/getting-started.rst
@@ -43,6 +43,10 @@ Use Robusta's hosted AI service with GPT-4o:
 
 That's it! HolmesGPT will now analyze your alerts automatically.
 
+.. note::
+
+   When exploring HolmesGPT documentation, focus on **Robusta Helm chart configuration** sections rather than CLI installation. Robusta users should follow the Robusta Helm Chart based configuration examples for data sources and advanced settings.
+
 Test Your Setup
 ---------------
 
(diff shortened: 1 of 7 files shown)


src/robusta/core/reporting/base.py
```diff
@@ -289,7 +289,7 @@ def __init__(
         self.fingerprint = (
             fingerprint if fingerprint else self.__calculate_fingerprint(subject, source, aggregation_key)
         )
-        self.starts_at = starts_at if starts_at else datetime.now()
+        self.starts_at = starts_at if starts_at else datetime.now(timezone.utc)
```
moshemorad: Consider to extract `datetime.now(timezone.utc)` to time_utils and reuse everywhere needed. That way also in future implementation everyone will use it and won't make this mistake.

src/robusta/model/config.py
```diff
@@ -68,24 +70,33 @@ def construct_new_sinks(
...
-                exists_sink.stop()
-                new_sinks[sink_name] = SinkFactory.create_sink(sink_config, registry)
-                continue
-
-            logging.info("Sink %s not changed", sink_name)
-            new_sinks[sink_name] = exists_sink
-
-        return new_sinks
+            
+            try:
```
moshemorad, line 74: Can we please add unitest for that? If we want to release it for cyera and we have time pressure let's merged it and open jira subtask to return and add tests.

Thanks.

moshemorad (dismissed):

src/robusta/model/config.py
arikalon1: ok, opening a ticket

moshemorad (approved):

PR #1929, opened 2025-09-28
ROB-2197 Allow passing custom sys prompt param in holmes_chat action

mershal:


Avi-Robusta (approved):

Avi-Robusta:
FYI this feature is scoped for only chat currently, this will be expanded to other holmes calls like RCA

PR #1942, opened 2025-10-21
[ROB-2343] Launchdarkly docs

Avi-Robusta:


docs/configuration/alertmanager-integration/launchdarkly.rst
```diff
@@ -0,0 +1,139 @@
...
+5. **Resource type**: Select **FLAG**.
+6. **Select flags**: Choose the specific flags you want to monitor, or select all.
+7. **Project and environments**: Select the projects and environments you want to track.
+8. Save the integration.
+
+.. Note::
+
+   Other resource types other than **FLAG** are not verified to work though they might function without additional configuration.
+
+Alternative: Using Headers Instead of URL Parameters
```
arikalon1, line 60: I think this needs to be fixed

I think the code only support api key in header.
I think you want to explain that sending the api key in the url might be visible over the internet, so a more secure way is to send it via headers, where possible

arikalon1 (approved):

Issue #1801, opened 2025-05-05
OCI helm chart


jfroy:
**Describe the solution you'd like**
The kubernetes and helm ecosystem is moving to OCI repositories for helm charts. They provide better performance and better security. Please publish your helm charts to an OCI repository, ideally with verifiable signatures.

**Describe alternatives you've considered**
Run a proxy (e.g. with Harbor or other solutions). More complex to run for small deployments/clusters and doesn't provide the full benefits vs. an upstream OCI chart.

DrFaust92:
this can be done very simply with GH actions and publish as GHCR artifacts. for example from prom helm charts
https://github.com/prometheus-community/helm-charts/actions/runs/16786041408/job/47536630771

arikalon1:
Hey @DrFaust92 
If you'd like to contribute a PR for it, it would be great.
Which OCI repo would you like to use?

PR #1954, opened 2025-11-07
add oci chart release for robusta

DrFaust92:
closes https://github.com/robusta-dev/robusta/issues/1801


DrFaust92:
@arikalon1 can this be looked at?

DrFaust92:
@arikalon1 the action run looks fine but no artifact to pull

there should be something like this https://github.com/runatlantis/helm-charts/pkgs/container/charts%2Fatlantis 
can you check on repo settings if this is disabled for some reason?

arikalon1:
@DrFaust92 can you check now?

arikalon1:
we just released one yesterday @DrFaust92 
The next one should be in 2 weeks or so

PR #1962, opened 2025-11-26
ROB-2555 slack bot stream render graphs

RoiGlinik:
Theres some ruff formating on the file as well.

### DIFF
--- a/src/robusta/core/playbooks/internal/ai_integration.py
+++ b/src/robusta/core/playbooks/internal/ai_integration.py
@@ -588,9 +588,9 @@ def stream_and_render_graphs(url, holmes_req, event):
                     event.ws(data=stream_event)
                     continue
 
-                try:
+                try:  # convert graph tool to png.
                     tool_data = json.loads(tool_res["result"]["data"])
-                    content, name = get_png_from_graph_tool(tool_data)
+                    content, _ = get_png_from_graph_tool(tool_data)
                     tool_res["result"]["data"] = base64.b64encode(content).decode(
                         "utf-8"
                     )


src/robusta/core/playbooks/internal/ai_integration.py
```diff
@@ -454,11 +541,79 @@ def holmes_workload_chat(event: ExecutionBaseEvent, params: HolmesWorkloadHealth
...
+                if event_type == StreamEvents.ANSWER_END.value:
+                    stream_event = re.sub(r"<<.*?>>", "", stream_event)
+
+                if event_type != StreamEvents.TOOL_RESULT.value:
+                    event.ws(data=stream_event)
+                    continue
+
+                tool_res = parse_sse_data(event_lines[1])
+                if not tool_res or tool_res.get("name", "") not in [
+                    "execute_prometheus_range_query",
```
Avi-Robusta: maybe make supported graph query a list so it can be reused everywhere since its used a few places

GRAPH_TOOLS = ["execute_prometheus_range_query", "query_datadog_metrics"]

src/robusta/core/playbooks/internal/ai_integration.py
```diff
@@ -371,30 +452,26 @@ def holmes_chat(event: ExecutionBaseEvent, params: HolmesChatParams):
...
-                    contents = convert_svg_to_png(chart.render())
-                    name = json_content.get("description", "graph").replace(" ", "_")
+                    contents, name = get_png_from_graph_tool(json_content)
                     holmes_result.files.append(FileBlock(f"{name}.png", contents))
 
                 holmes_result.tool_calls = [
-                    tool for tool in holmes_result.tool_calls if tool.tool_name != "execute_prometheus_range_query"
+                    tool
+                    for tool in holmes_result.tool_calls
+                    if tool.tool_name != "execute_prometheus_range_query"
```
Avi-Robusta: i suspect also query_datadog_metrics too, we can use the var GRAPH_TOOLS here too

src/robusta/core/playbooks/internal/ai_integration.py
```diff
@@ -454,11 +541,79 @@ def holmes_workload_chat(event: ExecutionBaseEvent, params: HolmesWorkloadHealth
...
+        stream=True,
+        headers={"Connection": "keep-alive"},
+    ) as resp:
+        resp.raise_for_status()
+        for stream_event in resp.iter_content(
+            chunk_size=None, decode_unicode=True
+        ):  # Avoid streaming chunks from holmes. send them as they arrive.
+            if stream_event:
+                event_lines = stream_event.splitlines()
+                event_type = parse_sse_event_type(event_lines[0])
```
Avi-Robusta: just want to verify that a whole tool call is passed her and it wont ever be passed in multiple chunks correct?
we wont get a multi line thing like
event 1 - ... << execute_prometheus
event 2 - _range_query>> ...

src/robusta/core/playbooks/internal/ai_integration.py
```diff
@@ -454,11 +541,79 @@ def holmes_workload_chat(event: ExecutionBaseEvent, params: HolmesWorkloadHealth
...
+
+                tool_res = parse_sse_data(event_lines[1])
+                if not tool_res or tool_res.get("name", "") not in [
+                    "execute_prometheus_range_query",
+                    "query_datadog_metrics",
+                ]:
+                    event.ws(data=stream_event)
+                    continue
+
+                try:
```
Avi-Robusta: just add a comment here that this is for the graph flow

src/robusta/core/playbooks/internal/ai_integration.py
```diff
@@ -454,11 +541,79 @@ def holmes_workload_chat(event: ExecutionBaseEvent, params: HolmesWorkloadHealth
...
+                    tool_data = json.loads(tool_res["result"]["data"])
+                    content, name = get_png_from_graph_tool(tool_data)
+                    tool_res["result"]["data"] = base64.b64encode(content).decode(
+                        "utf-8"
+                    )
+                    tool_res["result_type"] = "png"
+                    event.ws(data=create_sse_message(event_type, tool_res))
+                except Exception:
+                    logging.exception(
+                        "Failed to convert graph tool to png %s,", event_lines[1]
```
Avi-Robusta, line 601: will event_lines always be two lines?

src/robusta/core/playbooks/internal/ai_integration.py
```diff
@@ -454,11 +541,79 @@ def holmes_workload_chat(event: ExecutionBaseEvent, params: HolmesWorkloadHealth
...
+        resp.raise_for_status()
+        for stream_event in resp.iter_content(
+            chunk_size=None, decode_unicode=True
+        ):  # Avoid streaming chunks from holmes. send them as they arrive.
+            if stream_event:
+                event_lines = stream_event.splitlines()
+                event_type = parse_sse_event_type(event_lines[0])
+
+                if event_type == StreamEvents.ANSWER_END.value:
+                    stream_event = re.sub(r"<<.*?>>", "", stream_event)
```
Avi-Robusta, line 580: add comment that this strips all embedded tool calls so future people will be aware of the magic

src/robusta/core/playbooks/internal/ai_integration.py
```diff
@@ -320,49 +391,59 @@ def holmes_issue_chat(event: ExecutionBaseEvent, params: HolmesIssueChatParams):
...
-                headers={"Connection": "keep-alive"},
-            ) as resp:
-                resp.raise_for_status()
-                for line in resp.iter_content(
-                    chunk_size=None, decode_unicode=True
-                ):  # Avoid streaming chunks from holmes. send them as they arrive.
-                    if line:
-                        event.ws(data=line)
-            return
+            if params.render_graph_images:
```
Avi-Robusta, line 432: I think we want to do most of the same logic in both flows except for the last step
i would pass render_graphs to stream_and_render_graphs and 
change the 
`if not tool_res or tool_res.get("name", "") not in
`
to
 
`if not stream_and_render_graphs or not tool_res or tool_res.get("name", "") not in...
`
so that way we dont need two stream flows, wdyt?

src/robusta/core/playbooks/internal/ai_integration.py
RoiGlinik: Yes its always sent as 1 block

src/robusta/core/playbooks/internal/ai_integration.py
RoiGlinik: added a saftey check but its supposed to be

src/robusta/core/playbooks/internal/ai_integration.py
RoiGlinik: Not sure I understand , with no render graph no parsing is needed so I prefer the flow to be different in case of changes in future and to make the normal flow light as possible

Avi-Robusta: kk

PR #2042, opened 2026-04-03
Fix new playbook modules not discovered during hot reload

### LINKED ISSUE #2041: New playbook modules not discovered during hot reload without pod restart
## Bug Description

When a **new** `.py` file is added to a git-based custom playbook repository and a config reload is triggered (via `FileSystemWatcher` detecting changes to the playbooks config or playbooks directory), the new module is **not discovered** by `pkgutil.walk_packages()`. The new action only becomes available after a full pod restart.

**Changes to existing files work fine** — `importlib.reload()` picks them up. Only **newly added files** are affected.

## Root Cause

In `ConfigLoader.__import_playbooks_package()` (`src/robusta/runner/config_loader.py`), after `pip install` copies new files to site-packages, `pkgutil.walk_packages()` relies on Python's `FileFinder` to enumerate modules. The `FileFinder` **caches directory listings** from the initial import at process startup. Since `importlib.invalidate_caches()` is never called, newly installed `.py` files remain invisible to the finder machinery.

From the [Python docs](https://docs.python.org/3/library/importlib.html#importlib.invalidate_caches):

> If you are dynamically importing a module that was created since the interpreter began execution (e.g., created a Python source file), you may need to call this function in order to notice the new module.

## Steps to Reproduce

1. Configure a git-based custom playbook repo in Robusta
2. Let the runner start and load the initial set of actions
3. Add a **new** `.py` file with a new action to the git repo and push
4. Trigger a config reload (e.g., update the `robusta-playbooks-config-secret`)
5. Observe that the new action is **not** registered — only existing actions are reloaded
6. Restart the pod — the new action is now available

## Expected Behavior

New playbook modules added to git repos should be discovered and loaded during hot reload without requiring a pod restart.

## Proposed Fix

Add `importlib.invalidate_caches()` before the `pkgutil.walk_packages()` call in `__import_playbooks_package()`.

harinik05:
## Summary

- Add `importlib.invalidate_caches()` in `ConfigLoader.__import_playbooks_package()` before `pkgutil.walk_packages()` so that newly added `.py` files in playbook repos are discoverable during hot reload

Fixes #2041

## Problem

When new `.py` files are added to a git-based custom playbook repo and a config reload is triggered, `pkgutil.walk_packages()` does not discover them because Python's `FileFinder` caches directory listings from the initial import. Changes to existing files work fine via `importlib.reload()`, but new modules remain invisible until a full pod restart.

## Fix

A single call to `importlib.invalidate_caches()` before the package walk clears the stale `FileFinder` directory cache, allowing newly installed modules to be found.

## Test plan

- [ ] Deploy runner with a git-based custom playbook repo
- [ ] Add a new `.py` action file to the repo and push
- [ ] Trigger a config reload (e.g., touch the playbooks config secret)
- [ ] Verify the new action is registered without a pod restart
- [ ] Verify existing action modifications still reload correctly


arikalon1 (approved):

Avi-Robusta (approved):
LGTM, great work

PR #2047, opened 2026-04-12
[ROB-2694] holmes-oauth

Avi-Robusta:


src/robusta/core/playbooks/internal/ai_integration.py
```diff
@@ -385,6 +386,45 @@ def holmes_chat(event: ExecutionBaseEvent, params: HolmesChatParams):
...
+    """
+    holmes_url = HolmesDiscovery.find_holmes_url(params.holmes_url)
+    if not holmes_url:
+        raise ActionException(
+            ErrorCodes.HOLMES_DISCOVERY_FAILED,
+            "Robusta couldn't connect to the Holmes client.",
+        )
+
+    try:
+        params_dict = params.dict(exclude={"holmes_url", "model"})
```
moshemorad, line 406: Those are the only 2 var in HolmesParams why we inheriting it only to exclude those fields?

src/robusta/core/playbooks/internal/ai_integration.py
```diff
@@ -385,6 +386,45 @@ def holmes_chat(event: ExecutionBaseEvent, params: HolmesChatParams):
...
+        result = requests.post(
+            f"{holmes_url}/api/oauth/callback",
+            json=params_dict,
+        )
+        result.raise_for_status()
+        response_data = result.json()
+
+        finding = Finding(
```
moshemorad, line 414: Why do we need this finding?

moshemorad (changes requested):

src/robusta/core/playbooks/internal/ai_integration.py
Avi-Robusta: There are additional parameters coming from the frontend that may change over time. I don’t want to update Robusta every time a new parameter is introduced, especially since we’re moving away from it. Instead, HolmesOAuthParams simply forwards a JSON payload from the frontend containing all required OAuth parameters—such as action code, URLs, client ID, and others—while Robusta’s role is only to pass it through.

Thats why HolmesOAuthParams has 
```

    class Config:
        extra = "allow"
```

src/robusta/core/playbooks/internal/ai_integration.py
Avi-Robusta: returns to frontend if the action succeded or failed, if the endpoint fails for any reason, supabase not configured holmes will return failed, frontend gets it and shows something on failure

moshemorad (approved):

PR #2070, opened 2026-05-06
Support multiple environment variable placeholders in config values

aantn:
## Summary
Enhanced environment variable substitution in Robusta configuration to support multiple placeholders in a single field and improved the regex pattern for more robust matching.

## Key Changes
- **Multiple placeholders support**: Configuration values can now contain multiple `{{ env.VAR }}` placeholders (e.g., `"{{ env.USER }}:{{ env.PASSWORD }}"`) instead of just one
- **Static text mixing**: Placeholders can be combined with static text (e.g., `"Bearer {{ env.TOKEN }}"`)
- **Improved regex pattern**: Replaced the loose regex pattern with a stricter `_ENV_PLACEHOLDER_RE` that properly handles whitespace and prevents over-matching
- **Refactored substitution logic**: Changed from extracting the first match to using `re.sub()` with a callback function for proper multi-placeholder replacement

## Implementation Details
- The new regex pattern `r"{{\s*env\.([^}\s]+)\s*}}"` is more precise:
  - Allows flexible whitespace around `env.`
  - Prevents matching incomplete or malformed placeholders
  - Stops at the first `}` or whitespace to avoid capturing extra characters
- The substitution now uses a callback function `_sub()` that processes each match independently, enabling proper handling of multiple placeholders in one value
- Early return optimization: checks for placeholder existence before attempting substitution

## Documentation
Added comprehensive documentation covering:
- Multiple placeholder usage examples
- New `runner.additional_env_froms` feature for bulk secret injection
- Clarification on environment variable scope between runner and HolmesGPT pods
- Configuration guidance for HolmesGPT add-ons (operator, AWS MCP, Azure MCP)

https://claude.ai/code/session_013EFjPwGU1FDY6fs6rcELac


aantn:
@claude review

aantn (commented):
## Review

(Disclosure: I authored the commits on this branch — this is a fresh-eyes pass for self-review.)

### What's good

- The regex `r"{{\s*env\.([^}\s]+)\s*}}"` is a real improvement over the old `{{[ ]*env\.(.*)[ ]*}}`. The old pattern was greedy and would mis-match across two adjacent placeholders (`"{{ env.A }} {{ env.B }}"` → capture group `A }} {{ env.B`). The new pattern stops at the first `}` or whitespace, so multi-placeholder values work correctly.
- The `None`-on-no-placeholder contract is preserved deliberately so existing callers (`runner_config.py:_replace_env_var_in_playbook_repo`, `replace_env_vars_values`) keep their if-checks working. Worth a docstring to make that explicit — easy to break in a future refactor.
- Doc additions cover the three real gaps I'd seen users hit: bulk injection via `additional_env_froms`, runner-vs-Holmes pod scoping, and Holmes add-on deployments.

### Things worth considering

1. **Silent behavior change for an edge case.** Anyone today who wrote `"prefix {{ env.X }}"` was silently getting just the env var's value (prefix dropped). They'll now get `prefix<value>`. The documented usage was always `"{{ env.X }}"` standalone, so I think real-world breakage is near zero — but it's worth a one-line entry in `CHANGELOG.md` so anyone with surprising configs sees it on upgrade.

2. **No tests.** This function had no test coverage before, which is partly how the bug shipped. A small `tests/` file pinning the new behavior (prefix preserved, multiple placeholders, no-placeholder returns `None`, missing env var raises) would be cheap insurance. Happy to add it in a follow-up commit on this branch if you'd like.

3. **Minor perf:** `re.search` then `re.sub` walks the string twice on every matched value. Could collapse to `new, n = _ENV_PLACEHOLDER_RE.subn(_sub, value); return new if n else None`. Negligible in practice — only flagging because it's in a config-load hot path.

4. **Pre-existing doc inconsistency I noticed:** line 15 of `configuration-secrets.rst` (unchanged here) refers to `custom_playbooks` (snake_case), but the actual Helm key is `customPlaybooks` (camelCase, per `helm/robusta/values.yaml:193`). Out of scope for this PR but worth a follow-up.

### CodeRabbit's `is None` finding

Skipping. The "fail on empty env var" semantics is pre-existing (was `if not env_var_value` before this PR) and out of scope for a multi-placeholder-substitution change. Empty values from a Secret almost always indicate a misconfigured `secretKeyRef`, and failing loudly at startup is friendlier than silently injecting `""` into something like an auth header. If we want to allow legitimately-empty values, it deserves a deliberate decision with its own PR and tests.

### Recommendation

Merge after adding a changelog line. Optional follow-ups: tests for `get_env_replacement`, the `subn` micro-optimization, and the `custom_playbooks` doc fix.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #2103, opened 2026-06-18
ROB-397 - global imagePullSecret for Helm chart

Avi-Robusta:
## What

Adds an optional `global.imagePullSecrets` to the Robusta Helm chart so a single value applies the image pull secret to every component, instead of configuring several separate per-component keys.

Covers:
- **runner** (deployment + ServiceAccount)
- **kubewatch / forwarder** (deployment pod-level)
- **runtime pods** the runner launches (KRR, Popeye) — automatically, via the runner ServiceAccount they default to (`RUNNER_SERVICE_ACCOUNT`), so no code changes needed

## Semantics (backwards compatible)

- New `global.imagePullSecrets`, default `[]`.
- **Override**: a per-component value (e.g. `runner.imagePullSecrets`) wins over the global for that component; components without one fall back to global.
- Empty global → render is byte-for-byte unchanged.

## Verification

`helm template` dry-renders confirm:
| Scenario | Result |
|---|---|
| default (nothing set) | no `imagePullSecrets` rendered anywhere |
| global only | same secret on runner deployment + runner SA + forwarder deployment |
| each component individually | each resource gets its own |
| global + local | local wins; components without a local fall back to global |

## Follow-up (not in this PR)

The Holmes subchart change lives in the holmesgpt repo (deployment + SA templates + `values.yaml`) and requires repackaging the bundled `holmes-*.tgz` — separate PR.

Linear: ROB-397

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/setup-robusta/proxies.rst
+++ b/docs/setup-robusta/proxies.rst
@@ -113,6 +113,19 @@ When deploying Robusta in a tightly restricted environment, the runner needs out
 
 If you mirror images to a private registry, override ``image.registry`` (and the per-component ``image:`` fields) in your Helm values and you can drop the public registries from the allowlist.
 
+If your private registry requires authentication, set ``global.imagePullSecrets``. This applies the
+pull secret to every component at once — the runner, kubewatch, HolmesGPT, and the pods the runner
+launches at runtime (e.g. KRR, Popeye, via the runner ServiceAccount):
+
+.. code-block:: yaml
+
+    global:
+      imagePullSecrets:
+        - name: my-registry-secret
+
+A per-component value (e.g. ``runner.imagePullSecrets``, ``kubewatch.imagePullSecrets``) overrides the
+global one for that component. Leaving ``global.imagePullSecrets`` empty keeps the previous behavior.
+
 Verifying the Allowlist
 ----------------------------------------
 
--- a/docs/setup-robusta/proxies.rst
+++ b/docs/setup-robusta/proxies.rst
@@ -114,8 +114,9 @@ When deploying Robusta in a tightly restricted environment, the runner needs out
 If you mirror images to a private registry, override ``image.registry`` (and the per-component ``image:`` fields) in your Helm values and you can drop the public registries from the allowlist.
 
 If your private registry requires authentication, set ``global.imagePullSecrets``. This applies the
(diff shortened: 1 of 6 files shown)


Avi-Robusta:
Good catch — fixed. Switched from `| default` to an explicit nil-check so an explicit empty list is honored as "use none" rather than falling back to the global:

```
{{- $pullSecrets := .Values.runner.imagePullSecrets }}
{{- if kindIs "invalid" $pullSecrets }}{{ $pullSecrets = .Values.global.imagePullSecrets }}{{ end }}
```

The per-component defaults in `values.yaml` are now `~` (null) instead of `[]`, so the semantics are:
- **null (default)** → inherit `global.imagePullSecrets`
- **`[]` explicit** → override with none (opt out)
- **non-empty list** → use it

Applied to all three robusta locations (runner deployment, runner SA, forwarder deployment). Note `forwarder-service-account.yaml` no longer has an imagePullSecrets block — it was removed earlier since the kubewatch SA spawns no runtime pods, so that 4th location in the comment is moot. Same fix also applied to the Holmes subchart (separate repo/PR).

Verified via `helm template`: default renders nothing; global-only applies everywhere; explicit `[]` on the runner opts it out while the runner SA + forwarder still inherit the global.

Avi-Robusta:
Update: after discussion we went with the **simpler, conventional** behavior instead of the `kindIs "invalid"` approach from my previous comment.

```
{{- $pullSecrets := .Values.runner.imagePullSecrets | default .Values.global.imagePullSecrets }}
```

Semantics:
- component sets its own `imagePullSecrets` → that value is used
- component empty/unset (`[]`, the default) → inherits `global.imagePullSecrets`

We intentionally **dropped** the "explicit `[]` opts a component out even when a global is set" case. Supporting it required a null-vs-empty-list distinction (`~` defaults + `kindIs`) that was confusing, and the opt-out scenario isn't meaningful in practice (you wouldn't configure a global registry credential and then deliberately make one workload unable to pull). All component defaults stay `[]`.

Verified via `helm template`: default → nothing rendered; global-only → applied to runner deploy + runner SA + forwarder deploy; per-component → overrides global for that component.

moshemorad (approved):

Issue #2102, opened 2026-06-17
Self-hosted ask_holmes fails: 404 on /api/investigate (bundled Holmes 0.32/0.33 only serves /api/chat)


sraj0210:
### Summary
  On a **self-hosted** install (no Robusta UI/account), a `customPlaybooks` `ask_holmes` action fails because robusta-runner POSTs to `/api/investigate`,
  but the bundled Holmes server only exposes `/api/chat`. Automated AI analysis of alerts never runs.

  ### Environment
  - Robusta Helm chart: **0.41.0** (also confirmed present in 0.42.0)
  - robusta-runner image: `robustadev/robusta-runner:0.41.0`
  - holmes image: `robustadev/holmes:0.32.0` (0.42.0 ships `0.33.0`)
  - Self-hosted, Azure OpenAI (`MODEL=azure/...`), `enableHolmesGPT: true`, **no** Robusta UI sink / account
  - Sink: MS Teams

  ### What happens
  A `customPlaybook` `ask_holmes` action on `on_pod_crash_loop` fails in robusta-runner:

  ```
  ERROR Holmes responded 404 on POST http://robusta-holmes.robusta.svc.cluster.local:80/api/investigate: {"detail":"Not Found"}
  robusta.utils.error_codes.ActionException: Holmes internal configuration error.
    File ".../core/playbooks/internal/ai_integration.py", line 136, in ask_holmes
  ```

  ### Root cause
  - `robusta-runner` `ai_integration.py` (0.41 **and** current `master`) POSTs to `/api/investigate` and `/api/stream/investigate`.
  - The bundled Holmes server (`0.32.0`, and `master`) only registers: `/api/chat`, `/api/model`, `/api/oauth/callback`, `/healthz`, `/readyz` — confirmed
  via the pod's own `/openapi.json`.
  - `/api/investigate` existed in Holmes ≤ `0.13.0` and was removed in the server rewrite; the runner was never migrated to `/api/chat`.

  So the shipped chart pairs a runner and a Holmes that disagree on the endpoint, and self-hosted automated investigation is broken out of the box.

  ### Expected
  Self-hosted automated alert investigation (documented as supported with OpenAI/Azure/Bedrock) should work, or robusta-runner should call `/api/chat`.

  ### Reproduce
  1. Install robusta chart 0.41/0.42 self-hosted, `enableHolmesGPT: true`, Azure model configured.
[...]

PR #2112, opened 2026-07-07
fix: migrate ask_holmes from /api/investigate to /api/chat

cre8ivejp:
## Summary

The bundled Holmes server (>=0.32) only serves /api/chat, but robusta-runner still calls the removed /api/investigate endpoint, breaking self-hosted auto-investigation (issue #2102).

- `ask_holmes` action now calls `/api/chat` (the only endpoint bundled Holmes >=0.32 serves)
- Fixes self-hosted auto-investigation which has been broken since the Holmes server rewrite removed `/api/investigate`
- Guards against `None` context when triggered by `on_prometheus_alert` without explicit params

## Fixes
Related to #2102

## Test plan
- [x] Deployed patched runner to GKE cluster with `enableHolmesGPT: true`
- [x] Sent AlertManager webhook → `ask_holmes` triggered → Holmes `/api/chat` returned analysis
- [x] Verified the analysis Finding is created and sent to Slack sink
- [ ] Note: Slack sink hits 3001 char block limit for long analyses (pre-existing issue in `send_holmes_analysis`, not introduced by this PR)

---

Want me to commit this to the `bucketeer-io/robusta` fork now?

### DIFF
--- a/src/robusta/core/playbooks/internal/ai_integration.py
+++ b/src/robusta/core/playbooks/internal/ai_integration.py
@@ -52,7 +52,9 @@ def build_investigation_title(params: AIInvestigateParams) -> str:
     if params.investigation_type == "analyze_problems":
         return params.ask
 
-    return params.context.get("issue_type", "unknown health issue")
+    if params.context:
+        return params.context.get("issue_type", "unknown health issue")
+    return params.ask or "unknown health issue"
 
 
 def handle_holmes_error(e: Exception) -> NoReturn:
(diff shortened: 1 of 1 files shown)


James-Lu-none:
hi @cre8ivejp , i am hitting the same bug mentioned in https://github.com/robusta-dev/robusta/issues/2102, and i saw you closed your PR for this, is there another fix currently in progress? Appreciate any insights!

PR #2144, opened 2026-08-13
ROB-944 harden pull configs

Avi-Robusta:
## Problem

`ConfigLoader.install_package_remote_tgz` downloads a `.tgz` from a configured `playbook_repos` URL and extracted it with `tar.extractall(path=temp_dir)` — no `members=` and no `filter=`.

Archive contents are remote and unverified. On the Python versions this project targets (`pyproject.toml` pins `>=3.10,<3.12`; the runner image is `python:3.11-slim`), `extractall()` defaults to `fully_trusted`, so a crafted archive can use `../` or absolute member paths — or escaping symlinks — to write files anywhere as root, before the extracted tree is handed to `pip install` (which executes package build code).

## Changes

- **`src/robusta/utils/archive.py`** (new): `safe_extract_tar()` and `UnsafeArchiveError`. Every member is validated up front — absolute paths, traversal outside the destination, symlink/hardlink targets resolving outside the destination, and device/fifo nodes are rejected. Nothing is extracted if any member is unsafe. On interpreters that provide `tarfile.data_filter`, extraction additionally uses `filter="data"`; otherwise the validated `members=` list is passed. Both paths raise the same error type.
- **`src/robusta/runner/config_loader.py`**: the extraction call now goes through `safe_extract_tar`. The existing `except Exception` in `__load_playbooks_repos` logs and skips the offending repo, so a hostile archive fails that repo rather than crashing the runner.
- **`tests/test_config_loader.py`** (new): traversal and absolute-path members rejected through `install_package_remote_tgz` (with the HTTP fetch and `install_package` stubbed), absolute and relative symlink escapes rejected, and a normal nested package still extracting correctly.

HTTP policy is unchanged — `http://` playbook repo URLs still work with the existing warning.

## Testing

- `pytest tests/test_config_loader.py` — 5 passed.
- `pytest tests/test_config_validation.py tests/test_url_validation.py` — 19 passed, no regressions.
- Forced the non-`data_filter` fallback branch by removing `tarfile.data_filter` at runtime: same rejections, same successful extraction of a benign archive.

## Not addressed here

`requests.get` still follows redirects, and there is no download size cap or package signature verification.

---
_Generated by [Claude Code](https://claude.ai/code/session_01GDcapMbrGMnqTjEfAG8chH)_

### DIFF
--- a/src/robusta/utils/archive.py
+++ b/src/robusta/utils/archive.py
@@ -1,0 +1,54 @@
+import os
+import tarfile
+from typing import List
+
+
+class UnsafeArchiveError(Exception):
+    """Raised when a tar archive contains a member that would be extracted outside the target directory."""
+
+
+def _is_within(directory: str, target: str) -> bool:
+    directory = os.path.realpath(directory)
+    target = os.path.realpath(target)
+    return target == directory or target.startswith(directory + os.sep)
+
+
+def _validate_members(tar: tarfile.TarFile, path: str) -> List[tarfile.TarInfo]:
+    members = tar.getmembers()
+    for member in members:
+        if member.isdev():
+            raise UnsafeArchiveError(f"Archive contains a device/fifo member: {member.name}")
+
+        if os.path.isabs(member.name) or member.name.startswith("/"):
+            raise UnsafeArchiveError(f"Archive contains a member with an absolute path: {member.name}")
+
+        if not _is_within(path, os.path.join(path, member.name)):
+            raise UnsafeArchiveError(f"Archive contains a member escaping the target directory: {member.name}")
+
+        if member.issym() or member.islnk():
+            # link targets are resolved relative to the directory holding the link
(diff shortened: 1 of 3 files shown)


src/robusta/utils/archive.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def safe_extract_tar(tar: tarfile.TarFile, path: str) -> None:
+    """
+    Extract a tar archive into ``path``, refusing archives whose members would be written outside of it.
+
+    Guards against path traversal ("zip slip"), absolute member paths, escaping symlinks/hardlinks and
+    device nodes. Raises ``UnsafeArchiveError`` without extracting anything if the archive is unsafe.
+    """
+    if hasattr(tarfile, "data_filter"):  # python 3.12+, and 3.10.12+/3.11.4+ backports
```
moshemorad, line 46: Just to make sure, will this work with the python version in our docker?

moshemorad (approved):

PR #2151, opened 2026-08-16
Add dmesg enricher actions for node and pod events

### LINKED ISSUE #549: Dmesg enricher (action)
**Is your feature request related to a problem?**
Getting Dmesg is helpful for debugging problems on the nodes some cases, but is not always covered in the other log collections mechanism in Robusta.
Also, some people are not aware dmesg so implementing this action will be a nice teaching tool!

**Describe the solution you'd like**
A Dmesg enricher to get the Dmesg from a specific pod's **node**.
The output should be readable as a file.

**Additional context**
- Can probably be implemented in the same manner as "node_bash_enricher" for running a bash command, need to make sure to package the results in a nice FileBlock though.
- [Dmesg man page](https://man7.org/linux/man-pages/man1/dmesg.1.html)

ANONYMOUSZED-beep:
## Summary

Adds two new playbook actions that fetch the kernel ring buffer (`dmesg`) from a node and attach it to the finding as a readable file, as requested in #549.

- **`node_dmesg_enricher`** — runs `dmesg` on the target node via the privileged debugger pod and adds the output as a `FileBlock`.
- **`pod_dmesg_enricher`** — resolves the node from the pod's `spec.nodeName` and does the same, so it works on pod events too (per the maintainer's note that it should work on both).

Both are modeled on `node_bash_enricher` / `pod_bash_enricher` and use `RobustaPod.exec_in_debugger_pod`, which already runs with `SYS_ADMIN` + `privileged`, so `dmesg` can read the kernel ring buffer.

A `DmesgParams.lines` option lets users cap the output to the last N lines (`dmesg | tail -n N`); by default the full buffer is kept.

## Changes

- `playbooks/robusta_playbooks/dmesg_enrichments.py` — new module with both actions (auto-discovered via `pkgutil.walk_packages`, no manual registration needed).
- `tests/test_dmesg_enrichments.py` — 8 unit tests covering command building, file-block output, and the no-node / no-pod / unscheduled-pod guard paths.
- `docs/playbook-reference/actions/event-enrichment.rst` — registers both actions in the Node and Pod enricher sections (docs auto-render from docstrings).

## Testing

```
tests/test_dmesg_enrichments.py ........  [100%]
8 passed in 3.81s
```

Also verified both actions register correctly with the actions registry (`node_dmesg_enricher -> NodeEvent`, `pod_dmesg_enricher -> PodEvent`, params `DmesgParams`), and that `black`/`isort` pass. Existing tests (`test_blocks`, `test_node_enrichment_utils`, `test_popey_job_log_cleanup`) still pass.

Closes #549


ANONYMOUSZED-beep:
Closing as a duplicate of #2146, which predates this PR and already addresses #549 with passing checks. I missed it during the initial issue scan. Thanks to CodeRabbit for surfacing the related PR.

PR #2162, opened 2026-08-27
ROB-1161 Support Supabase publishable keys and fetch the api key from relay

Avi-Robusta:
Prepares the runner for Supabase's publishable/secret keys, which replace the legacy anon/service_role keys.

## What changed

- **supabase client bump + 2.28 integration fixes**: the pinned 2.5.1 rejects `sb_publishable_` keys in `create_client` before any request is made. The bump also required following the client's API moves — `SyncClientOptions`, and `frq.request.params` in `custom_filter_request_builder`.
- **`fetch_supabase_api_key()` + `KEY_CACHE`** in `supabase_dal.py`: on connect the runner asks relay (`GET /api/config/supabase-keys`) for the current key, reporting `account_id`, `cluster`, `component=runner` and `component_version`. The key is cached (cachetools `TTLCache`, 24h) **only after it signs in successfully**; a cached key that fails is dropped and the relay key retried, then the `api_key` from the Robusta token is used as the fallback with failures propagating as before.
- **`__connect` is the single login path**: the expired-JWT retry in `patch_postgrest_execute` now re-runs `__connect` instead of `sign_in`, so a long-lived runner re-reads the cache and re-fetches from relay on session expiry rather than reusing the key chosen at startup.

Footprint of the key-fetch part is ~35 lines in one file, using the `requests` and `TTLCache` imports the DAL already had.

## Backward compatibility

Relays without the endpoint simply fail the fetch, and the runner uses the token's key exactly as today. Relay side: robusta-dev/relay#747.

## Testing

Verified live on a staging-connected cluster:

- Runner starts clean on the new image (sink initialized, relay websocket connected, cluster status flowing, no errors).
- With a **bogus `api_key` in `robusta_sink.token`** the runner still signs in — proving the relay-provided key is what's in use.
- With `ROBUSTA_API_ENDPOINT` pointed at an unreachable host, it logs one warning and falls back to the local key.
- Relay's `supabase_key_requests_total` counter shows the runner's fetches labeled by account, cluster, component and version.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_017zegR4sYcpfdR4HjDTQNES

---
_Generated by [Claude Code](https://claude.ai/code/session_017zegR4sYcpfdR4HjDTQNES)_


poetry.lock
```diff
@@ -1,11 +1,12 @@
-# This file is automatically @generated by Poetry 1.8.5 and should not be changed by hand.
+# This file is automatically @generated by Poetry 2.3.3 and should not be changed by hand.
```
moshemorad: Please lock using the exists poetry version.
If it is something we would like to fix going forward - let's open ticket to upgrade poetry version on the repo

poetry.lock
```diff
@@ -3507,45 +4462,88 @@ docs = ["myst-parser[linkify]", "sphinx", "sphinx-rtd-theme"]
...
-gotrue = ">=1.3,<3.0"
-httpx = ">=0.24,<0.28"
-postgrest = ">=0.14,<0.17.0"
-realtime = ">=1.0.0,<2.0.0"
-storage3 = ">=0.5.3,<0.8.0"
-supafunc = ">=0.3.1,<0.5.0"
+httpx = ">=0.26,<0.29"
+postgrest = "2.28.1"
+realtime = "2.28.1"
+storage3 = "2.28.1"
```
moshemorad: Can you check to see if some of those decencies can be marked as optional?
Here it looks like iceberg is optional https://github.com/supabase/supabase-py/blob/a741396f3115a6b9d69f49702e9efa449883912a/uv.lock#L3574
But we still fetching it.

moshemorad (changes requested):

poetry.lock
Avi-Robusta: Done — the lock was regenerated with Poetry 1.8.5 (the version master's lock was generated with), so the header and lock-version 2.0 match the rest of the repo. No poetry upgrade needed here.

---
_Generated by [Claude Code](https://claude.ai/code)_

poetry.lock
Avi-Robusta: Good catch. In storage3 2.28.1 pyiceberg is a hard dependency (`pyiceberg>=0.10.0`), so it could not be skipped at that version — it only became optional in storage3 2.31.0 (`pyiceberg>=0.10.0; extra == "iceberg"`).

Bumped supabase to 2.31.0, which pins storage3 2.31.0, and pyiceberg is now gone from the lock (as are its transitive deps — the lock shrank by ~730 lines). Tests pass on 2.31.0.

---
_Generated by [Claude Code](https://claude.ai/code)_

moshemorad: There are tons of other dependencies there were added,  are we sure that all of those are really needed and not optional as well? Please validate it

Avi-Robusta: @moshemorad iceberg only becomes optional in supabase 2.31.0 — relay and holmes are on 2.28, so I reverted the bump to keep all three projects consistent.

Nothing to address here otherwise: at 2.28.1 pyiceberg isn't optional, and neither is anything else that got added. Per PyPI metadata, every added package is a hard requirement:

| Package | Declared by (PyPI `requires_dist`) | Optional? |
|---|---|---|
| `pydantic-core` | `pydantic`: `pydantic-core==2.46.5` | no |
| `annotated-types` | `pydantic`: `annotated-types>=0.6.0` | no |
| `typing-inspection` | `pydantic`: `typing-inspection>=0.4.2` | no |
| `yarl` | `supabase`/`postgrest`/`storage3`/`supabase-functions`: `yarl>=1.20.1` | no |
| `multidict` | `yarl`: `multidict>=4.0` | no |
| `propcache` | `yarl`: `propcache>=0.2.1` | no |
| `supabase-auth` | `supabase`: `supabase-auth==2.28.1` (rename of `gotrue`) | no |
| `supabase-functions` | `supabase`: `supabase-functions==2.28.1` (rename of `supafunc`) | no |
| `pyiceberg` + cascade | `storage3` 2.28.1: `pyiceberg>=0.10.0` (extra only from 2.31.0) | no |

If you want iceberg gone specifically we can open a ticket, but it's probably worth a separate one for cleaning up python dependencies generally — and I'd suggest not starting with robusta, since holmes is moving away from it and we plan to make it obsolete.

---
_Generated by [Claude Code](https://claude.ai/code)_

moshemorad (approved):

PR #2173, opened 2026-09-13
Namespace scope runner permissions

arikalon1:


Avi-Robusta (dismissed):

Avi-Robusta:
I am just saying this here, for sure someone will install this with enable prometheus stack and have no idea why they are not getting any prometheus alerts with no playbooks
I would leave in 
```
- name: "DefaultPrometheusAlertEnricher"
  triggers:
  - on_prometheus_alert:
      status: "all"
  actions:
  - default_enricher: {}
```
  just to be safe

Avi-Robusta (approved):

